PR #18, opened 2026-03-06
Key link profiles by git remote URL for cross-clone sharing

rafa-thayto:
## Summary

- Extract inline ANSI color helpers into a shared `src/lib/color.ts` module
- Add `src/lib/git.ts` utility for resolving git repo root, common-dir, and normalized remote URL (e.g. `github.com/org/repo`)
- `resolveProfile` now resolves by normalized remote URL → git common-dir → directory walk, enabling profiles to be shared across clones and worktrees of the same repository
- Add `moveProfile` for migrating old directory-keyed profiles to remote-based keys, with an interactive upgrade prompt
- Replace `select` with a searchable `search` picker in the link command (type-to-filter by app name)
- Show auto-link notice when a profile is resolved via a remote URL from another clone

## Test plan

- [x] Run `bun test` — all existing and new tests pass
- [x] `clerk link` in a repo with a remote stores the profile keyed by normalized remote URL
- [x] Clone the same repo elsewhere — `clerk link` auto-detects the existing link
- [x] `clerk link` in a repo without a remote falls back to git common-dir, then cwd
- [x] `clerk unlink` displays the repo root path in confirmation prompt
- [x] Searchable picker filters apps by name as you type
- [x] Existing directory-keyed profiles trigger an upgrade prompt to migrate to remote-based keys

### DIFF
--- a/src/lib/git.ts
+++ b/src/lib/git.ts
@@ -1,0 +1,79 @@
+import { resolve } from "node:path";
+
+const $ = Bun.$;
+
+interface GitRepoInfo {
+  toplevel: string;
+  commonDir: string;
+  normalizedRemote?: string;
+}
+
+let cached: GitRepoInfo | undefined | null = null;
+
+async function getGitRepoInfo(): Promise<GitRepoInfo | undefined> {
+  if (cached !== null) return cached ?? undefined;
+
+  const result = await $`git rev-parse --show-toplevel --git-common-dir`.quiet().nothrow();
+  if (result.exitCode !== 0) {
+    cached = undefined;
+    return undefined;
+  }
+
+  const lines = result.text().trim().split("\n");
+  const toplevel = lines[0];
+  const commonDir = lines[1];
+  if (!toplevel || !commonDir) {
+    cached = undefined;
+    return undefined;
+  }
+
+  // Fetch remote URL (non-blocking since rev-parse already succeeded)
+  const remoteResult = await $`git remote get-url origin`.quiet().nothrow();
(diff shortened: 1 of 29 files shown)


src/lib/git.ts
```diff
@@ -0,0 +1,79 @@
...
+interface GitRepoInfo {
+  toplevel: string;
+  commonDir: string;
+  normalizedRemote?: string;
+}
+
+let cached: GitRepoInfo | undefined | null = null;
+
+async function getGitRepoInfo(): Promise<GitRepoInfo | undefined> {
+  if (cached !== null) return cached ?? undefined;
```
dmoerner, line 14: Do we need some way to bust this cache?

dmoerner (approved):
Nice! One large-scale comment and one specific comment, but I don't see any blockers at this stage.

The large-scale comment is that I would have liked to see this broken up into 3 PRs. The tests stubbing, the color changes, and the git linking. I think smaller PRs can help improve review velocity and make it easier to focus on things.

src/lib/git.ts
rafa-thayto: in this case we don't because it's just to avoid redundant `git rev-parse` calls if multiple commands read the same info within the same invocation

and it's safe because as CLI process is short-lived, so the cached data can’t go stale

rafa-thayto:
noted your points here @dmoerner ty!

src/lib/git.ts
dmoerner: Got it, thanks!

PR #19, opened 2026-03-06
feat(config): Schema support

dmoerner:
Implement `config schema` command to allow fetching the schema from PLAPI. This will make Clerk CLI config self-documenting, to an extent. The bare command dumps the whole schema, there is also a `--keys` argument to dump only a specific key. This can be useful to save context.

### DIFF
--- a/src/commands/config/schema.ts
+++ b/src/commands/config/schema.ts
@@ -13,7 +13,7 @@ export async function configSchema(
   const resolved = await resolveProfile(process.cwd());
   if (!resolved) {
     console.error(
-      "No Clerk project linked to this directory. Run `clerk init` to set up.",
+      "No Clerk project linked to this directory. Run `clerk link` to set up.",
     );
     process.exit(1);
   }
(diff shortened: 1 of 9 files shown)


src/commands/config/schema.ts
```diff
@@ -0,0 +1,61 @@
...
+  keys?: string[];
+}
+
+export async function configSchema(
+  options: ConfigSchemaOptions,
+): Promise<void> {
+  const resolved = await resolveProfile(process.cwd());
+  if (!resolved) {
+    console.error(
+      "No Clerk project linked to this directory. Run `clerk init` to set up.",
```
brkalow: I think we should recommend `link` here, as init will go through the full Clerk setup

src/commands/config/schema.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  let instance: { id: string; label: string };
+  try {
+    instance = resolveInstanceId(profile, options.instance);
+  } catch (error) {
+    console.error((error as Error).message);
+    process.exit(1);
+  }
+
+  // Use `console.error` for informational messages so stdout is just the JSON response.
```
brkalow, line 31: clever

src/commands/config/schema.ts
wyattjoh: Maybe worthwhile to consider in the future (not now) a standardized error class which can enforce links + formatting for errors with the CLI?

brkalow (approved):

PR #27, opened 2026-03-12
feat: add key detection and autolink to link command

rafa-thayto:
## Summary

- Adds `src/lib/autolink.ts` with `findClerkKeys()` and `matchKeyToApp()` to detect publishable keys from env vars and `.env` files, then match them against Platform API applications
- Uses `detectPublishableKeyName(cwd)` from `framework.ts` to resolve the correct env var name based on the project's framework (e.g., `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` for Next.js)
- Integrates key detection into `clerk link`: when keys are found, suggests the matching app before showing the interactive picker
- Refactors `link()` to extract `handleExistingProfile()`, `resolveApp()`, and `ensureAuth()` helpers, flattening nested control flow
- Skips key suggestion when re-linking (user already linked, chose to switch apps) to avoid suggesting the same app twice
- Shows the target app name in the re-link prompt when `--app` is provided (e.g., "Re-link to My App (app_123)?")

## Test plan

- [x] `bun test` — all 365 tests pass
- [x] Run `clerk link` in a project with a `.env.local` containing a valid publishable key — should suggest the matching app
- [x] Confirm the suggestion and verify the profile is persisted
- [x] Decline the suggestion and verify the interactive picker is shown
- [x] Run `clerk link` when already linked, accept re-link — should show picker directly without suggesting the detected app
- [x] Run `clerk link --app <id>` when already linked — re-link prompt should show the target app name
- [x] Run `clerk link --app <id>` — should skip key detection entirely
- [x] Run `clerk link` with no keys present — should go straight to the picker

### DIFF
--- a/src/commands/link/README.md
+++ b/src/commands/link/README.md
@@ -1,7 +1,7 @@
 # Link Command
 
 Links the current git repository to a Clerk application, storing the app ID
-and instance IDs in `~/.clerk/config.json`. The link is keyed by the
+and instance IDs in the config file. The link is keyed by the
 normalized git remote URL (e.g., `github.com/org/repo`), so it is
 automatically shared across all clones and worktrees of the same repository.
 
@@ -33,12 +33,12 @@ interactive flow.
 3. If `skipIfLinked` and not already linked, tries silent autolink (detect keys → match → persist without prompting)
 4. Checks for authentication (calls `clerk auth login` if needed)
 5. If `--app` is provided, fetches that app directly
-6. Otherwise, fetches the list of applications and scans for publishable keys in env vars / `.env` files
+6. Otherwise, fetches the list of applications and scans for publishable keys in env vars / `.env` / `.env.local`
 7. If a key matches an application, suggests it: "We found \<app\>. Link to this application?"
    - If the user confirms (default), links to the detected app
    - If the user declines, falls through to the interactive picker
 8. If no match, presents a searchable picker (type to filter by name)
-9. Stores the profile in `~/.clerk/config.json` keyed by the normalized remote URL
+9. Stores the profile in the config file keyed by the normalized remote URL
 10. Falls back to git-common-dir or the current directory path if no remote is configured
 
 ## Key Detection
--- a/src/commands/link/README.md
+++ b/src/commands/link/README.md
@@ -28,12 +28,43 @@ interactive flow.
 
 1. Resolves the normalized git remote URL (e.g., `github.com/org/repo`) for cross-clone matching
 2. Checks if already linked — if resolved via a remote URL from another clone, prints an auto-link notice
-3. Checks for authentication (calls `clerk auth login` if needed)
-4. If `--app` is provided, uses that app ID directly
-5. Otherwise, fetches the list of applications and presents a searchable picker (type to filter by name)
-6. Fetches application details to retrieve instance IDs
-7. Stores the profile in `~/.clerk/config.json` keyed by the normalized remote URL
-8. Falls back to git-common-dir or the current directory path if no remote is configured
+   - With `skipIfLinked` (used by `clerk init`), returns immediately
+   - Otherwise, offers to upgrade the profile key to use the git remote if available, or asks to re-link
+3. If `skipIfLinked` and not already linked, tries silent autolink (detect keys → match → persist without prompting)
+4. Checks for authentication (calls `clerk auth login` if needed)
+5. If `--app` is provided, fetches that app directly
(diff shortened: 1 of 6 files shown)


src/lib/autolink.ts
```diff
@@ -0,0 +1,109 @@
...
+export async function findClerkKeys(cwd: string): Promise<DetectedKey[]> {
+  const keys: DetectedKey[] = [];
+  const seen = new Set<string>();
+  const publishableKeyName = await detectPublishableKeyName(cwd);
+
+  function add(key: string, source: string) {
+    if (!key || seen.has(key)) return;
+    seen.add(key);
+    keys.push({ key, source });
+  }
```
wyattjoh, line 32: Right now if it finds one environment variable, the next time it sees it it'll be ignored. I think it should be the other way around, which would be that the last one wins. Might want to double check the loading order of the `.env` files in a package like `dotenv` to check to see if the order of the `ENV_FILES` is correct.

src/lib/autolink.ts
```diff
@@ -0,0 +1,109 @@
...
+      byKey.set(instance.publishable_key, { app, instance });
+    }
+  }
+
+  for (const { key, source } of keys) {
+    const match = byKey.get(key);
+    if (match) return { ...match, source };
+  }
+  return undefined;
+}
```
wyattjoh, line 76: Seems a bit confusing that this is `matchKeyToApp` yet you have to provide multiple keys. Worthwhile adding some JSDoc comments here to the functions so it's clear that it'll return the first matching key/app pair.

src/lib/autolink.ts
```diff
@@ -0,0 +1,109 @@
...
+): Promise<{ path: string; profile: Profile } | undefined> {
+  const detectedKeys = await findClerkKeys(cwd);
+  if (detectedKeys.length === 0) return undefined;
+
+  let apps: Application[];
+  try {
+    apps = await listApplications();
+  } catch {
+    return undefined;
+  }
```
wyattjoh, line 90: Bit nervous around places where we swallow the error and don't even log anything, just a nit, but may be valuable to log perhaps?

src/lib/autolink.ts
```diff
@@ -0,0 +1,109 @@
...
+  const prodInstance = match.app.instances.find((i) => i.environment_type === "production");
+
+  if (!devInstance) return undefined;
+
+  const profile: Profile = {
+    workspaceId: "",
+    appId: match.app.application_id,
+    instances: {
+      development: devInstance.instance_id,
+      ...(prodInstance ? { production: prodInstance.instance_id } : {}),
```
wyattjoh: Always a bit uncomfortable with ternarys like this, just a nit, but maybe breaking it into the following line so you can do an `if` block and conditionally assign the `production` instance key would be cleaner?

src/commands/link/README.md
```diff
@@ -28,12 +28,43 @@ interactive flow.
...
-5. Otherwise, fetches the list of applications and presents a searchable picker (type to filter by name)
-6. Fetches application details to retrieve instance IDs
-7. Stores the profile in `~/.clerk/config.json` keyed by the normalized remote URL
-8. Falls back to git-common-dir or the current directory path if no remote is configured
+   - With `skipIfLinked` (used by `clerk init`), returns immediately
+   - Otherwise, offers to upgrade the profile key to use the git remote if available, or asks to re-link
+3. If `skipIfLinked` and not already linked, tries silent autolink (detect keys → match → persist without prompting)
+4. Checks for authentication (calls `clerk auth login` if needed)
+5. If `--app` is provided, fetches that app directly
+6. Otherwise, fetches the list of applications and scans for publishable keys in env vars / `.env` files
```
wyattjoh: Right now the code only searches for `.env` and `.env.local`. We should either specify that here or maybe add a flag to specify the `.env.<>` file to use?

src/commands/link/README.md
```diff
@@ -28,12 +28,43 @@ interactive flow.
...
+7. If a key matches an application, suggests it: "We found \<app\>. Link to this application?"
+   - If the user confirms (default), links to the detected app
+   - If the user declines, falls through to the interactive picker
+8. If no match, presents a searchable picker (type to filter by name)
+9. Stores the profile in `~/.clerk/config.json` keyed by the normalized remote URL
```
wyattjoh: This is no longer the case, should instead just say "in the config file" or something more generic as it's now platform specific config file locations.

wyattjoh (approved):

PR #28, opened 2026-03-12
refactor: migrate to @napi-rs/keyring and env-paths

wyattjoh:
## Summary

- Replace macOS-only Keychain shell commands (`security add-generic-password` / `find-generic-password` / `delete-generic-password`) with the cross-platform `@napi-rs/keyring` library, enabling credential storage on macOS, Windows, and Linux
- Switch config/data/cache paths from hardcoded `~/.clerk` to XDG-compliant directories via `env-paths` (e.g. `~/.config/clerk-cli`, `~/.local/share/clerk-cli`, `~/.cache/clerk-cli` on Linux; platform-appropriate paths on macOS/Windows)
- `CLERK_CONFIG_DIR` env var continues to work as an override for all paths

## Follow Ups

This doesn't adjust the agent output which currently references hard coded configuration paths — addressed in https://github.com/clerk/cli-new/pull/29.

AIE-635

### DIFF
--- a/src/lib/credential-store.test.ts
+++ b/src/lib/credential-store.test.ts
@@ -1,7 +1,7 @@
 import { test, expect, describe, beforeEach, afterAll, mock, setDefaultTimeout } from "bun:test";
 
 // Keyring initialization can be slow on first access (macOS Keychain, etc.)
-setDefaultTimeout(15_000);
+setDefaultTimeout(5_000);
 import { mkdtemp, rm, mkdir, chmod, unlink } from "node:fs/promises";
 import { join, dirname } from "node:path";
 import { tmpdir } from "node:os";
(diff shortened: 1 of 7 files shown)


src/lib/credential-store.test.ts
```diff
@@ -1,5 +1,8 @@
-import { test, expect, describe, beforeEach, afterAll, mock } from "bun:test";
-import { mkdtemp, rm, mkdir, chmod } from "node:fs/promises";
+import { test, expect, describe, beforeEach, afterAll, mock, setDefaultTimeout } from "bun:test";
+
+// Keyring initialization can be slow on first access (macOS Keychain, etc.)
+setDefaultTimeout(15_000);
```
rafa-thayto:
```suggestion
setDefaultTimeout(5_000);
```

I believe 5s is generous enough. first Keychain access on macOS should take 1-3 seconds at most

rafa-thayto (approved):

PR #33, opened 2026-03-13
feat: add strict CLI typings with @commander-js/extra-typings

wyattjoh:
## Summary

- **Strict option inference** — swap the `commander` import for `@commander-js/extra-typings` (types-only, zero runtime change) so `.opts()` and `.action()` callbacks get inferred types from the `.option()` / `.argument()` chain
- **Chained program creation** — capture `new Command().name().option()...` in a single expression so TypeScript can track the accumulated option types (`{ mode?: string; verbose?: true }`)
- **Typed `runProgram`** — use `ReturnType<typeof createProgram>` instead of bare `Command` so `program.opts().verbose` is typed in the error handler
- Existing manual option interfaces in command files (e.g. `ApiOptions`, `ConfigPushOptions`) are now validated against the command chain at compile time — drift between `.option()` definitions and handler signatures becomes a type error

## Test plan

- [ ] `bun test` — 362 tests pass
- [ ] `bunx tsc --noEmit` — no new type errors in `cli-program.ts`
- [ ] `bun run src/cli.ts --help` — CLI runs normally

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,6 +5,7 @@
     "": {
       "name": "marseille",
       "dependencies": {
+        "@commander-js/extra-typings": "^14.0.0",
         "@inquirer/prompts": "^8.2.0",
         "@napi-rs/keyring": "^1.2.0",
         "commander": "^14.0.3",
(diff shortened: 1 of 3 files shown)


dmoerner (approved):
👍

PR #37, opened 2026-03-16
poc: add `schema` command to fetch Clerk API specs

kylemac:
## Summary

- Adds a `clerk schema` command that fetches OpenAPI specs from the [clerk/openapi-specs](https://github.com/clerk/openapi-specs) repo
- Supports **path-based introspection** to extract specific endpoints or schema types instead of dumping the full spec
- Supports **`--resolve-refs`** to inline all `$ref` references for self-contained output

## Usage

```
clerk schema backend                          # Full Backend API spec (YAML)
clerk schema backend /users                   # Just the /users endpoint
clerk schema backend User                     # Just the User schema type
clerk schema backend /users --resolve-refs    # Endpoint with all $refs inlined
clerk schema backend User --resolve-refs      # Type with all $refs inlined
clerk schema backend --format json            # Full spec as JSON
clerk schema backend --spec-version 2024-10-01  # Specific version
```

## Changes

- **New `clerk schema` command** with public API names (backend, frontend, platform, webhooks) and internal aliases (bapi, fapi)
- **Path introspection**: drill into specific endpoints (`/users`) or types (`User`) with prefix-aware matching and "did you mean?" suggestions
- **`--resolve-refs` flag**: recursively inlines `$ref` references with circular reference detection
- **24-hour local caching** to reduce network requests
- **36 tests** covering aliases, path/type lookup, ref resolution, circular refs, format output, and error handling

## Test plan

- [x] `bun test` — all 398 tests pass
- [x] `bun run lint` / `bun run format` — clean
- [x] Smoke tested `clerk schema backend /users`, `clerk schema backend User`, and `--resolve-refs` against live specs

### DIFF
--- a/src/cli-program.ts
+++ b/src/cli-program.ts
@@ -13,6 +13,7 @@ import { api } from "./commands/api/index.js";
 import { link } from "./commands/link/index.js";
 import { unlink } from "./commands/unlink/index.js";
 import { doctor } from "./commands/doctor/index.js";
+import { openapi } from "./commands/openapi/index.js";
 import { CliError, UserAbortError, ApiError, EXIT_CODE, throwUsageError } from "./lib/errors.js";
 import { red } from "./lib/color.js";
 
@@ -135,6 +136,15 @@ export function createProgram(): Command {
     .option("--dry-run", "Show the request without executing it")
     .option("--yes", "Skip confirmation for mutating requests")
     .action(api);
+
+  program
+    .command("openapi")
+    .description("Fetch the OpenAPI spec for a Clerk API")
+    .argument("[api]", "API name: backend, frontend, platform, or webhooks")
+    .option("--spec-version <version>", "Spec version (default: latest)")
+    .option("--format <format>", "Output format: yaml (default) or json")
+    .option("--output <file>", "Write spec to a file instead of stdout")
+    .action(openapi);
 
   program
     .command("doctor")
--- a/src/cli-program.ts
+++ b/src/cli-program.ts
@@ -13,7 +13,7 @@ import { api } from "./commands/api/index.js";
 import { link } from "./commands/link/index.js";
 import { unlink } from "./commands/unlink/index.js";
 import { doctor } from "./commands/doctor/index.js";
-import { openapi } from "./commands/openapi/index.js";
+import { schema } from "./commands/schema/index.js";
 import { CliError, UserAbortError, ApiError, EXIT_CODE, throwUsageError } from "./lib/errors.js";
 import { red } from "./lib/color.js";
 
@@ -138,13 +138,13 @@ export function createProgram(): Command {
     .action(api);
 
   program
-    .command("openapi")
+    .command("schema")
     .description("Fetch the OpenAPI spec for a Clerk API")
     .argument("[api]", "API name: backend, frontend, platform, or webhooks")
     .option("--spec-version <version>", "Spec version (default: latest)")
     .option("--format <format>", "Output format: yaml (default) or json")
     .option("--output <file>", "Write spec to a file instead of stdout")
-    .action(openapi);
+    .action(schema);
 
   program
     .command("doctor")
--- a/src/cli-program.ts
+++ b/src/cli-program.ts
@@ -141,9 +141,11 @@ export function createProgram(): Command {
     .command("schema")
     .description("Fetch the OpenAPI spec for a Clerk API")
     .argument("[api]", "API name: backend, frontend, platform, or webhooks")
+    .argument("[path]", "Endpoint path (e.g. /users) or schema type (e.g. User)")
     .option("--spec-version <version>", "Spec version (default: latest)")
     .option("--format <format>", "Output format: yaml (default) or json")
     .option("--output <file>", "Write spec to a file instead of stdout")
+    .option("--resolve-refs", "Inline $ref references for self-contained output")
     .action(schema);
 
   program
--- a/src/commands/schema/index.ts
+++ b/src/commands/schema/index.ts
@@ -1,10 +1,13 @@
 /**
  * `clerk schema` — fetch and output the OpenAPI spec for Clerk APIs.
+ *
+ * Supports full-spec output, path-based introspection (e.g. `/users`),
+ * type lookups (e.g. `User`), and `$ref` resolution.
  */
 
 import { mkdir } from "node:fs/promises";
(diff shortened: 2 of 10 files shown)


src/cli-program.ts
```diff
@@ -136,6 +137,17 @@ export function createProgram(): Command {
     .option("--yes", "Skip confirmation for mutating requests")
     .action(api);
 
+  program
+    .command("schema")
+    .description("Fetch the OpenAPI spec for a Clerk API")
+    .argument("[api]", "API name: backend, frontend, platform, or webhooks")
+    .argument("[path]", "Endpoint path (e.g. /users) or schema type (e.g. User)")
+    .option("--spec-version <version>", "Spec version (default: latest)")
+    .option("--format <format>", "Output format: yaml (default) or json")
```
rafa-thayto, line 146: Nice to have: you could use `.choices(["yaml", "json"])` here instead of the manual validation in `schema()`. Commander would reject invalid values automatically *and* display the allowed values in `--help` output. Same idea for the `[api]` argument -- `.choices(["backend", "frontend", "platform", "webhooks", "bapi", "fapi"])` would give you free validation and help text.

```ts
.argument("[api]", "API name").choices(["backend", "frontend", "platform", "webhooks", "bapi", "fapi"])
// ...
.option("--format <format>", "Output format").choices(["yaml", "json"])
```

src/commands/schema/index.ts
```diff
@@ -0,0 +1,337 @@
...
+  return writeOutput(output, options.output);
+}
+
+// ── Write output ──────────────────────────────────────────────────────────────
+
+async function writeOutput(content: string, outputPath?: string): Promise<void> {
+  if (outputPath) {
+    await Bun.write(outputPath, content + "\n");
+    console.error(`Spec written to ${outputPath}`);
+  } else {
```
rafa-thayto, line 97: Consider adding a `--no-cache` or `--refresh` flag. Right now there's no escape hatch if the cached file gets corrupted or the user needs a fresh copy within the 24h TTL window. Something like:

```ts
// in SchemaOptions
noCache?: boolean;

// in fetchSpec
if (!options.noCache) {
  const cached = await readCache(api, version);
  if (cached) return cached;
}
```

src/lib/constants.ts
```diff
@@ -41,9 +41,48 @@ export const BAPI_BASE_URL = process.env.CLERK_BACKEND_API_URL ?? "https://api.c
...
 
+export const OPENAPI_SPECS_BASE_URL = "https://raw.githubusercontent.com/clerk/openapi-specs/main";
+
+export interface OpenApiSpecInfo {
+  /** Directory name in the clerk/openapi-specs repo */
+  repoDir: string;
+  latest: string;
+  versions: string[];
+}
+
```
rafa-thayto, line 52: This version list will go stale whenever a new API version is released. You might want to add a comment noting that, or consider fetching the directory listing from the GitHub API at some point (e.g. `GET repos/clerk/openapi-specs/contents/bapi`) so the CLI can discover versions dynamically. For now, a `TODO` and maybe a simple test that fetches the repo to flag drift would be a low-cost safety net.

```ts
// TODO: consider discovering versions dynamically from the clerk/openapi-specs repo
```

src/commands/schema/index.ts
```diff
@@ -0,0 +1,337 @@
...
+    `No schema type "${query}" found in spec.${hint}\n\nAvailable types (first 10):\n${keys
+      .slice(0, 10)
+      .map((k) => `  ${k}`)
+      .join("\n")}`,
+  );
+}
+
+// ── Ref resolution ───────────────────────────────────────────────────────────
+
+export function resolveAllRefs(node: unknown, root: unknown, seen?: Set<string>): unknown {
```
rafa-thayto, line 190: The `visited.delete(refPath)` pattern here is correct -- it tracks the current *ancestor chain* so sibling refs to the same schema both get resolved. Just wanted to confirm this is intentional since it's the kind of thing that looks like it might be a bug at first glance. Maybe a short comment would help future readers:

```ts
// Remove from visited so sibling references to the same schema
// are resolved (only circular *ancestor* chains are blocked).
visited.delete(refPath);
```

rafa-thayto (commented):
Hey! Nice feature, this would be really useful for agents and developers introspecting the API. A few things to address:

### 1. File paths are under `src/` instead of `packages/cli-core/src/`

This branch looks like it predates the monorepo restructuring. Current `main` has all CLI source under `packages/cli-core/src/`. A rebase + move would be needed before this can merge.

### 2. Uses `console.log`/`console.error` directly

The project has a strict `no-console` oxlint rule and uses `log.*` methods everywhere (see `.claude/rules/logging.md`). This will fail the lint check.

```suggestion
import { log } from "../../lib/log.ts";

// For pipeable data output:
log.data(content);

// For status messages:
log.info(`Spec written to ${outputPath}`);
```

### 3. Tests delete the real `CLERK_CACHE_DIR`

`beforeEach` does `await rm(CLERK_CACHE_DIR, { recursive: true, force: true })` which resolves to the developer's actual cache directory. Running tests locally would wipe your real CLI cache.

```suggestion
// Use a temp directory instead:
const testCacheDir = await mkdtemp(join(tmpdir(), "clerk-schema-test-"));
process.env.CLERK_CONFIG_DIR = testCacheDir;
```

### 4. Cache TTL mismatch

The README says "cached locally for 24 hours" but the code uses `CACHE_TTL_MS` which is 1 hour in `constants.ts`. Either update the docs or define a separate `SCHEMA_CACHE_TTL_MS = 24 * 60 * 60 * 1000`.

### 5. Tests should use `captureLog()` instead of `spyOn(console)`

The project has a `captureLog()` test utility that integrates with the `log.*` system. Once you switch from `console.log` to `log.*`, the tests should use `captureLog()` too.

### 6. Missing changeset

The `Enforce Changeset` workflow will block merge. This needs a minor changeset since it adds a new user-facing command.

### 7. `resolveAllRefs` cycle detection could be stricter

The `visited` set is shared across sibling properties with add/delete. For diamond-shaped ref patterns, cloning the set when entering a new ref would be safer:

```suggestion
visited.add(refPath);
const resolved = resolveAllRefs(target, root, new Set(visited));
visited.delete(refPath);
```

The core implementation is really well done though. The path introspection, type lookup, "did you mean?" suggestions, and circular ref handling are all solid. Just needs the structural updates to land on current main.

PR #31, opened 2026-03-13
feat: add release infrastructure with cross-compiled binaries

wyattjoh:
## Summary

- **Monorepo restructure**: Migrated from flat `src/` layout to a Bun workspace monorepo with two packages — `clerk` (public npm wrapper) and `@clerk/cli-core` (private, core implementation)
- **Cross-compiled binary distribution**: Added a GitHub Actions release workflow that compiles standalone Bun binaries for 8 platform targets (macOS arm64/x64, Linux arm64/x64 glibc+musl, Windows arm64/x64), runs native smoke tests, and publishes platform-specific npm packages (`@clerk/cli-{target}`)
- **Release automation**: Integrated release-please for conventional-commit-based versioning, added a releaser script (`scripts/releaser/`) that generates platform packages with correct `os`/`cpu`/`libc` fields and publishes them to npm with provenance
- **Canary and snapshot channels**: Added `canary` (from main, subset smoke-tested) and `snapshot` (from PR branches via `!snapshot` comment) release workflows for pre-release testing
- **Smart npm shim**: `packages/cli/bin/clerk` detects platform/arch/libc at runtime and resolves the correct native binary via `optionalDependencies`, so `npx clerk` just works without requiring Node.js at runtime
- **Unscoped wrapper package**: The wrapper publishes as `clerk` (unscoped) on npm, while platform-specific binary packages remain scoped under `@clerk/cli-{target}`
- **Safety guardrails**: Binary format verification at build time, smoke tests on native runners before stable publishing, idempotent publish checks, `prepublishOnly` guards, and OIDC trusted publishing (no stored npm tokens)
- **Added MIT license**

## Test plan

- [ ] Verify `bun install` resolves correctly from root
- [ ] Run `bun test` from root — all tests should pass
- [ ] Verify `bun run dev` still works for local development
- [ ] Run `bun run lint` and `bun run format:check` — should pass clean
- [ ] Spot-check release workflow YAML against the target matrix in `scripts/releaser/targets.ts`
- [ ] Review the npm shim (`packages/cli/bin/clerk`) for platform detection edge cases
- [ ] Review canary (`release.yml`) and snapshot (`snapshot.yml`) workflow triggers
- [ ] Verify releaser publishes wrapper as `clerk` and platform packages as `@clerk/cli-{target}`

## Release Notes

* **New Features**
  * Added pre-compiled binary distribution for multiple platforms (macOS ARM64/x64, Linux ARM64/x64 with glibc and musl variants, Windows ARM64/x64)
[...]

PR #38, opened 2026-03-18
feat(init): add init command with framework scaffolding and agent prompts

rafa-thayto:
## Summary

- Rewrite `clerk init` to detect the project's framework, install the Clerk SDK, pull env vars, and scaffold framework-specific boilerplate (middleware, providers, auth pages)
- Support 10 frameworks for detection (Next.js, Astro, Nuxt, TanStack Start, React Router, Vue, Expo, React, Express, Fastify) and 8 for scaffolding
- Fix incorrect SDK package names (`@clerk/clerk-react` → `@clerk/react`, `@clerk/tanstack-start` → `@clerk/tanstack-react-start`) and remove standalone Vite detection
- Add pre-scaffold auth library detection (NextAuth, Auth0, Supabase, Firebase, Passport, Better Auth, Kinde) with migration docs
- Add post-scaffold code scanning for hardcoded keys and leftover auth imports
- All scaffolding is idempotent — existing Clerk setup is detected and skipped, existing middleware is composed rather than overwritten
- Skip login/link when user is already authenticated and linked
- Restructure agent mode prompts into individual markdown templates matching the clerk-docs format (sections, ALWAYS/NEVER rules, deprecated patterns, verification checklists, after-setup guidance)
- Add agent prompts for Expo, Express, and Fastify frameworks

## Agent prompts

Prompts for `clerk init --mode agent` now live as individual `.md` files in `src/commands/init/prompts/`, using `{{PLACEHOLDER}}` interpolation for dynamic context (package manager, file extensions, paths). Each prompt follows the clerk-docs structure:

- One-line summary → Install → Copy-paste code blocks → Rules (ALWAYS/NEVER) → Deprecated patterns → Verify checklist → After Setup

| Template | Framework |
|----------|-----------|
| `nextjs-app-router.md` | Next.js App Router (with Keyless Mode) |
| `nextjs-pages-router.md` | Next.js Pages Router (with Keyless Mode) |
| `react.md` | React / Vite |
| `react-router.md` | React Router |
| `nuxt.md` | Nuxt |
| `tanstack-start.md` | TanStack Start |
| `astro.md` | Astro |
| `vue.md` | Vue |
| `expo.md` | Expo (React Native) |
| `express.md` | Express |
| `fastify.md` | Fastify |
| `generic.md` | Fallback (no framework detected) |
| `generic-fallback.md` | Fallback (framework detected, no specific template) |

## Scaffolding support

| Framework | Middleware | Provider | Auth pages |
|-----------|-----------|----------|------------|
| Next.js (App Router) | `clerkMiddleware` with route protection | `ClerkProvider` in layout | sign-in/sign-up catch-all pages |
| Next.js (Pages Router) | Same as App Router | `ClerkProvider` in `_app` | sign-in/sign-up catch-all pages |
| React Router | `clerkMiddleware` + `rootAuthLoader` | `ClerkProvider` with `loaderData` | sign-in/sign-up routes |
| TanStack Start | `clerkMiddleware` in `requestMiddleware` | `ClerkProvider` in root | sign-in/sign-up routes |
| Nuxt | Auto-configured via module | Auto-configured | sign-in/sign-up Vue pages |
| Astro | `clerkMiddleware` with `onRequest` | `clerk()` integration | sign-in/sign-up Astro pages |
| React / Vite | — | `ClerkProvider` in entry | — |
| Vue | — | `clerkPlugin` in entry | — |

## Test plan

- [x] `bun test` — 415 tests pass, 0 failures
- [x] `bun run lint` — 0 errors
- [x] `bun run format:check` — all files formatted
- [x] TypeScript checks pass for all init command files
- [ ] Manual test: run `clerk init` in a fresh Next.js project
- [ ] Manual test: run `clerk init` in a fresh React (Vite) project
- [ ] Manual test: run `clerk init --mode agent` to verify prompt output

## Summary by CodeRabbit

* **New Features**
  * Init command expanded into a multi-mode, agent-aware onboarding flow with preview & confirm and automatic framework-specific scaffolding (Next.js App/Pages, React, Vue, Nuxt, React Router, TanStack Start, Astro, Expo, Express, Fastify)
  * Automatic post-write formatting of generated files and post-scaffold code scanning for potential issues

* **Documentation**
  * Added comprehensive, framework-specific setup guides and templated prompts to streamline integration and follow-up instructions


packages/cli-core/src/commands/init/README.md
```diff
@@ -1,39 +1,128 @@
...
+| Dependency              | Framework      | Clerk SDK                     | Publishable Key Env Var             |
+| ----------------------- | -------------- | ----------------------------- | ----------------------------------- |
+| `next`                  | Next.js        | `@clerk/nextjs`               | `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` |
+| `astro`                 | Astro          | `@clerk/astro`                | `PUBLIC_CLERK_PUBLISHABLE_KEY`      |
+| `nuxt`                  | Nuxt           | `@clerk/nuxt`                 | `NUXT_PUBLIC_CLERK_PUBLISHABLE_KEY` |
+| `@tanstack/react-start` | TanStack Start | `@clerk/tanstack-react-start` | `VITE_CLERK_PUBLISHABLE_KEY`        |
+| `react-router`          | React Router   | `@clerk/react-router`         | `VITE_CLERK_PUBLISHABLE_KEY`        |
+| `vue`                   | Vue            | `@clerk/vue`                  | `VITE_CLERK_PUBLISHABLE_KEY`        |
+| `expo`                  | Expo           | `@clerk/expo`                 | `EXPO_PUBLIC_CLERK_PUBLISHABLE_KEY` |
+| `react`                 | React          | `@clerk/react`                | `VITE_CLERK_PUBLISHABLE_KEY`        |
```
dmoerner, line 58: You may or may not be aware, but the rename from `clerk-react` to `react` happened very recently with the Core 3 transition, the majority of instances (97% or something) are still on Core 2. We definitely should be forward-looking and focusing on Core 3 here, but I wonder if we should have some sort of fallback detection for Core 2. That's potentially a separate PR since there's a lot of layers possible there, but just flagging it here.

src/commands/init/context.ts
```diff
@@ -0,0 +1,167 @@
...
+  if (await fileExists(join(cwd, `${base}middleware.${ext}`))) return "middleware";
+
+  // Fall back to version detection
+  if (!nextVersion) return "proxy";
+
+  const major = parseNextMajorVersion(nextVersion);
+  if (major === null) return "proxy"; // Unknown version (e.g., "latest", "*")
+
+  return major >= 16 ? "proxy" : "middleware";
+}
```
wyattjoh: Framework specific features should be isolated within their `frameworks/<name>.ts` file, perhaps with an optional context fetching function defined as a part of the `FrameworkScaffold`

src/commands/init/scaffold.ts
```diff
@@ -0,0 +1,43 @@
...
+const SCAFFOLDS: Record<string, FrameworkScaffold> = {
+  "next:app-router": nextjsApp,
+  "next:pages-router": nextjsPages,
+  react: reactVite,
+  "react-router": reactRouter,
+  nuxt: nuxt,
+  "@tanstack/react-start": tanstackStart,
+  astro: astro,
+  vue: vue,
+};
```
wyattjoh: Just a comment, not needed right now. You actually lose some opportunity for better typing here. Instead you could write something like:

```ts
} satisfies Record<string, FrameworkScaffold>;
```

Which will allow the keys to be tied to the actual implementation.

src/commands/init/scaffold.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function getScaffoldKey(ctx: ProjectContext): string {
+  if (ctx.framework.dep === "next") {
+    return `next:${ctx.variant ?? "app-router"}`;
+  }
+  return ctx.framework.dep;
+}
```
wyattjoh: Maybe it's valuable here to consider where the framework scaffold exports the framework name and variant label/key, that way it's not special cased here just for Next.js right now. Seems reasonable that we might add more variants/options in the future.

src/commands/init/scan.ts
```diff
@@ -0,0 +1,199 @@
...
+  return results;
+}
+
+export async function scanForIssues(cwd: string, frameworkDep: string): Promise<ScanFinding[]> {
+  const glob = new Bun.Glob("**/*.{ts,tsx,js,jsx}");
+  const findings: ScanFinding[] = [];
+
+  for await (const relPath of glob.scan({ cwd })) {
+    if (isIgnored(relPath)) continue;
+    if (findings.length >= MAX_FINDINGS) break;
```
wyattjoh: I think as it's not configurable, that it would make sense to remove the `MAX_FINDINGS` for now so that context isn't lost.

src/commands/init/frameworks/astro.ts
```diff
@@ -0,0 +1,161 @@
...
+      };
+    }
+
+    // Existing non-Clerk middleware — skip to avoid overwriting user code
+    return {
+      path,
+      type: "modify",
+      content: "",
+      description: "Create Clerk middleware",
+      skipReason: "Existing middleware found — add clerkMiddleware() manually",
```
wyattjoh: In situations like this, would be really useful to have a "skip" `FileAction`, perhaps changing the `FileAction` into something like:

```ts
export type FileAction =
  | { type: "create"; path: string; content: string; description: string }
  | { type: "modify"; path: string; content: string; description: string }
  | { type: "skip"; path: string; skipReason: string };
```

src/commands/init/frameworks/astro.ts
```diff
@@ -0,0 +1,161 @@
...
+  return addClerkToIntegrations(addClerkImport(content));
+}
+
+async function scaffoldConfig(ctx: ProjectContext): Promise<FileAction | null> {
+  const configPath = await findFirstFile(ctx.cwd, [
+    "astro.config.mjs",
+    "astro.config.ts",
+    "astro.config.js",
+  ]);
+  if (!configPath) return null;
```
wyattjoh: Seems like an issue if we can't configure this and it's an Astro project, maybe this should be an error or at least a skipped file action instead?

packages/cli-core/src/commands/init/frameworks/astro.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function addClerkToIntegrations(content: string): string {
+  if (content.includes("integrations:")) {
+    return content.replace(/(integrations:\s*\[)/, "$1clerk(), ");
+  }
+  if (content.includes("defineConfig")) {
+    return content.replace(/(defineConfig\s*\(\s*\{)/, "$1\n  integrations: [clerk()],");
+  }
+  return content;
+}
```
wyattjoh, line 73: Curious if it's worth it here to use the AST instead? This seems quite fragile.

src/commands/init/frameworks/nextjs-app.ts
```diff
@@ -0,0 +1,94 @@
...
+  scaffoldNextjsMiddleware,
+} from "./helpers.js";
+import type { FileAction, FrameworkScaffold, ProjectContext, ScaffoldPlan } from "./types.js";
+
+async function scaffoldLayout(ctx: ProjectContext): Promise<FileAction | null> {
+  if (!ctx.layoutPath) return null;
+
+  const fullPath = join(ctx.cwd, ctx.layoutPath);
+  const file = Bun.file(fullPath);
+  if (!(await file.exists())) return null;
```
wyattjoh: Seems problematic that if the layout can't be found, that would be a problem worth flagging or erroring on?

src/commands/init/frameworks/nextjs-app.ts
```diff
@@ -0,0 +1,94 @@
...
+      description: "Add ClerkProvider to layout",
+      skipReason: "Already has ClerkProvider",
+    };
+  }
+
+  let newContent = safeAddImport(content, "@clerk/nextjs", "ClerkProvider");
+
+  if (newContent.includes("<body")) {
+    newContent = newContent.replace(/(<body[^>]*>)(\s*)/, "$1$2<ClerkProvider>\n");
+    newContent = newContent.replace(/(\s*)(<\/body>)/, "\n</ClerkProvider>$1$2");
```
wyattjoh: Also seems like a good candidate for the AST, especially, as that way you can enforce modifying the default export for example.

src/commands/init/frameworks/nuxt.ts
```diff
@@ -0,0 +1,92 @@
...
+  const content = await Bun.file(join(ctx.cwd, configPath)).text();
+
+  if (content.includes("@clerk/nuxt")) {
+    return {
+      path: configPath,
+      type: "modify",
+      content,
+      description: "Add @clerk/nuxt to modules",
+      skipReason: "Already has @clerk/nuxt module",
+    };
```
wyattjoh: Seems strange that this returns content/type=modify given it's skipping, maybe makes more sense with the new suggested `FileAction` type where it doesn't accept `content` for the skippable actions anymore.

packages/cli-core/src/commands/init/frameworks/vue.ts
```diff
@@ -0,0 +1,80 @@
...
+    path: entryPath,
+    type: "modify",
+    content: newContent,
+    description: "Add clerkPlugin with publishableKey to Vue app",
+  };
+}
+
+export const vue: FrameworkScaffold = {
+  name: "Vue",
+
```
wyattjoh, line 53: Maybe valuable for us to have these export a supporter version range so that the init won't run on old versions?

packages/cli-core/src/commands/init/prompts/index.ts
```diff
@@ -0,0 +1,134 @@
...
+function loadTemplate(name: string): string {
+  const cached = templateCache.get(name);
+  if (cached) return cached;
+
+  // The project formatter escapes underscores in markdown headings (e.g. `_app` → `\_app`).
+  // These templates are output as plain text, so undo that escaping.
+  const template = readFileSync(join(PROMPTS_DIR, `${name}.md`), "utf-8").replaceAll("\\_", "_");
+  templateCache.set(name, template);
+  return template;
+}
```
wyattjoh: This file is going to break once we start compiling native binaries in #31.

The `loadTemplate` function uses `readFileSync` with `import.meta.dir` to locate the `.md` prompt files at runtime:

```ts
const PROMPTS_DIR = import.meta.dir;
// ...
const template = readFileSync(join(PROMPTS_DIR, `${name}.md`), "utf-8")
```

Two problems with the compiled binary model:

1. **`import.meta.dir` will be wrong.** In a compiled Bun binary, this resolves to the directory of the executable at runtime — not the source tree. The `.md` files won't be there.

2. **`readFileSync` requires files on disk.** With `--compile`, there's no sidecar directory to ship the `.md` files in. Any template load will throw `ENOENT`.

The fix is to replace the dynamic `readFileSync` approach with static text imports, which Bun embeds into the binary at build time:

```ts
import genericMd from "./generic.md" with { type: "text" };
// ...
```

We should sort this out before merging the compilation infrastructure, otherwise `clerk init` will crash at the template-loading step in any distributed binary.

packages/cli-core/src/commands/init/format.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  if (hasBiome) {
+    const proc = Bun.spawn(["npx", "@biomejs/biome", "format", "--write", ...files], {
+      cwd,
+      stdout: "ignore",
+      stderr: "ignore",
+    });
+    await proc.exited;
+  }
```
wyattjoh, line 34: Might be valuable to abstract this a tiny bit, make it easier to support adding new formatters.

rafa-thayto:
@coderabbitai review

jfoshee:
Is this PR ready for review? I see ongoing changes and it is already about 5000 lines. Is some of this generated code?

rafa-thayto:
@jfoshee yeah, it's ready.

It's already about 5,000 lines, with half of it being prompts and the rest scaffolding code

packages/cli-core/src/commands/init/frameworks/helpers.ts
jfoshee, line 1: ❓ Are these helpers just for tests? Or app code as well?

packages/cli-core/src/commands/init/frameworks/nextjs-app.test.ts
```diff
@@ -0,0 +1,473 @@
...
+  await Bun.write(
+    join(tempDir, "middleware.ts"),
+    `import { clerkMiddleware } from "@clerk/nextjs/server";\nexport default clerkMiddleware();`,
+  );
+  await mkdir(join(tempDir, "app"), { recursive: true });
+  await Bun.write(join(tempDir, "app/layout.tsx"), "<html><body>{children}</body></html>");
+
+  const plan = await nextjsApp.scaffold(makeCtx());
+
+  expect(plan.actions[0]).toMatchObject({
```
jfoshee: :question:
```
  expect(plan.actions).toHaveLength(1);
```

packages/cli-core/src/commands/init/frameworks/nextjs-app.test.ts
```diff
@@ -0,0 +1,473 @@
...
+test("skips layout when already has ClerkProvider", async () => {
+  await mkdir(join(tempDir, "app"), { recursive: true });
+  await Bun.write(
+    join(tempDir, "app/layout.tsx"),
+    `import { ClerkProvider } from "@clerk/nextjs";\nexport default function L({ children }) { return <ClerkProvider>{children}</ClerkProvider>; }`,
+  );
+
+  const plan = await nextjsApp.scaffold(makeCtx());
+
+  expect(plan.actions[1]).toMatchObject({
```
jfoshee: Tests that assume arrays are always a certain order can be flaky or fragile. And the reader may not know what is at that index. Can we find the element we are looking for by `type` or something semantic?

packages/cli-core/src/commands/init/index.ts
```diff
@@ -1,110 +1,253 @@
...
+import { scaffold, enrichProjectContext } from "./scaffold.js";
+import { previewPlan, previewAndConfirm } from "./preview.js";
+import { runFormatters } from "./format.js";
+import { detectAuthLibraries, scanForIssues, printFindings } from "./scan.js";
+import { buildAgentPrompt, GENERIC_AGENT_PROMPT, pmInstallCommand } from "./prompts/index.js";
+import type { ProjectContext, ScaffoldPlan } from "./frameworks/types.js";
+import type { ScanFinding } from "./scan.js";
+
+// ---------------------------------------------------------------------------
+// Helpers
```
jfoshee: Minor: I'd try to avoid these banners by just moving these helper functions to another file. wdyt of a file like `heuristics.ts` or something where we have some of the auto detection heuristics for like git dirty, package manager detection, src dir detection, etc?

packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -0,0 +1,652 @@
...
+/** Insert a snippet after the last import statement in a source file. */
+export function insertAfterLastImport(source: string, snippet: string): string {
+  const lastImportIdx = source.lastIndexOf("import ");
+  const lineEnd = source.indexOf("\n", lastImportIdx);
+  if (lineEnd === -1) return source;
+  return source.slice(0, lineEnd + 1) + snippet + source.slice(lineEnd + 1);
+}
+
+/** Wrap the contents of a `<body>` tag with a provider component (e.g. `<ClerkProvider>`). */
+export function wrapBodyWithProvider(content: string, provider: string): string {
```
jfoshee: Should we have a file like `transformations.ts` for gathering these different text transforms? I'm sure it will evolve a lot as we see what patterns emerge.

packages/cli-core/src/commands/init/frameworks/nextjs-app.ts
jfoshee, line 1: I tested on a new next app and it modified code correctly, but whitespace/formatting was wrong:
```
      <body className="min-h-full flex flex-col"><ClerkProvider>
{children}
</ClerkProvider></body>
```
(plus the newline at the EOF was removed).

Let's save that problem for another day, but track it in Linear?

packages/cli-core/src/commands/init/frameworks/nuxt.ts
```diff
@@ -0,0 +1,93 @@
+import { parseModule } from "magicast";
```
jfoshee, line 1: :question: Plan to add tests for nuxt?

packages/cli-core/src/commands/init/frameworks/astro.ts
```diff
@@ -0,0 +1,176 @@
+import { join } from "node:path";
```
jfoshee, line 1: :question: Plan to add tests for Astro?

packages/cli-core/src/commands/init/frameworks/types.ts
```diff
@@ -0,0 +1,45 @@
+import type { FrameworkInfo } from "../../../lib/framework.js";
```
jfoshee, line 1: I think this is a great start on the interfaces. I like them. Let's keep watching for patterns and evolve them.

packages/cli-core/src/commands/init/prompts/astro.md
```diff
@@ -0,0 +1,111 @@
...
+- Import from `@clerk/react` or `@clerk/nextjs` — use `@clerk/astro`
+- Skip `output: 'server'` (Clerk requires SSR)
+
+## Deprecated (DO NOT use)
+
+```typescript
+import { authMiddleware } from '@clerk/astro' // WRONG — use clerkMiddleware
+<SignedIn>                                    // WRONG — use <Show when="signed-in">
+<SignedOut>                                   // WRONG — use <Show when="signed-out">
+output: 'static'                             // WRONG — Clerk requires SSR
```
jfoshee:
```suggestion
output: 'static'                              // WRONG — Clerk requires SSR
```

packages/cli-core/src/commands/init/prompts/astro.md
```diff
@@ -0,0 +1,111 @@
...
+- Add `clerk()` to `integrations` in `astro.config.mjs`
+- Set `output: 'server'` (SSR required) with an SSR adapter
+- Use `clerkMiddleware()` from `@clerk/astro/server` in `src/middleware.ts`
+- Import components from `@clerk/astro/components`
+- Use `<Show>` for conditional rendering
+- Use existing package manager (`{{PM}}`)
+
+NEVER:
+
+- Use `authMiddleware()` (replaced by `clerkMiddleware()`)
```
jfoshee, line 85: I don't know much prompt engineering... But I do know that LLMs can be very "sticky" in how they learn to do something. I'm wondering if each "NEVER" rule should also start with the word "Never".

jfoshee (approved):
I time-boxed my review due to the size of the PR and youth of this project. I think it's more important to get this into people's hands for testing than this have perfect code design. I did not see any blockers and I do hope people start trying this soon.

The test matrix is already getting kind-of big. I know @wyattjoh was looking at E2E tests. I expect we will want to regularly ensure that the edge version of `clerk init` works with the edge versions of all `create` templates for the various frameworks. And, as you know, each framework has variants/options. So, it would be good to start to enumerate all of those cases so we can eventually cover them with automated tests.

I manually tested NextJS and Astro taking the defaults and they worked fine. I expect we will iterate on the output to include things like sign-in buttons / user button, but totally fine to save that for later.

I like the general structure here and presence of unit tests. There are some gaps in testing, so I expect those will be covered in time. 

And, I expect we will continue to keep an eye on the code structure and do refactorings as we see the patterns and logical groupings of functions. I noticed we have heuristics for detecting things in the file system, heuristics for detecting things within code text, operations for transforming text, etc.

Thanks for this huge effort, Rafael.

packages/cli-core/src/commands/init/README.md
rafa-thayto: hey Daniel, I've decided not to add this support
since ⁠`clerk init` is intended for new projects, I believe we should focus on current versions rather than keeping Core 2 available

wdyt?

src/commands/init/frameworks/nextjs-app.ts
rafa-thayto: Good point! 

I agree, `ts-morph` would make this safer, but I chose not to go that way here because it would mean adding another dependency just for this transform, and I believe will be better to keep the CLI lightweight for now. If this ends up being too fragile in practice, I’m happy to revisit it!

> `ts-morph` at about 1.46 MB unpacked

packages/cli-core/src/commands/init/frameworks/helpers.ts
rafa-thayto: for app code.

they're utility functions to reduce the code repetition

PR #45, opened 2026-03-20
feat: add release infrastructure with changesets [3/4]

wyattjoh:
## Summary

#43 and #47 give us a monorepo structure and cross-compiled binaries, but no way to get them to users. We need an automated release pipeline that handles versioning, npm publishing of 9+ packages (1 wrapper + 8 platform packages), and GitHub Releases -- across three channels with different triggers and safeguards.

This PR adds the full release infrastructure:

- **Changesets** (`.changeset/config.json`, `@changesets/cli`) for version management. Contributors add changeset files to PRs; the `changesets/action` bot creates a "Version Packages" PR that bumps versions when merged.
- **`scripts/releaser/index.ts`** -- the core publish script. Generates a platform package for each target (with the correct `os`/`cpu`/`libc` fields), publishes them to npm, updates the wrapper's `optionalDependencies`, publishes the wrapper, creates a git tag, and uploads binaries to a GitHub Release. Supports `--dry-run` and `--tag` for channel selection.
- **`scripts/snapshot.ts`** -- versions packages for snapshot and canary channels using Changesets snapshot mode, producing monotonically-sortable versions like `0.1.0-snapshot.v20260313145959`.
- **`scripts/check-release.ts`** -- detects whether the current version is already published on npm, used by CI to decide whether to trigger a stable release.
- **`scripts/lib/npm.ts`** -- shared `isPublished()` helper that queries the npm registry.
- **`release.yml`** -- CI workflow for stable and canary releases. On push to main: runs `check-release.ts`, if unpublished triggers build → smoke-test → publish. If no stable release needed, publishes a canary instead.
- **`snapshot.yml`** -- CI workflow triggered by `!snapshot` comments on PRs. Builds from the PR branch, smoke-tests, publishes snapshot packages, and posts the install command back as a PR comment.

## Merge instructions

This is part of a 4-PR stack (#43, #47, #45, #46). After squash-merging this PR, retarget #46 to `main` before deleting this branch, then:

```sh
git fetch origin main
git rebase origin/main wyattjoh/release-docs
git push --force-with-lease origin wyattjoh/release-docs
```

Then merge #46.

## Test plan

- [ ] CI passes: `format:check`, `lint`, `build`, `test`
- [ ] Review `release.yml` workflow triggers and job dependencies
- [ ] Review `snapshot.yml` comment-trigger logic and PR comment format
- [ ] Verify `scripts/releaser/index.ts` publish flow handles all 3 channels (stable, canary, snapshot)

[...]

PR #43, opened 2026-03-20
refactor: restructure into Bun workspace monorepo [1/4]

wyattjoh:
## Summary

Today the CLI is distributed as a bundled JS file (`bun build → dist/cli.js`) that runs under Node. This works for development but doesn't support distributing a standalone `npx clerk` binary -- users would need Bun installed, and we can't cross-compile or ship platform-specific native executables.

This PR restructures the repo into a Bun workspace monorepo to enable that:

- **`packages/cli-core/`** -- all CLI source code, commands, and tests (moved from `src/`). Compiled via `bun build --compile` into a standalone native binary per platform.
- **`packages/cli/`** -- the `clerk` npm package users install via `npx clerk` or `npm install -g clerk`. Contains a lightweight Node shim (`bin/clerk`) that:
  1. Detects `process.platform`, `process.arch`, and libc variant (glibc vs musl)
  2. Resolves the matching platform package (`@clerk/cli-darwin-arm64`, `@clerk/cli-linux-x64`, etc.)
  3. Spawns the prebuilt native binary, forwarding args and exit codes

This is the same pattern used by esbuild, SWC, Turbopack, and other tools that ship native binaries through npm -- a thin wrapper package with `optionalDependencies` on per-platform packages, so `npm install` only downloads the binary for the user's OS.

### Other changes

- Update CI workflow to use workspace-filtered scripts
- Add `.ts` extension fixes in imports and test isolation improvements

> **Review note:** 106 files changed, but ~57 are pure renames (`src/` → `packages/cli-core/src/`) with zero content diff -- collapse them in the review UI.

## Merge instructions

This is PR 1 of a 4-PR stack. After squash-merging this PR:

```sh
git fetch origin main
git rebase origin/main wyattjoh/cross-compile
git push --force-with-lease origin wyattjoh/cross-compile
```

Then merge PR 2, and repeat for the rest of the stack.

## Test plan

- [ ] CI passes: `bun install`, `format:check`, `lint`, `build`, `test`
- [ ] `bun run dev` launches the CLI from the new workspace path
- [ ] `bun run build:compile` produces a working binary

PR #47, opened 2026-03-20
feat: add cross-compile build scripts and smoke tests [2/4]

wyattjoh:
## Summary

#43 sets up the monorepo with a wrapper shim that resolves platform-specific binaries, but we can only build a binary for the local machine today. To actually ship `npx clerk`, we need to produce binaries for every platform users run on.

This PR adds the cross-compilation pipeline:

- `scripts/build.ts` -- compiles the CLI via `bun build --compile` for 8 platform targets (darwin-arm64, darwin-x64, linux-arm64, linux-arm64-musl, linux-x64, linux-x64-musl, windows-x64, windows-arm64), with binary format verification using `file` output
- `scripts/releaser/targets.ts` -- single source of truth for target definitions, Bun target mappings, and verify patterns
- `build-binaries.yml` -- reusable CI workflow that matrix-builds all targets on the appropriate runner OS
- `smoke-test.yml` -- reusable CI workflow that runs the compiled binary on native runners to catch runtime issues (e.g. missing shared libraries) before publishing

Also extends root lint/format scripts to cover the new `scripts/` directory and adds `*.bun-build` to `.gitignore`.

## Merge instructions

This is part of a 4-PR stack (#43, #47, #45, #46). After squash-merging this PR, retarget #45 to `main` before deleting this branch, then:

```sh
git fetch origin main
git rebase origin/main wyattjoh/release-infra
git push --force-with-lease origin wyattjoh/release-infra
```

Then merge #45, and repeat.

## Test plan

- [ ] CI passes: `format:check`, `lint`, `build`, `test`
- [ ] `bun run build:compile:all` completes for the local platform target
- [ ] Review `build-binaries.yml` matrix against `targets.ts` definitions

## Summary by CodeRabbit

* **Chores**
  * Added automated multi-platform binary build and reusable smoke-test workflows.
  * Extended official build outputs to macOS, Linux (glibc/musl), and Windows across arm64/x64.
  * Added a cross-compile build tool that compiles, validates, and packages platform-specific binaries.
  * Updated dev scripts and tooling to include additional format/lint steps and a Bun engine requirement.
  * Ignored local compile artifacts from source control.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -1,0 +1,79 @@
+name: Build Binaries
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+      ref:
+        required: true
+        type: string
+      artifact-prefix:
+        required: true
+        type: string
+        description: "Prefix for artifact names (e.g. 'clerk', 'clerk-canary', 'clerk-snapshot')"
+
+jobs:
+  build:
+    runs-on: blacksmith-2vcpu-ubuntu-2404
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref }}
+      - uses: oven-sh/setup-bun@v2
+      - run: bun install --frozen-lockfile
+
+      - name: Build all targets
+        env:
+          CLI_VERSION: ${{ inputs.version }}
+        run: bun run scripts/build.ts --version="${CLI_VERSION}"
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.artifact-prefix }}-darwin-arm64
+          path: dist/artifacts/darwin-arm64/clerk
+          retention-days: 1
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.artifact-prefix }}-darwin-x64
+          path: dist/artifacts/darwin-x64/clerk
+          retention-days: 1
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.artifact-prefix }}-linux-arm64
+          path: dist/artifacts/linux-arm64/clerk
+          retention-days: 1
+
+      - uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.artifact-prefix }}-linux-arm64-musl
+          path: dist/artifacts/linux-arm64-musl/clerk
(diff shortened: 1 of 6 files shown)


wyattjoh:
> **PR Stack** -- Supersedes #31
>
> | | PR | Description |
> |---|-----|-------------|
> | 1 | #43 | Monorepo restructure (merged) |
> | 2 | #47 | Cross-compile build + smoke tests -- **← this PR** |
> | 3 | #45 | Release infrastructure (changesets) |
> | 4 | #46 | Release documentation + LICENSE |
>
> Review in parallel, merge sequentially. Squash-merge each PR, retarget the next PR to `main` before deleting the branch, then rebase.

brkalow:
@wyattjoh Is this an invalid target or a platform-specific issue?

```
Building win32-arm64 (bun-windows-arm64)...
  FAIL: error: Unsupported compile target: bun-windows-aarch64-v1.3.6
```

wyattjoh:
Not an invalid target -- `bun-windows-arm64` is supported in newer Bun versions (works on 1.3.11+, which is what CI uses via `oven-sh/setup-bun@v2` without a pinned version). The error on 1.3.6 is because that release predates Windows ARM64 cross-compilation support.

The build script (`scripts/build.ts`) is only invoked in CI via the `build-binaries.yml` workflow, which always gets the latest Bun. It's not part of the normal local dev loop (`bun run build` / `bun test`), so this shouldn't affect local development on older Bun versions.

That said, we could pin a minimum Bun version in CI or add a version check to the build script if you'd prefer an explicit guard. What do you think?

brkalow:
@wyattjoh makes sense! Maybe we can do a min version check in the script. I was assuming I'd be able to use that to build the CLI locally and install the binary, similar to @jfoshee's PR. Happy to use a different flow for that too

brkalow (approved):

wyattjoh:
Replaced the runtime version check with `"engines": { "bun": ">=1.3.10" }` in the root `package.json`. `oven-sh/setup-bun` already reads this field to select the Bun version in CI. Bun doesn't enforce it locally yet ([oven-sh/bun#5846](https://github.com/oven-sh/bun/issues/5846)), but the error from Bun itself (`Unsupported compile target`) is clear enough when running an older version.

For building locally: `bun run build:compile` (single target, your platform) works on any recent Bun. `bun run build:compile:all` is the cross-compile script that needs 1.3.10+ for the `bun-windows-arm64` target.

PR #48, opened 2026-03-20
chore: add script for devs to "install" cli globally

jfoshee:
by leveraging `bun link`

So `install` is not really the right term. But the end result is after running the new script you can execute `clerk` from any directory on your system and it will always use the latest code.

So this is for CLI devs who want convenience to demo/test the CLI

### Considerations

I also looked at using:

```
bun add -g ./
```

to install it globally. That did not work out of the box. And I believe it installs a snapshot of the files rather than a live link.

Because I am trying to solve a local dev story here I think the live link is preferable. The idea is I make a change in my CLI code, then go over to my test project and try `clerk xyz` immediately. No need to build or anything.  So I think the `link` serves this story quite elegantly.

## Summary by CodeRabbit

* **Chores**
  * Added root-level scripts to link and unlink the CLI package for local development and testing.
  * Registered the CLI as a system executable and switched its runtime to the Bun environment, updating how the command is invoked.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "format:check": "bun run --filter @clerk/cli-core format:check && oxfmt --check scripts/",
     "build:compile": "bun run --filter @clerk/cli-core build:compile",
     "build:compile:all": "bun run scripts/build.ts",
-    "dev:global": "bun link --cwd packages/cli-core",
-    "dev:global:unlink": "bun unlink --cwd packages/cli-core",
+    "global:link": "bun link --cwd packages/cli-core",
+    "global:unlink": "bun unlink --cwd packages/cli-core",
     "start": "bun run build:compile && ./packages/cli-core/dist/clerk",
     "prepare": "git config core.hooksPath .hooks"
   },
--- a/package.json
+++ b/package.json
@@ -13,6 +13,8 @@
     "format:check": "bun run --filter @clerk/cli-core format:check && oxfmt --check scripts/",
     "build:compile": "bun run --filter @clerk/cli-core build:compile",
     "build:compile:all": "bun run scripts/build.ts",
+    "dev:global": "bun link --cwd packages/cli-core",
+    "dev:global:unlink": "bun unlink --cwd packages/cli-core",
     "start": "bun run build:compile && ./packages/cli-core/dist/clerk",
     "prepare": "git config core.hooksPath .hooks"
   },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,6 +13,8 @@
     "format:check": "bun run --filter @clerk/cli-core format:check && oxfmt --check scripts/",
     "build:compile": "bun run --filter @clerk/cli-core build:compile",
     "build:compile:all": "bun run scripts/build.ts",
+    "dev:global": "bun link --cwd packages/cli-core",
```
jfoshee: I really don't know what the right script name should be...  Is there a common convention for this?

wyattjoh: Maybe just be clear what it's doing instead, `link:global` instead?

jfoshee: Sounds good to me. And for unlink: `unlink:global`?

jfoshee: After sleeping on it, wdyt of `global:link` and `global:unlink`?  So they are still siblings? 🤔

rafa-thayto: wdyt about just `link` and `unlink`?

It is always going to be global

packages/cli-core/src/cli.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env bun
```
brkalow, line 1: This file is really only used locally right? When we publish we'll be using the compiled executables. I ask because otherwise this change from `node` to `bun` would be problematic, but in this case I think it's safe.

jfoshee: This file actually fails to run w/ node because this is the "core" entry point. The other entry point for npx correctly has the node shebang.

package.json
jfoshee: Unfortunately that would collide w/ the existing bun commands.

wyattjoh: But it would still be `bun run link` right? I assume that still works.

jfoshee: We should avoid script names that collide w/ built-in commands to [avoid surprises](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

It's common practice to invoke the scripts w/out the `run` subcommand. e.g. `bun dev` rather than `bun run dev`.  So if we have a script called `link`, the developer must now be aware that invoking `bun link` is different than `bun run link`.

rafa-thayto (approved):

PR #53, opened 2026-03-24
fix: Strip `config_version` before pushing

dmoerner:
The backend returns a `config_version` as part of each PATCH or GET for the config. Strip this when doing a push (PATCH or PUT) to allow seamless pull-push workflows, like transferring the config from one application to another.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed config push operations to exclude the `config_version` field from the payload sent to the backend for both PUT and PATCH API requests, including during dry-run mode.

* **Tests**
  * Added test cases verifying that `config_version` is properly excluded from outbound requests while other configuration content remains intact.

### DIFF
--- a/packages/cli-core/src/commands/config/push.test.ts
+++ b/packages/cli-core/src/commands/config/push.test.ts
@@ -253,6 +253,48 @@ describe("config push", () => {
     expect(errorSpy).toHaveBeenCalledWith("Replacing config on development instance...");
   });
 
+  // --- config_version stripping ---
+
+  test("put strips config_version from payload before sending", async () => {
+    let capturedBody = "";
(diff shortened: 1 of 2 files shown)


jfoshee (approved):
Nice!  In a future pass we could put it in the `If-Match` header? Not sure whether that will be intuitive to people or not, though.

dmoerner:
> Nice! In a future pass we could put it in the `If-Match` header? Not sure whether that will be intuitive to people or not, though.

I don't think that's what `If-Match` is for, I think that's for some sort of protection against concurrent updates. We do want a flow that allows both idempotent no-op updates that match the current config, and transformative updates that perform real changes.

PR #54, opened 2026-03-25
feat(config): Support destructive flag

dmoerner:
Add support for the `--destructive` flag on `put` and `patch`. We need to use this for operations that destroyed created config objects, like custom JWT templates or custom OAuth providers.

The behavior of `--destructive` is a bit subtle so I tried to make the help text as clear as possible, but it's probably possible to improve it.

## Summary by CodeRabbit

* **New Features**
  * Added a `--destructive` flag to `config patch` and `config put`, allowing operations that can remove resources via an explicit destructive option.

* **Tests**
  * Added test coverage verifying the destructive flag is sent when enabled and omitted when not set.

### DIFF
--- a/packages/cli-core/src/lib/plapi.ts
+++ b/packages/cli-core/src/lib/plapi.ts
@@ -52,14 +52,15 @@ export async function fetchInstanceConfigSchema(
 ): Promise<Record<string, unknown>> {
   const token = await getAuthToken();
   const url = new URL(
-    `${PLAPI_BASE_URL}/v1/platform/applications/${applicationId}/instances/${instanceId}/config/schema`,
+    `/v1/platform/applications/${applicationId}/instances/${instanceId}/config/schema`,
+    PLAPI_BASE_URL,
   );
   if (keys?.length) {
     for (const key of keys) {
       url.searchParams.append("keys", key);
     }
   }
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/lib/plapi.ts
```diff
@@ -132,9 +132,13 @@ async function sendInstanceConfig(
...
   instanceId: string,
   config: Record<string, unknown>,
+  options?: { destructive?: boolean },
 ): Promise<Record<string, unknown>> {
   const token = await getAuthToken();
-  const url = `${PLAPI_BASE_URL}/v1/platform/applications/${applicationId}/instances/${instanceId}/config`;
+  let url = `${PLAPI_BASE_URL}/v1/platform/applications/${applicationId}/instances/${instanceId}/config`;
+  if (options?.destructive) {
+    url += "?destructive=true";
+  }
```
brkalow: Recommend that we use `URL` / `URLSearchParams` instead of doing string concatenation: https://nodejs.org/api/url.html#class-urlsearchparams

dmoerner: Good callout, updated the library to use `URL`.

brkalow (approved):

PR #56, opened 2026-03-25
feat(config): Support pipes in prompts; config pull --keys

dmoerner:
Two connected improvements to `config` to support common CLI workflows:

1. When using pipes, don't clobber prompts.
2. Allow pulling only specific keys, so that then you could do a patch on just those keys as well.

## Summary by CodeRabbit

* **New Features**
  * Added a --keys option to clerk config pull to fetch only specified config keys.
  * Improved confirmation handling so interactive prompts still work when stdin is not a TTY.

* **Documentation**
  * Updated pull command docs with a --keys example and aligned option formatting.

* **Tests**
  * Added tests for key-filtering behavior and prompt handling edge cases.


Railly (approved):
LGTM. One minor note: `config schema` does `--keys` server-side via query params, `config pull` does it client-side. Not blocking, we can optimize later if needed 👍

PR #57, opened 2026-03-25
feat(config): Diff support for push

dmoerner:
- When doing a push (put/patch), pull the existing config. If there are no changes, return early.
- Display the diff between the two configs, allowing a clearer confirmation flow when in interactive mode.

Sample output, using dim/bold, +/-, and red/green to be accessible to a variety of users:

CleanShot 2026-03-27 at 13 38 36@2x

## Summary by CodeRabbit

* **New Features**
  * Push now previews a grouped leaf-value diff of current vs target configuration (colorized in human mode), and exits early with "No changes detected." when there are no effective changes.
* **Tests**
  * Expanded tests for diff rendering and change-detection, including no-op behavior when unchanged and method-sensitive API stubs to distinguish GET vs mutating requests.

### DIFF
--- a/packages/cli-core/src/commands/config/push.test.ts
+++ b/packages/cli-core/src/commands/config/push.test.ts
@@ -431,6 +431,26 @@ describe("config push", () => {
     expect(errorSpy).toHaveBeenCalledWith("No changes detected.");
   });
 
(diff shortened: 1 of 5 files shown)


rafa-thayto (approved):
remember to resolve the merge conflicts

PR #58, opened 2026-03-25
feat: Store and display app name

dmoerner:
Save the app name when we link and display it in our messages so that users can verify which app they are operating on. When there is no app name saved, fall back to the app ID.

## Summary by CodeRabbit

* **New Features**
  * Status messages now display both the target app identifier and environment label (e.g., "Pulling config from MyApp (production)") for clearer context.
  * App names are now persisted and referenced in command output, improving readability when managing multiple applications.

* **Chores**
  * Updated test expectations to match enhanced status messaging across config and environment commands.


rafa-thayto (approved):
nice!

PR #64, opened 2026-03-26
fix(init): scaffold Vue files instead of showing post-instructions

rafa-thayto:
## Summary
- **Create `main.ts` when it doesn't exist** — instead of falling back to a post-instruction telling the user to manually add `clerkPlugin`, the scaffold now generates a complete entry file with Clerk plugin setup
- **Scaffold auth pages + env vars** — creates `sign-in.vue`, `sign-up.vue` pages and adds `VITE_CLERK_*` env vars to `.env`
- **Auto-modify Vue Router config** — finds `src/router/index.ts` (or `.js`) and inserts sign-in/sign-up routes. Only falls back to a post-instruction if `vue-router` is a dep but no router file is found
- Previously running `clerk init` on a Vue project without `main.ts` would show "No files to scaffold" with manual instructions — now it writes all the files

## Test plan
- [x] All 14 Vue scaffold tests pass (`bun test packages/cli-core/src/commands/init/frameworks/vue.test.ts`)
- [x] Full test suite passes (555/557, 2 pre-existing failures in `api/index.test.ts`)
- [x] Lint and format pass
- [ ] Manual test: run `clerk init` on a fresh Vue project without `main.ts` — should create entry file, auth pages, env vars
- [ ] Manual test: run `clerk init` on a Vue project with `vue-router` — should modify router config
- [ ] Manual test: run `clerk init` twice — second run should skip all files (idempotent)

### DIFF
--- a/packages/cli-core/src/commands/init/README.md
+++ b/packages/cli-core/src/commands/init/README.md
@@ -132,9 +132,13 @@ Nuxt's module system auto-configures middleware and auto-imports components.
 
 ### Vue
 
-| Action | File      | Description                                        |
-| ------ | --------- | -------------------------------------------------- |
-| MODIFY | `main.ts` | Add `clerkPlugin` with `publishableKey` to Vue app |
+| Action        | File                    | Description                                        |
+| ------------- | ----------------------- | -------------------------------------------------- |
+| CREATE/MODIFY | `main.ts`               | Add `clerkPlugin` with `publishableKey` to Vue app |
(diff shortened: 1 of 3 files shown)


Railly (approved):
LGTM 👍. Nit: the route check in `addSignRoutes` and `scaffoldRouter` uses `AND` (`includes("/sign-in") && includes("/sign-up")`), so if only one route exists it'll insert both and duplicate the existing one. Would be safer with OR. Not blocking, just edge-casey.

PR #65, opened 2026-03-26
feat: add next-step suggestions after every command

Railly:
## Summary

Adds contextual next-step suggestions after successful commands. Human-mode only (stderr) — agents and piped output are unaffected.

- New `printNextSteps()` utility in `lib/next-steps.ts`
- Added to: `auth login`, `link`, `unlink`, `env pull`, `config pull`, `config patch/put`, `deploy`
- Unified `init`'s inline next-steps rendering to use the shared utility
- Skipped: `whoami` (already suggests login inline), `doctor` (check marks communicate success), `config schema` (read-only)

### Example output

```
Linked to My SaaS App in /Users/dev/my-project

Next steps:
  → Run `clerk env pull` to fetch your environment variables
  → Run `clerk doctor` to verify your setup
```

Closes AIE-696

### DIFF
--- a/packages/cli-core/src/commands/init/heuristics.ts
+++ b/packages/cli-core/src/commands/init/heuristics.ts
@@ -1,6 +1,7 @@
 import { join, dirname } from "node:path";
 import { mkdir } from "node:fs/promises";
 import { dim, cyan, green, yellow, bold } from "../../lib/color.js";
+import { printNextSteps } from "../../lib/next-steps.js";
 import { getToken } from "../../lib/credential-store.js";
 import { fetchUserInfo } from "../../lib/token-exchange.js";
 import { printFindings } from "./scan.js";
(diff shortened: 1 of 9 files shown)


packages/cli-core/src/commands/init/heuristics.ts
```diff
@@ -79,12 +80,7 @@ export function printOutro(plan: ScaffoldPlan, findings: ScanFinding[]): void {
     console.log(`  ${dim("-")} ${dim(a.path)} ${dim(`(${a.skipReason})`)}`);
   }
 
-  if (plan.postInstructions.length > 0) {
-    console.log(dim("\nNext steps:"));
-    for (const instr of plan.postInstructions) {
-      console.log(dim(`  • ${instr}`));
-    }
-  }
```
Railly, line 87: I replaced this with the shared `printNextSteps()` utility so all commands render next steps the same way. Also moves from stdout to stderr, it avoids corrupting piped JSON. cc @rafa-thayto

dmoerner:
I like this idea in principle but I feel like it will get old fast, especially with config. Should we add some sort of a config to opt out of it, or maybe set different levels of verbosity (run always, run on auth/link/unlink/deploy, do not run)?

Railly:
Good call. What about keeping hints on setup commands (`auth login`, `init`, `link`, `deploy`) and drop them from repetitive workflow commands (`config`, `env pull`, `unlink`) @dmoerner

dmoerner:
@Railly Hm, for `unlink` I think the natural flow there is `unlink -> link`. But maybe the semantic connection makes that so obvious that the next-step will start to feel like noise?

Railly:
yeah, unlink->link is obvious enough for me, i'll push an update shortly!

dmoerner (approved):

PR #77, opened 2026-03-27
feat(auth): prompt for re-authentication when already logged in

rafa-thayto:
## Summary
- When a user runs `clerk auth login` with a valid session, show a confirmation prompt ("You're already logged in as X. Re-authenticate?") instead of silently returning
- In agent mode, preserves the existing behavior (logs and returns early)
- Refactors `login()` into small focused functions (`getExistingSession`, `openBrowser`, `performOAuthFlow`) and exports the `UserInfo` type from `token-exchange.ts`

## Test plan
- [x] Run `clerk auth login` when already authenticated — should see re-auth prompt
- [x] Decline the prompt — should exit cleanly (no error output)
- [x] Accept the prompt — should open browser for fresh OAuth flow
- [x] Run `clerk auth login` when not authenticated — should proceed directly to OAuth
- [x] Run in agent mode (`--mode agent`) when authenticated — should log and return silently (no prompt)

## Screenshots
- `clerk auth` when unauthenticated
image

- `clerk auth` when authenticated
image

- `clerk auth --mode agent` when authenticated
image

### DIFF
--- a/packages/cli-core/src/commands/auth/README.md
+++ b/packages/cli-core/src/commands/auth/README.md
@@ -8,7 +8,7 @@ Manage authentication with Clerk.
 
 Authenticates the user via an OAuth 2.0 PKCE flow.
 
-1. Checks for an existing valid token and skips login if found
+1. Checks for an existing valid token — if found, prompts to re-authenticate (in agent mode, skips login silently)
 2. Generates PKCE parameters (code verifier, challenge, state)
 3. Starts a local HTTP callback server on `127.0.0.1`
 4. Opens the browser to the Clerk OAuth authorization URL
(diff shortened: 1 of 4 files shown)


Railly (commented):
Hey! The existing test `"returns early when already authenticated"` passes but it's actually hitting the agent path (`isTTY` is undefined in the test runner so `isHuman()` returns false). The confirm prompt isn't covered.

Would add mocks for `mode.ts` and `prompts.ts` to test:

- Human + existing session + accept -> OAuth runs
- Human + existing session + decline -> `UserAbortError`
- Agent + existing session → early return

rafa-thayto:
just added @Railly

Railly (approved):

PR #80, opened 2026-03-31
feat(init): support keyless mode for zero-friction setup

rafa-thayto:
## Summary

- Allow `clerk init` to work **without a Clerk account** by leveraging Clerk's keyless mode (SDK auto-generates temporary dev keys)
- When unauthenticated, prompt users to choose: "Continue with temporary keys" or "Log in to an existing Clerk account"
- Move `env pull` to **after scaffolding** so project files are set up before pulling API keys
- `--yes` flag defaults to keyless mode when not already authenticated (fastest path)

## Test plan

- [ ] Run `clerk init` without auth → prompt appears with keyless/login choices
- [ ] Select "Continue with temporary keys" → SDK installs, files scaffold, keyless info prints, no env pull
- [ ] Select "Log in" → OAuth flow → link → scaffold → env pull at end
- [ ] Run `clerk init` when already authenticated + linked → no prompt, authenticated flow
- [ ] Run `clerk init -y` without auth → keyless mode automatically
- [ ] Run `clerk init -y` when already auth+linked → authenticated mode
- [ ] Run `clerk init --prompt` → unchanged agent mode behavior
- [ ] `bun test` passes (705/707, 2 pre-existing failures in api/index.test.ts)
- [ ] `bun run lint` passes
- [ ] `bun run format` passes

### DIFF
--- a/packages/cli-core/src/commands/init/README.md
+++ b/packages/cli-core/src/commands/init/README.md
@@ -18,7 +18,7 @@ clerk init --yes
 | -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `--framework <name>` | Framework to set up (skips auto-detection). Valid values: `next`, `astro`, `nuxt`, `tanstack-start`, `react-router`, `vue`, `expo`, `react`, `express`, `fastify` |
 | `--prompt`           | Output a prompt for an AI agent to integrate Clerk, then exit                                                                                                     |
-| `-y, --yes`          | Skip confirmation prompts                                                                                                                                         |
+| `-y, --yes`          | Skip confirmation prompts. Defaults to keyless mode if not already authenticated and linked.                                                                      |
 
 ## Agent Mode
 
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/commands/init/index.ts
```diff
@@ -46,36 +54,51 @@ export async function init(options: InitOptions = {}) {
...
 
-  // Not authenticated — full flow
-  await login();
+  // Not authenticated + --yes flag: default to keyless (fastest path, no browser)
+  if (yes) return { mode: "keyless", email: null };
+
+  const mode = await select<AuthMode>({
+    message: "How would you like to set up Clerk?",
+    choices: [
+      { name: "Continue with temporary keys (no account needed)", value: "keyless" },
```
jfoshee: What do you think of "connect your account later" instead of "no account needed"?  Or something similar, so they know this is not like an irreversible decision.

packages/cli-core/src/commands/init/README.md
```diff
@@ -18,7 +18,7 @@ clerk init --yes
 | -------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `--framework <name>` | Framework to set up (skips auto-detection). Valid values: `next`, `astro`, `nuxt`, `tanstack-start`, `react-router`, `vue`, `expo`, `react`, `express`, `fastify` |
 | `--prompt`           | Output a prompt for an AI agent to integrate Clerk, then exit                                                                                                     |
-| `-y, --yes`          | Skip confirmation prompts                                                                                                                                         |
+| `-y, --yes`          | Skip confirmation prompts. Defaults to keyless mode if not already authenticated and linked.                                                                      |
```
jfoshee: I am not sure we need to spell this out here... It seems like an implementation detail. Open to other opinions though.

packages/cli-core/src/commands/init/README.md
jfoshee, line 27: Who is this documentation for? CLI devs? or CLI users?

jfoshee (approved):
We do need an integration/e2e test for this somewhere... Maybe after Wyatt merges the testing framework you can add a test.

packages/cli-core/src/commands/init/README.md
rafa-thayto: Users, agents, and future developers who wish to contribute to the CLI. This pattern is common in our project, and we can discuss the possibility of removing it.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -46,36 +54,51 @@ export async function init(options: InitOptions = {}) {
...
+  // Not authenticated + --yes flag: default to keyless (fastest path, no browser)
+  if (yes) return { mode: "keyless", email: null };
+
+  const mode = await select<AuthMode>({
+    message: "How would you like to set up Clerk?",
+    choices: [
+      { name: "Continue with temporary keys (connect your account later)", value: "keyless" },
+      { name: "Log in to an existing Clerk account", value: "authenticated" },
+    ],
+  });
```
brkalow: Do we need the optionality here? Or what do you think about defaulting to keyless if unauthenticated?

rafa-thayto: makes sense

PR #81, opened 2026-03-31
fix: Next steps title color rendering

alexcarpenter:
| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-31 at 10 21 09 AM | Screenshot 2026-03-31 at 10 21 23 AM |

### DIFF
--- a/packages/cli-core/src/lib/next-steps.ts
+++ b/packages/cli-core/src/lib/next-steps.ts
@@ -1,4 +1,4 @@
-import { cyan, dim } from "./color.ts";
+import { cyan, dimNeutral } from "./color.ts";
 import { isHuman } from "../mode.ts";
 
 /**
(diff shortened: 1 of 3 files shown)


kylemac (approved):

kylemac:
yup, thats better

packages/cli-core/src/lib/next-steps.ts
```diff
@@ -7,8 +7,8 @@ import { isHuman } from "../mode.ts";
  */
 export function printNextSteps(steps: string[]): void {
   if (!isHuman() || steps.length === 0) return;
-  console.error(`\n${dim("Next steps:")}`);
+  console.log(`\n${dim("Next steps:")}`);
   for (const step of steps) {
-    console.error(`  ${cyan("\u2192")} ${step}`);
+    console.log(`  ${cyan("\u2192")} ${step}`);
```
brkalow: The use of `console.error()` is intentional, as it allows us to pipe the output without having ancillary output like this included 🤔 

We might need to think of a different approach instead of using `console.error()` however

brkalow (changes requested):

packages/cli-core/src/lib/next-steps.ts
alexcarpenter: yeah was wondering if that'd be an issue, claude thought not 😆 

alt approach https://github.com/clerk/cli/pull/81/commits/24113bca4b05452b2b9e07e91f2004c744bc9b8a

Railly (approved):

brkalow (approved):
✨

PR #87, opened 2026-04-01
feat(init): bootstrap new project

rafa-thayto:
## Summary

- When `clerk init` runs in a directory with no `package.json`, it now offers to bootstrap a new project with the user's chosen framework, install deps + Clerk SDK, pull env vars, and scaffold auth files — all in one command
- New `--starter` flag to explicitly trigger bootstrap (even in non-blank dirs, with a warning)
- After bootstrap, the user can choose **keyless mode** (skip auth, scaffold permissive middleware) or authenticate normally
- Refactors `authenticateAndLink` into `resolveAuthLabel` + linking — cleaner separation between resolving the auth display label (API key, email, or fresh login) and the linking step
- Moves the agent/prompt check before context resolution so blank directories in agent mode print the guidance message instead of a confusing "no package.json" error
- Reorders package manager choices to bun, pnpm, yarn, npm
- Fixes stale quickstart URLs in React and React Router post-instructions

### Supported frameworks
Next.js, React (Vite), Vue (Vite), React Router, Astro, Nuxt, TanStack Start

### UX flow
```
clerk init (in blank dir)
  → "No project detected. Would you like to create a new one?"
  → Pick framework (searchable)
  → Pick package manager (searchable, default order: bun, pnpm, yarn, npm)
  → Enter project name (default: my-clerk-{framework}-app)
  → Generator runs (e.g., create-next-app)
  → Dependencies installed
  → "Skip authentication and use keyless mode?"
  → Clerk SDK installed + files scaffolded
  → Next steps: cd my-clerk-next-app → bun dev
```

Or explicitly: `clerk init --starter --framework next`

### New files
- `bootstrap-registry.ts` — framework generator commands and package manager runners
- `bootstrap.ts` — interactive bootstrap orchestrator (prompts, project generation, dep install)
- `bootstrap.test.ts` — unit tests for the bootstrap flow

### Key changes to existing files
- `init/index.ts` — new `resolveProjectContext` / `handleStarter` paths, `resolveAuthLabel` extracted from `authenticateAndLink`, keyless/bootstrap wiring
- `init/context.ts` — new `hasPackageJson` helper
- `init/heuristics.ts` — `printOutro` refactored, `printKeylessInfo` added

## Test plan
- [x] Registry tests: all 7 frameworks, all 4 package managers, correct flags
- [x] Init flow tests: blank dir (human/agent), bootstrap success/decline, unrecognized dir, existing project, `--starter`, `--starter` agent mode, framework override
- [x] Manual smoke test: `cd $(mktemp -d) && clerk init --starter --framework next` with keyless mode
- [x] `bun run format` / `bun run lint` / `bun test` pass

### DIFF
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -82,6 +82,7 @@ export function createProgram() {
       ),
     )
     .option("--prompt", "Output a prompt for an AI agent to integrate Clerk")
+    .option("--starter", "Create a new project from a starter template")
     .option("-y, --yes", "Skip confirmation prompts")
     .option("--no-skills", "Skip the optional agent skills install prompt")
     .setExamples([
@@ -90,6 +91,7 @@ export function createProgram() {
         command: "clerk init --framework next",
         description: "Set up for Next.js (skips detection)",
       },
+      { command: "clerk init --starter", description: "Create a new project with Clerk" },
       { command: "clerk init --prompt", description: "Output a setup prompt for an AI agent" },
       { command: "clerk init -y", description: "Skip all confirmation prompts" },
       { command: "clerk init --no-skills", description: "Skip the agent skills install prompt" },
--- a/packages/cli-core/src/commands/init/bootstrap-registry.ts
+++ b/packages/cli-core/src/commands/init/bootstrap-registry.ts
@@ -1,0 +1,146 @@
+import type { ProjectContext } from "./frameworks/types.js";
+
+export type PackageManager = ProjectContext["packageManager"];
+
+export type BootstrapEntry = {
+  label: string;
+  dep: string;
+  defaultProjectName: string;
+  buildCommand(pm: PackageManager, projectName: string): string[];
+};
+
+const PM_RUNNERS: Record<PackageManager, string[]> = {
+  npm: ["npx"],
+  yarn: ["yarn", "dlx"],
+  pnpm: ["pnpm", "dlx"],
+  bun: ["bunx"],
+};
+
+function runner(pm: PackageManager): string[] {
+  return PM_RUNNERS[pm];
+}
+
+// WARNING: All generators are pinned to @latest. This means installs are non-deterministic
+// and upstream CLI changes can silently break bootstrap. If a generator changes its flags,
+// update the corresponding entry here and the matching test in bootstrap.test.ts.
+//
+// NOTE on package-manager forwarding: Only Nuxt (--packageManager) and TanStack (--package-manager)
+// accept an explicit PM flag. The others (create-next-app, create-vite, create-react-router,
+// create-astro) infer the PM from npm_config_user_agent, which is set automatically when run
+// via bunx/pnpm dlx/yarn dlx/npx. This works in practice since we run them via the selected
+// PM's runner, and our own installDependencies() step uses the correct PM regardless.
+export const BOOTSTRAP_REGISTRY: BootstrapEntry[] = [
+  {
+    label: "Next.js",
+    dep: "next",
+    defaultProjectName: "my-clerk-next-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-next-app@latest",
+      name,
+      "--yes",
+      "--skip-install",
+      "--disable-git",
+    ],
+  },
+  {
+    label: "React",
+    dep: "react",
+    defaultProjectName: "my-clerk-react-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-vite@latest",
+      name,
+      "--template",
+      "react-ts",
+    ],
+  },
+  {
+    label: "Vue",
+    dep: "vue",
+    defaultProjectName: "my-clerk-vue-app",
+    buildCommand: (pm, name) => [...runner(pm), "create-vite@latest", name, "--template", "vue-ts"],
+  },
+  {
+    label: "React Router",
+    dep: "react-router",
+    defaultProjectName: "my-clerk-react-router-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-react-router@latest",
+      name,
+      "--yes",
+      "--no-install",
+      "--no-git-init",
+    ],
+  },
+  {
+    label: "Astro",
+    dep: "astro",
+    defaultProjectName: "my-clerk-astro-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-astro@latest",
+      name,
+      "--yes",
+      "--no-install",
+      "--no-git",
+      "--skip-houston",
+    ],
+  },
+  {
+    label: "Nuxt",
+    dep: "nuxt",
+    defaultProjectName: "my-clerk-nuxt-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-nuxt@latest",
+      name,
+      "--template",
+      "minimal",
+      "--no-install",
+      "--no-gitInit",
+      // --no-modules suppresses the interactive modules prompt; --template minimal already
+      // implies a minimal setup, but both flags are kept for belt-and-suspenders safety.
+      "--no-modules",
+      "--packageManager",
+      pm,
+    ],
+  },
+  {
+    label: "TanStack Start",
+    dep: "@tanstack/react-start",
+    defaultProjectName: "my-clerk-tanstack-start-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "@tanstack/cli@latest",
+      "create",
+      name,
+      "--framework",
+      "react",
+      "--no-install",
+      "--no-git",
+      "--package-manager",
+      pm,
+    ],
+  },
+  {
+    label: "JavaScript",
+    dep: "vite",
+    defaultProjectName: "my-clerk-vite-app",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-vite@latest",
+      name,
+      "--template",
+      "vanilla",
+    ],
+  },
+];
+
+export const PM_INSTALL_COMMANDS: Record<PackageManager, string[]> = {
+  npm: ["npm", "install"],
+  yarn: ["yarn", "install"],
+  pnpm: ["pnpm", "install"],
+  bun: ["bun", "install"],
+};
--- a/packages/cli-core/src/commands/init/bootstrap.test.ts
+++ b/packages/cli-core/src/commands/init/bootstrap.test.ts
@@ -1,0 +1,126 @@
+import { test, expect, describe } from "bun:test";
+import { BOOTSTRAP_REGISTRY } from "./bootstrap-registry.ts";
+
+function entryFor(dep: string) {
+  const entry = BOOTSTRAP_REGISTRY.find((e) => e.dep === dep);
+  if (!entry) throw new Error(`No bootstrap entry for dep: ${dep}`);
+  return entry;
+}
+
+describe("BOOTSTRAP_REGISTRY", () => {
+  const packageManagers = ["npm", "yarn", "pnpm", "bun"] as const;
+
+  test("contains all 8 supported frameworks", () => {
+    expect(BOOTSTRAP_REGISTRY).toHaveLength(8);
+    const deps = BOOTSTRAP_REGISTRY.map((e) => e.dep);
+    expect(deps).toContain("next");
+    expect(deps).toContain("react");
+    expect(deps).toContain("vue");
+    expect(deps).toContain("react-router");
+    expect(deps).toContain("astro");
+    expect(deps).toContain("nuxt");
+    expect(deps).toContain("@tanstack/react-start");
+    expect(deps).toContain("vite");
+  });
+
+  test("each entry produces a non-empty string[] for all package managers", () => {
+    for (const entry of BOOTSTRAP_REGISTRY) {
+      for (const pm of packageManagers) {
+        const cmd = entry.buildCommand(pm, "test-app");
+        expect(cmd.length).toBeGreaterThan(0);
+        expect(cmd.every((arg) => typeof arg === "string")).toBe(true);
+      }
+    }
+  });
+
+  test("Next.js uses project name as target directory", () => {
+    const cmd = entryFor("next").buildCommand("bun", "my-cool-app");
+    expect(cmd[0]).toBe("bunx");
+    expect(cmd).toContain("create-next-app@latest");
+    expect(cmd).toContain("my-cool-app");
+    expect(cmd).toContain("--yes");
+    expect(cmd).toContain("--skip-install");
+    expect(cmd).toContain("--disable-git");
+
+    expect(entryFor("next").buildCommand("npm", "app")[0]).toBe("npx");
+  });
+
+  test("React uses project name as target directory", () => {
+    const cmd = entryFor("react").buildCommand("pnpm", "my-app");
+    expect(cmd).toContain("create-vite@latest");
+    expect(cmd).toContain("my-app");
+    expect(cmd).toContain("--template");
+    expect(cmd).toContain("react-ts");
+  });
+
+  test("Vue uses create-vite with vue-ts template", () => {
+    const cmd = entryFor("vue").buildCommand("npm", "my-vue");
+    expect(cmd).toContain("create-vite@latest");
+    expect(cmd).toContain("my-vue");
+    expect(cmd).toContain("vue-ts");
+  });
+
+  test("Nuxt uses create-nuxt with --no-modules to suppress interactive prompt", () => {
+    const cmd = entryFor("nuxt").buildCommand("bun", "my-nuxt");
+    expect(cmd).toContain("create-nuxt@latest");
+    expect(cmd).toContain("my-nuxt");
+    expect(cmd).toContain("--no-modules");
+    expect(cmd).toContain("--no-install");
+    expect(cmd).toContain("--no-gitInit");
+    expect(cmd).toContain("--packageManager");
+    expect(cmd).toContain("bun");
+  });
+
+  test("TanStack uses project name as positional arg", () => {
+    const cmd = entryFor("@tanstack/react-start").buildCommand("yarn", "my-app");
+    expect(cmd[0]).toBe("yarn");
+    expect(cmd[1]).toBe("dlx");
+    expect(cmd).toContain("@tanstack/cli@latest");
+    expect(cmd).toContain("create");
+    expect(cmd).toContain("my-app");
+    expect(cmd).toContain("--framework");
+    expect(cmd).toContain("react");
+    expect(cmd).toContain("--no-install");
+    expect(cmd).toContain("--no-git");
+  });
+
+  test("React Router uses project name as target", () => {
+    const cmd = entryFor("react-router").buildCommand("npm", "my-rr");
+    expect(cmd).toContain("create-react-router@latest");
+    expect(cmd).toContain("my-rr");
+    expect(cmd).toContain("--yes");
+    expect(cmd).toContain("--no-install");
+    expect(cmd).toContain("--no-git-init");
+  });
+
+  test("Astro uses --yes for basics template with --skip-houston", () => {
+    const cmd = entryFor("astro").buildCommand("bun", "my-astro");
+    expect(cmd).toContain("create-astro@latest");
+    expect(cmd).toContain("my-astro");
+    expect(cmd).toContain("--yes");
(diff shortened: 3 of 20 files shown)


packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -447,8 +456,10 @@ export async function scaffoldNextjsMiddleware(ctx: {
     return {
       path,
       type: "create",
-      content: nextjsMiddlewareContent(),
-      description: "Create Clerk middleware with route protection",
+      content: nextjsMiddlewareContent(ctx.keyless),
+      description: ctx.keyless
+        ? "Create Clerk middleware (permissive — keyless mode)"
```
alexcarpenter: @eatmorespinach does labeling keyless mode here align with some of the changes we made to the prompt removing `keyless` as folks don't understand what that means really?

packages/cli-core/src/commands/init/bootstrap-registry.ts
```diff
@@ -0,0 +1,122 @@
...
+      "create",
+      name,
+      "--framework",
+      "react",
+      "--no-install",
+      "--no-git",
+      "--package-manager",
+      pm,
+    ],
+  },
```
alexcarpenter, line 126: ❓ would we support JS via vite here too?

https://clerk.com/docs/js-frontend/getting-started/quickstart

packages/cli-core/src/commands/init/frameworks/helpers.ts
eatmorespinach: Since we can authenticate the user and get their real keys with the CLI, yeah  we should consider removing the "keyless" step altogether - just automatically use the user's real keys since we have file system access. 

Seems like signing up first is a known pattern when starting a CLI (eg. Vercel). thoughts @brkalow

wyattjoh (commented):
Combined review: deep local pass plus a partial Codex review (Codex runtime crashed mid-verification, so I hand-verified each claim before posting). Overall architecture is solid, tests run, lint/format are clean. Flagging the issues below as a comment review.

**Nothing to post on lines that are not in the diff, so two findings live here in the body:**

### README drift (blocks merge per project conventions)
Per `CLAUDE.md`, both command and root READMEs must stay in sync when commands change. Neither was touched on this branch:
- `packages/cli-core/src/commands/init/README.md` — Usage and Options table have no mention of `--starter`, bootstrap, or the `javascript` framework alias. The Options "Valid values" list does not include `javascript`. The Framework Detection table is missing the `vite -> JavaScript` row. The "Scaffolding is supported for the first 8 frameworks" note is stale (the map is now 11 entries), and there is no new JavaScript (Vite) scaffolding subsection.
- `README.md` (root) — the `clerk init` help block is missing the `--starter` option line and the `clerk init --starter` example. Regenerate via `bun run dev -- --help` per the project README.

### Overall verdict
Needs fixes before merge. Critical-ish: the silent-success-on-failure behavior (failed bootstraps exit 0), the `-y` gap, and the README drift. Everything else is small and localized. See inline comments for details. Nice work on the shell-out safety (every generator uses argv arrays, zero injection surface) and the clean three-file separation between flow, registry, and orchestration.

packages/cli-core/src/commands/init/bootstrap-registry.ts
```diff
@@ -0,0 +1,135 @@
...
+  {
+    label: "JavaScript",
+    dep: "vite",
+    docsUrl: "https://clerk.com/docs/js-frontend/getting-started/quickstart",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-vite@latest",
+      name,
+      "--template",
+      "vanilla-ts",
```
wyattjoh: The "JavaScript" label here generates a TypeScript project because `vanilla-ts` is Vite's TS template (the JS template is `vanilla`). Combined with `lib/framework.ts:80-85` which names this entry "JavaScript" with `@clerk/clerk-js` (the framework-free vanilla SDK) and the `javascript`/`js` aliases, a user selecting "JavaScript" will be surprised to find `.ts` files. The scaffold adapter at `frameworks/javascript.ts` already supports both via `ctx.typescript`, so switching to `vanilla` would give users what they probably expect. If TS was intentional, rename the label to avoid confusion. Note that `bootstrap.test.ts:106` will need to change alongside the fix.

packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -0,0 +1,178 @@
...
+  });
+  return name.trim();
+}
+
+async function generateProject(label: string, command: string[], cwd: string): Promise<boolean> {
+  console.log(`\nCreating ${cyan(label)} project...\n`);
+
+  const exitCode = await spawnInherited(command, cwd);
+  if (exitCode !== 0) {
+    console.log(yellow(`\nProject generation failed (exit code ${exitCode}).`));
```
wyattjoh: This (and the sibling paths at lines 54-56, 170, plus `index.ts:98, 131`) prints a yellow warning and returns null, which propagates up to `init()` and still reaches `outro("Done")`. A failed bootstrap is visually indistinguishable from success and exits with code 0, which will mislead CI scripts and agents wrapping the CLI.

Per `CLAUDE.md`, known user-facing failures should throw `CliError` so the global handler formats them and exits non-zero. The install-failure warning at lines 106-110 is fine as non-fatal since the project was still created.

packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -0,0 +1,178 @@
...
+ * Interactive bootstrap flow.
+ * When skipConfirm is true (e.g. --starter flag), skips the "create a new one?" prompt.
+ * Returns bootstrap result on success, or null on failure.
+ */
+export async function promptAndBootstrap(
+  cwd: string,
+  frameworkOverride?: FrameworkInfo,
+  { skipConfirm = false } = {},
+): Promise<BootstrapResult | null> {
+  if (!skipConfirm) {
```
wyattjoh, line 145: `-y` does not reach here. In a blank directory, `resolveProjectContext` calls `bootstrapAndDetect(cwd, fw)` at `index.ts:143` without propagating `options.yes`, so this confirm still fires even with `clerk init -y`. `confirmOverwrite` in `handleStarter` has the same problem.

`-y` is described in `cli-program.ts:79` as "Skip confirmation prompts", so propagating it into `bootstrapAndDetect` / `promptAndBootstrap` as `skipConfirm: true` and also skipping `confirmOverwrite` would honor the flag's stated behavior.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -15,91 +17,191 @@ import {
...
-  intro("clerk init");
-
-  const ctx = await withSpinner("Detecting framework...", () =>
-    gatherContext(cwd, frameworkOverride),
-  );
-
-  // Populate framework-specific context (variant, layoutPath, middlewareBasename)
-  if (ctx) await enrichProjectContext(ctx);
-
   if (options.prompt || isAgent()) {
```
wyattjoh, line 48: This early-return for agent mode happens before `handleStarter` / `resolveProjectContext` can run, which means the two `!isHuman()` branches further down (lines 108-111 and 136-141) are unreachable in production: in production, `!isHuman() <=> isAgent()`.

Separately, per `CLAUDE.md`: "Commands with interactive prompts should check `isAgent()` and, when in agent mode, output a structured prompt that an AI agent can follow." This branch expands init's interactive surface significantly (bootstrap, framework picker, PM picker, project name input, keyless prompt) without updating the agent-mode output. The preexisting one-liner is now stale relative to what the new flow can do. Worth either writing a richer agent prompt for the new paths, or filing a follow-up.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -15,91 +17,191 @@ import {
...
+async function handleStarter(
   cwd: string,
-  ctx: ProjectContext | null,
-  options: InitOptions,
-): Promise<void> {
-  if (!ctx) {
+  frameworkOverride?: FrameworkInfo,
+): Promise<ResolvedContext | null> {
+  if (!isHuman()) {
+    console.log("The --starter flag requires interactive mode.");
```
wyattjoh: Dead code. `init()` already early-returns at line 45 whenever `isAgent()` is true, and in production `!isHuman() <=> isAgent()`, so this branch is unreachable.

The corresponding test at `index.test.ts:282-291` only sets `mockIsHuman(false)` without setting `mockIsAgent(true)`, so it passes against an impossible "neither human nor agent" state. Delete both, or restructure the gating in `init()` so `--starter` can legitimately reach this branch.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -15,91 +17,191 @@ import {
...
+
+  if (!isBlank) {
     console.log(
       `Could not detect a framework. Install the appropriate Clerk SDK manually: ${dim("https://clerk.com/docs")}`,
     );
+    return null;
+  }
+
+  if (!isHuman()) {
+    console.log(
```
wyattjoh: Same dead-branch issue as line 109. Unreachable because of the agent-mode early return at line 45.

packages/cli-core/src/commands/init/index.test.ts
```diff
@@ -1,72 +1,292 @@
...
+
+    await init({ starter: true, yes: true });
+
+    expect(mockPromptAndBootstrap).toHaveBeenCalledWith(expect.any(String), undefined, {
+      skipConfirm: true,
+    });
+    expect(mockConfirmOverwrite).toHaveBeenCalledWith(expect.any(String));
+  });
+
+  test("--starter in agent mode prints guidance without bootstrap", async () => {
```
wyattjoh, line 218: This test sets `mockIsHuman(false)` but leaves `mockIsAgent()` at its default (`undefined`, falsy). That's an impossible production state: in real code `isHuman() <=> !isAgent()`. The test passes only because `handleStarter`'s `!isHuman()` branch fires while `init()`'s `isAgent()` check at line 45 sees undefined and falls through. If anyone fixes the mock defaults or the dead branch, this test starts failing for the wrong reason.

Either delete this test and the dead branch, or fix it to set `mockIsAgent(true)` and decide whether `handleStarter`'s agent branch should actually be reachable.

packages/cli-core/src/commands/init/index.test.ts
```diff
@@ -1,72 +1,292 @@
...
+  link: (...args: unknown[]) => mockLink(...args),
+}));
+
+mock.module("../env/pull.js", () => ({
+  pull: async () => {},
+}));
+
+const mockIsAgent = mock();
+const mockIsHuman = mock();
+mock.module("../../mode.js", () => ({
```
wyattjoh: This file was previously a pure `spyOn` test. Converting it to `mock.module` reintroduces a pattern the old version deliberately avoided - the deleted comment from the previous version explicitly explained: "Bun's mock.module globally replaces modules for the entire test run, which pollutes other test files (link, env/pull, config, context, etc.) that import the same modules. spyOn restores cleanly."

Per `rules/testing.md`: "In Bun 1.x, mock.module() registrations are process-lifetime and will pollute the module registry for any later test file that imports the same module via a non-mocked path... do not introduce it in test files that will run alongside files importing the real module."

This file now mocks 14 widely-shared modules including `lib/framework.js`, `lib/config.js`, `mode.js`, `./context.js`, `./scaffold.js`, `./preview.js`, `./format.js`, `./scan.js`, `./heuristics.js`, `./bootstrap.js`, `../link/index.js`, `../env/pull.js`, `../auth/login.js`, `../../lib/next-steps.js`. At least some of those are imported by other test files in the suite. Worth verifying with a full `bun run test` in a clean process to confirm nothing else regressed, and ideally sticking with the `spyOn` pattern the previous version used.

packages/cli-core/src/commands/init/bootstrap-registry.ts
```diff
@@ -0,0 +1,135 @@
+import type { ProjectContext } from "./frameworks/types.js";
+
+export type PackageManager = ProjectContext["packageManager"];
+
+export type BootstrapEntry = {
+  label: string;
+  dep: string;
+  docsUrl: string;
```
wyattjoh: `docsUrl` is populated for every entry (lines 27, 40, 52, 58, 71, 85, 102, 119) but grep confirms it is never read anywhere. The `docsUrl` used elsewhere is a separate field on `FRAMEWORK_PROMPTS` in `prompts/index.ts`. Either wire this into `printBootstrapNextSteps` (so users see a docs link after bootstrap), or drop the field.

packages/cli-core/src/commands/init/bootstrap-registry.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export const BOOTSTRAP_REGISTRY: BootstrapEntry[] = [
+  {
+    label: "Next.js",
+    dep: "next",
+    docsUrl: "https://clerk.com/docs/nextjs/getting-started/quickstart",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-next-app@latest",
```
wyattjoh, line 39: All 8 entries pin generators to `@latest`, which means user runs and CI are non-deterministic, and upstream CLI changes can silently break us. The two Nuxt fix commits on this branch (`bd660e8` and `bc66793`) are evidence this has already bitten once.

Worth at least a comment here warning future maintainers. Better: a cheap E2E smoke test (run `create-vite` in a tmpdir, gated behind a CI flag) would catch this class of regression before shipping.

packages/cli-core/src/commands/init/bootstrap-registry.ts
```diff
@@ -0,0 +1,135 @@
...
+    docsUrl: "https://clerk.com/docs/nuxt/getting-started/quickstart",
+    buildCommand: (pm, name) => [
+      ...runner(pm),
+      "create-nuxt@latest",
+      name,
+      "--template",
+      "minimal",
+      "--no-install",
+      "--no-gitInit",
+      "--no-modules",
```
wyattjoh, line 105: Passing both `--no-modules` and `--template minimal` is likely redundant since the minimal template already implies no modules. Not broken today, but if `create-nuxt` later removes `--no-modules` the test at `bootstrap.test.ts:67` will fail for the wrong reason. Consider dropping one or adding an explanatory comment.

packages/cli-core/src/commands/init/bootstrap-registry.ts
```diff
@@ -0,0 +1,135 @@
...
+      ...runner(pm),
+      "@tanstack/cli@latest",
+      "create",
+      name,
+      "--framework",
+      "react",
+      "--no-install",
+      "--no-git",
+      "--package-manager",
+      pm,
```
wyattjoh, line 124: Package-manager forwarding is inconsistent across entries: TanStack gets `--package-manager`, Nuxt gets `--packageManager`, but `create-next-app`, `create-vite`, `create-react-router`, and `create-astro` do not receive `pm` at all. Those generators fall back to `npm_config_user_agent` detection via `bunx`/`pnpm dlx`, which works today but is fragile and means a project bootstrapped with `bunx create-next-app` thinks bun is the PM while the subsequent `bun install` step is ours. Not blocking, just worth a comment.

packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -0,0 +1,178 @@
...
+
+async function pickPackageManager(): Promise<PackageManager> {
+  return search<PackageManager>({
+    message: "Which package manager?",
+    source: (term) => filterChoices(PM_CHOICES, term),
+  });
+}
+
+function defaultProjectName(entry: BootstrapEntry): string {
+  const slug = entry.dep
```
wyattjoh: For `@tanstack/react-start`, the first replace produces `my-clerk--tanstack-react-start-app` (note the double dash) which the `-+` collapse regex on line 70 then flattens. Works, but it's clever-in-a-bad-way. Consider adding an explicit `defaultProjectName` field per entry in the registry so the default is obvious at the definition site.

packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+async function askProjectName(entry: BootstrapEntry): Promise<string> {
+  const name = await input({
+    message: "Project name:",
+    default: defaultProjectName(entry),
+    validate: (value) => {
+      if (!value.trim()) return "Project name is required";
+      if (/[A-Z]/.test(value)) return "Project name must be lowercase";
+      if (/\s/.test(value)) return "Project name cannot contain spaces";
```
wyattjoh, line 77: The validator only rejects uppercase and spaces, which means `/`, `$`, `*`, `..`, and digit-only names all pass. Downstream generators will reject some of these, but upfront validation would be a better UX. At minimum, rejecting path separators (`/`, `..`) would prevent a user from accidentally creating projects in sibling directories (though `join(cwd, projectName)` at line 164 partially mitigates that).

packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -0,0 +1,178 @@
...
+  return choices.filter((c) => c.name.toLowerCase().includes(lower));
+}
+
+async function pickFramework(frameworkOverride?: FrameworkInfo): Promise<BootstrapEntry | null> {
+  if (!frameworkOverride) {
+    const chosen = await search({
+      message: "Which framework?",
+      source: (term) => filterChoices(FRAMEWORK_CHOICES, term),
+    });
+    return findEntry(chosen) ?? null;
```
wyattjoh: `?? null` is redundant: `findEntry` already returns `BootstrapEntry | undefined`, and `chosen` comes from `FRAMEWORK_CHOICES` which is derived from `BOOTSTRAP_REGISTRY`, so the lookup is guaranteed to succeed. Could collapse to `return findEntry(chosen)!;` with a comment, or leave it as-is if you want explicit null-safety.

packages/cli-core/src/commands/init/bootstrap.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  const entry = BOOTSTRAP_REGISTRY.find((e) => e.dep === dep);
+  if (!entry) throw new Error(`No bootstrap entry for dep: ${dep}`);
+  return entry;
+}
+
+describe("BOOTSTRAP_REGISTRY", () => {
+  const packageManagers = ["npm", "yarn", "pnpm", "bun"] as const;
+
+  test("contains all 8 supported frameworks", () => {
+    expect(BOOTSTRAP_REGISTRY).toHaveLength(8);
```
wyattjoh, line 14: The registry shape assertions are good, but there is no coverage for the bootstrap flow's actual failure paths. Worth adding tests for:
- Unsupported framework override (e.g. `--framework expo --starter` hitting the "not supported" branch in `bootstrap.ts:53-57`)
- `confirmOverwrite` user abort
- `promptAndBootstrap` returning null when the user declines "Would you like to create a new one?"
- `gatherContext` returning null after a successful generator run (`index.ts:96-100`)
- Generator exit-code != 0
- `hasPackageJson(projectDir)` returning false after the generator claims success

These are the paths most likely to regress silently (per the "failed bootstraps exit zero" comment on `bootstrap.ts`).

packages/cli-core/src/cli-program.ts
```diff
@@ -75,13 +75,15 @@ export function createProgram() {
       ),
     )
     .option("--prompt", "Output a prompt for an AI agent to integrate Clerk")
+    .option("--starter", "Create a new project from a starter template")
```
wyattjoh, line 85: `--starter` as a boolean flag means the non-interactive path is `clerk init --framework next --starter` (awkward), and the example at line 86 (`clerk init --starter`) silently drops into an interactive framework picker. Consider either:
- `--starter [framework]` with an optional value, so `clerk init --starter next` works directly
- Or keep boolean and document that `--framework` is required alongside for non-interactive use

Either way, this and the example at line 86 should be kept in sync with the decision.

packages/cli-core/src/commands/init/frameworks/nextjs-app.test.ts
wyattjoh, line 1: Not required for this PR (but would be neat!) is if you instead implemented this in the e2e tests! That would be a better place as it already represents a complete Next.js app. You can add additional tests there instead of having to construct them here.

packages/cli-core/src/commands/init/bootstrap.test.ts
wyattjoh, line 1: TBH I don't really see that this test adds any additional coverage. I don't expect us to be in a situation where we're deleting entries from the registry as it's already strongly typed.

wyattjoh (approved):

PR #96, opened 2026-04-02
feat: create app from interactive picker

rafa-thayto:
## Summary

- Adds a **"+ Create a new application"** option pinned at the bottom of the interactive app picker in `clerk link` (and `clerk init`, which calls `link` internally)
- When selected, prompts for an app name and creates it via `POST /v1/platform/applications`
- When no apps exist, the picker shows only the create option instead of throwing an error
- Adds `createApplication()` to the PLAPI client
- Extracts `tryDetectApp()` for flatter key-detection logic in `resolveApp`

## Test plan

- [x] Unit tests pass: `bun test packages/cli-core/src/lib/plapi.test.ts packages/cli-core/src/commands/link/index.test.ts` (71 tests)
- [x] `clerk link` → picker shows "+ Create a new application" at bottom
- [x] Selecting it prompts for name, creates app, and links
- [x] Filtering narrows apps but always keeps create option visible
- [x] `clerk init` from a new project → create option appears during link step
- [x] Empty account (no apps) → picker shows only the create option


wyattjoh:
### Code review

Reviewed with deep analysis and a parallel Codex pass. Found 5 issues:

1. New PLAPI calls in `link` skip `withApiContext`, breaking the project convention. Sibling command `apps/list.ts` wraps `listApplications()` with `withApiContext(..., "Failed to list applications")`. The new `createApplication()` and follow-up `fetchApplication()` (and the pre-existing `listApplications()` / `--app` lookup) surface failures as generic PLAPI errors instead of action-specific messages. CLAUDE.md is explicit: "Wrap API calls with `withApiContext` to attach a human-readable context string."

https://github.com/clerk/cli/blob/f3c844b686964e9615ff3b8baf07fa605f06fb68/packages/cli-core/src/commands/link/index.ts#L127-L137

2. README tells users to "create apps with `clerk apps` first" for non-interactive flows, but no such subcommand exists. The `apps/` directory only ships `list` (https://github.com/clerk/cli/blob/f3c844b686964e9615ff3b8baf07fa605f06fb68/packages/cli-core/src/commands/apps/list.ts#L1-L10). Reword to point at the Dashboard / Platform API, or add the missing subcommand.

https://github.com/clerk/cli/blob/f3c844b686964e9615ff3b8baf07fa605f06fb68/packages/cli-core/src/commands/link/README.md#L38-L43

3. `AGENT_PROMPT` lists the new `POST /v1/platform/applications` endpoint in the table but the numbered Steps prose still only describes listing and selecting an existing app. Agents reading the prompt have no narrative cue to offer "create a new app" as an alternative when the user has none, so the new behavior silently disappears in agent mode.

https://github.com/clerk/cli/blob/f3c844b686964e9615ff3b8baf07fa605f06fb68/packages/cli-core/src/commands/link/index.ts#L20-L36

4. `Created ...` is logged immediately after the POST, before `fetchApplication()` and `setProfile()` run. If either of those fail (network blip, missing dev instance, config write error), the user sees a success-looking "Created" line followed by a hard error from the same command. Move the log after the link succeeds, or downgrade it to past tense after `setProfile()` returns.

https://github.com/clerk/cli/blob/f3c844b686964e9615ff3b8baf07fa605f06fb68/packages/cli-core/src/commands/link/index.ts#L127-L137

5. Empty-name validation throws `CliError` instead of using inquirer's `validate` callback, so a user who picks "+ Create a new application" and accidentally enters whitespace gets a hard CLI exit and has to re-run `clerk link` from scratch. Pass `validate: (v) => v.trim() ? true : "Application name cannot be empty"` to `input()` so the picker re-prompts in place. The `docsUrl` on the current `CliError` also points at `quickstart/setup-clerk`, which is unrelated to the actual error.

https://github.com/clerk/cli/blob/f3c844b686964e9615ff3b8baf07fa605f06fb68/packages/cli-core/src/commands/link/index.ts#L222-L225

Test coverage gaps worth filling: no test for `createApplication()` failure or the follow-up `fetchApplication()` failure (assert `setProfile` is not called), and no agent-mode test asserting the prompt covers the create-app path.

Generated with [Claude Code](https://claude.ai/code) (ultrathink + parallel Codex review).

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

wyattjoh (commented):
A few things worth addressing:

1. New PLAPI calls (`createApplication`, `fetchApplication`, `listApplications`) skip `withApiContext` - sibling `apps/list.ts:41` shows the convention.
2. `link/README.md` tells users to "create apps with `clerk apps` first" but no such subcommand exists (only `clerk apps list`).
3. `AGENT_PROMPT` Steps prose isn't updated to mention the create-app path - only the endpoint table is.
4. `Created ...` is logged before `fetchApplication` and `setProfile` run, so a failure shows success then error.
5. Empty-name validation throws and exits the CLI - use inquirer's `validate` callback so the picker re-prompts in place. The `docsUrl` also points at an unrelated setup page.

.github/workflows/ci.yml
```diff
@@ -83,6 +83,12 @@ jobs:
       - run: bun install --frozen-lockfile
       - run: bunx playwright install chromium --with-deps
 
+      - name: Cleanup stale test users
+        run: bun test/e2e/cleanup-test-users.ts
+        env:
+          CLERK_CLI_TEST_APP_ID: ${{ secrets.CLERK_CLI_TEST_APP_ID }}
+          CLERK_PLATFORM_API_KEY: ${{ secrets.CLERK_PLATFORM_API_KEY }}
+
```
wyattjoh, line 105: maybe worth moving this to `scripts/`?

wyattjoh (approved):
Left one nit, great otherwise!

PR #97, opened 2026-04-03
refactor(log): introduce structured log utility with proper stdout/stderr semantics

alexcarpenter:
## Summary

- Introduces `packages/cli-core/src/lib/log.ts` — a structured logging utility that enforces
  stdout/stderr channel discipline across all CLI commands
- Migrates all 48 command and library files from raw `console.log`/`console.error` to the
  new `log.*` API
- Replaces the monkey-patching of `console.log`/`console.error` in `spinner.ts` with a
  clean `pushPrefix()`/`popPrefix()` mechanism backed by the same module
- Adds `withCapturedLogs()` (AsyncLocalStorage-based) for test output capture, replacing
  brittle `spyOn(console, "log")` patterns across all test files
- **Log levels** (`silent`/`error`/`warn`/`info`/`debug`) with global `--verbose` flag support
- **`log.debug()`** method gated by level — replaces ad-hoc debug closures in commands
- **Tagged child loggers** via `log.withTag("name")` — scoped, nestable, async-safe
- **Spam throttle** — deduplicates repeated stderr messages within a 1s window
- **Inline highlighting** — backtick-wrapped spans auto-highlighted in cyan

## Motivation

The CLI had no consistent output-channel contract. Every command used `console.log()` for
everything — JSON data, human status messages, error context — all landing on stdout. This
made it impossible to:

1. **Pipe machine-readable output** (`clerk api /users | jq .`) without human chrome polluting
   the stream
2. **Test output reliably** — `console.log` spies are global singletons, so concurrent async
   tests could bleed into each other
3. **Reason about what goes where** — no compile-time or review-time signal about whether a
   given print statement is data or decoration

The `spinner.ts` intro/outro flow also relied on replacing `console.log` and `console.error`
at runtime with prefixed wrappers, which broke any code that captured a reference to the
original functions.

Commands also had no consistent way to emit debug output — the `deploy` command created its
own ad-hoc `debug()` closure, and other commands had no debug logging at all. The new log
level system provides a unified `log.debug()` method gated by a global `--verbose` flag.

### `log.*` API

| Method         | Stream   | Color/prefix     | Level gate | Use for                                          |
|----------------|----------|------------------|------------|--------------------------------------------------|
| `log.data()`   | **stdout** | none           | —          | Machine-readable output: JSON, agent prompts, piped values |
| `log.info()`   | stderr   | none (dim `│` when inside intro/outro) | `info` | Human-facing status, confirmations |
| `log.success()`| stderr   | green            | `info`     | Completion messages                              |
| `log.warn()`   | stderr   | yellow           | `warn`     | Warnings                                         |
| `log.error()`  | stderr   | red, `error: ` prefix | `error` | Error messages                              |
| `log.debug()`  | stderr   | dim              | `debug`    | Verbose/diagnostic output (hidden by default)    |
| `log.blank()`  | stderr   | —                | —          | Visual spacing                                   |
| `log.raw()`    | stderr   | none, no prefix  | —          | Machine-readable stderr (agent JSON on stderr)   |

**Rule of thumb:** if a message contains color codes (`cyan()`, `dim()`) or is meant for
human eyes, it belongs on stderr via `log.info()`. If it's the command's primary data output
(JSON, an email address for scripting, an agent prompt), it belongs on stdout via `log.data()`.

### Log levels

Five levels with numeric ordering: `silent` (0) → `error` (1) → `warn` (2) → `info` (3, default) → `debug` (4).

Each `log.*` method checks `isLevelEnabled()` before writing. `log.data()`, `log.blank()`, and
`log.raw()` are never gated — stdout/raw output is always emitted regardless of level.

The global `--verbose` flag calls `setLogLevel("debug")` in the commander `preAction` hook,
enabling `log.debug()` across all commands. Commands can also call `setLogLevel()` directly
for per-command flags (e.g., `deploy --debug`).

### Tagged child loggers

`log.withTag("api")` returns a new logger that prefixes messages with a dim `[api]` tag.
Tags are nestable: `log.withTag("http").withTag("req")` → `[http:req]`.

Tagged loggers share the same level, capture, and prefix state as the root logger. Unlike
`pushPrefix()`/`popPrefix()`, they're scoped and don't require cleanup — safe for async code:

```ts
const apiLog = log.withTag("api");
apiLog.info("fetching users...");   // → [api] fetching users...
apiLog.debug("response: 200 OK");  // → [api] response: 200 OK (only if --verbose)
```

### Spam throttle

Identical stderr messages within a 1-second window are suppressed after 5 occurrences. When the
window expires, a summary is printed: `(repeated N more times)`. This prevents terminal flooding
from polling loops or retry sequences without losing information.

Throttling only applies to stderr (never stdout/data) and is disabled in test capture mode.

### Inline highlighting

Messages passed to `log.info()` (and tagged variants) auto-highlight backtick-wrapped spans
in cyan. e.g., `` log.info("Run `clerk link` to continue") `` renders `clerk link` in cyan
without manual `cyan()` wrapping at each call site.

### Intro/outro prefix

The old `patchConsole()`/`unpatchConsole()` that swapped `console.log` globally is replaced by
`pushPrefix()`/`popPrefix()` in `log.ts`. When `prefixDepth > 0`, `log.info/success/warn/error`
automatically prepend `│  ` to each line. `log.data()` is never prefixed — stdout stays clean.

### Test capture

`withCapturedLogs()` uses `AsyncLocalStorage` to intercept `log.*` calls within a scope.
Concurrent async tests get isolated capture buffers with no global state mutation:

```ts
import { captureLog } from "../../test/stubs.ts";

const captured = captureLog();

test("outputs the email to stdout", async () => {
  await captured.run(() => whoami());
  expect(captured.out).toContain("user@example.com");
});

test("status messages go to stderr", async () => {
  await captured.run(() => login());
  expect(captured.err).toContain("Logged in as");
});
```

The `captureLog()` helper (in `test/stubs.ts`) wraps this with convenience getters:
- `captured.out` — joined stdout lines
- `captured.err` — joined stderr lines
- `captured.run(fn)` — execute `fn` inside the capture scope

## Files changed

- **New:** `packages/cli-core/src/lib/log.ts` — the logging utility (log levels, tagged loggers, throttle, highlighting)
- **New:** `packages/cli-core/src/lib/log.test.ts` — async isolation, nesting, log level, tag, and highlighting tests
- **Modified:** `packages/cli-core/src/lib/spinner.ts` — removed 35-line console monkey-patch,
  replaced with `pushPrefix()`/`popPrefix()` imports
- **Modified:** `packages/cli-core/src/cli-program.ts` — wired `--verbose` flag to `setLogLevel("debug")`
- **Modified:** `packages/cli-core/src/commands/deploy/index.ts` — replaced ad-hoc `debug()` with `log.debug()`
- **Modified:** `packages/cli-core/src/test/stubs.ts` — added `captureLog()` helper
- **Modified:** 44 command/test files — migrated from `console.log`/`console.error` + spy-based
  assertions to `log.*` + `captureLog()`

## Follow-up ideas

- Add a lint rule to flag `console.log`/`console.error` in command files, suggesting `log.*`

## Test plan

- [x] `bun test packages/cli-core/src/lib/log.test.ts` — capture isolation, nesting, log levels, tags, highlighting (12 tests)
- [x] `bun test packages/cli-core/src/commands/deploy/index.test.ts` — deploy with log.debug (7 tests)
- [x] `bun test packages/cli-core/src/commands/auth/login.test.ts` — login output on stderr
- [x] `bun test packages/cli-core/src/commands/link/index.test.ts` — link output on stderr, agent prompt on stdout
- [x] `bun test` — 745 pass, 2 pre-existing failures (unrelated api test flake)
- [x] `bun run lint` — clean

## Summary by CodeRabbit

* **Refactor**
  * Unified CLI output through a centralized logging system for improved consistency and debugging across all commands.

* **Documentation**
  * Added logging guidelines for developers to standardize output handling.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -84,6 +84,73 @@ When creating or modifying a command, evaluate whether it needs an agent mode. C
 
 `README.md` at the project root contains the CLI help output. When commands are added, removed, or their options change, update the help output in `README.md` to stay in sync. You can regenerate it by running `bun run dev -- --help`.
 
+## Logging
+
+All output goes through the `log` object from `packages/cli-core/src/lib/log.ts`. **Never use `console.log`, `console.error`, or `process.stderr.write` directly** — use `log.*` methods so output respects log levels, throttling, and test capture.
+
+```ts
+import { log } from "../../lib/log.ts";
+```
+
+### Which method to use
+
+| Method          | Stream     | When to use                                     |
+| --------------- | ---------- | ----------------------------------------------- |
+| `log.data()`    | **stdout** | Pipeable output (JSON, lists, machine-readable) |
+| `log.info()`    | stderr     | Status messages                                 |
+| `log.success()` | stderr     | Completion confirmations (green)                |
+| `log.warn()`    | stderr     | Warnings (yellow)                               |
+| `log.error()`   | stderr     | Errors (red, auto-prefixed `error: `)           |
+| `log.debug()`   | stderr     | Diagnostic info, only with `--verbose`          |
+| `log.raw()`     | stderr     | Machine-readable JSON for agent mode            |
+| `log.blank()`   | stderr     | Blank line                                      |
+
+`log.data()` writes to **stdout** — this is what gets piped (e.g., `clerk apps list | jq`). Everything else writes to **stderr** as UI for humans. Never mix these.
+
+### Debug logging
+
+`log.debug()` is gated by `--verbose`. Use for diagnostic details (request URLs, timing, intermediate state):
+
+```ts
+log.debug(`Fetching instance ${instanceId}…`);
+```
+
+### Tagged loggers
+
+`log.withTag()` adds scoped context in complex flows:
+
+```ts
+const apiLog = log.withTag("api");
+apiLog.info("Fetching config…"); // [api] Fetching config…
+```
+
+### Inline highlighting
+
+Backtick-wrapped text auto-highlights in cyan:
+
+```ts
+log.info("Linked to `my-app` on `development`");
+```
+
+### Testing log output
+
+Use `captureLog()` from `packages/cli-core/src/test/lib/stubs.ts`:
+
+```ts
+import { captureLog } from "../../test/lib/stubs.ts";
+
+let captured: ReturnType<typeof captureLog>;
+beforeEach(() => {
+  captured = captureLog();
+});
+
+test("outputs result", async () => {
+  await captured.run(() => myCommand());
+  expect(captured.out).toContain("expected stdout"); // log.data()
+  expect(captured.err).toContain("expected stderr"); // log.info/warn/etc.
+});
+```
+
 ## Error Handling
 
 All error classes and helpers live in `packages/cli-core/src/lib/errors.ts`. The global error handler in `packages/cli-core/src/cli.ts` catches thrown errors and formats them for the user. **Never call `console.error` + `process.exit` directly in commands** — throw an error instead and let the global handler deal with output and exit codes.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -134,17 +134,13 @@ log.info("Linked to `my-app` on `development`");
 
 ### Testing log output
 
-Use `captureLog()` from `packages/cli-core/src/test/lib/stubs.ts`:
+Use `captureLog()` from `packages/cli-core/src/test/lib/stubs.ts`. Capture is scoped via `AsyncLocalStorage` — no teardown needed:
 
 ```ts
 import { captureLog } from "../../test/lib/stubs.ts";
 
-let captured: ReturnType<typeof captureLog>;
-beforeEach(() => {
-  captured = captureLog();
-});
-
 test("outputs result", async () => {
+  const captured = captureLog();
   await captured.run(() => myCommand());
   expect(captured.out).toContain("expected stdout"); // log.data()
   expect(captured.err).toContain("expected stderr"); // log.info/warn/etc.
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -1,4 +1,5 @@
 import { Command, createOption, createArgument } from "@commander-js/extra-typings";
+import { setLogLevel } from "./lib/log.ts";
 import { setMode, type Mode } from "./mode.ts";
 import { init } from "./commands/init/index.ts";
 import { login } from "./commands/auth/login.ts";
@@ -48,10 +49,13 @@ export function createProgram() {
       "--mode <mode>",
       "Force interaction mode (human or agent). Defaults to auto-detect based on TTY.",
     )
-    .option("--verbose", "Show detailed error output");
+    .option("--verbose", "Show detailed output (enables debug messages)");
 
   program.hook("preAction", async () => {
     const opts = program.opts();
+    if (opts.verbose) {
+      setLogLevel("debug");
+    }
     if (opts.mode) {
       if (opts.mode !== "human" && opts.mode !== "agent") {
         throwUsageError(`Invalid mode "${opts.mode}". Must be "human" or "agent".`);
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -52,6 +52,9 @@ export function createProgram() {
     .option("--verbose", "Show detailed output (enables debug messages)");
 
   program.hook("preAction", async () => {
+    // Reset log level at the start of each command invocation so a previous
+    // --verbose or --debug flag doesn't leak into subsequent runs.
+    setLogLevel("info");
     const opts = program.opts();
     if (opts.verbose) {
       setLogLevel("debug");
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -29,8 +29,8 @@ import {
 } from "./lib/errors.ts";
 import { clerkHelpConfig } from "./lib/help.ts";
 import { ExitPromptError } from "@inquirer/core";
-import { red } from "./lib/color.ts";
 import { isAgent } from "./mode.ts";
+import { log } from "./lib/log.ts";
 
 export function createProgram() {
   const program = new Command()
@@ -535,10 +535,10 @@ export async function runProgram(
         outputJsonError(error.code, error.message, error.docsUrl);
       } else {
         if (error.message) {
-          console.error(red(`error: ${error.message}`));
+          log.error(error.message);
         }
         if (error.docsUrl) {
-          console.error(`\nFor more information, see: ${error.docsUrl}`);
+          log.info(`\nFor more information, see: ${error.docsUrl}`);
         }
       }
       process.exit(error.exitCode);
@@ -557,9 +557,9 @@ export async function runProgram(
           apiErrors,
         );
       } else {
-        console.error(red(`error: ${prefix} (${error.status}): ${detail}`));
+        log.error(`${prefix} (${error.status}): ${detail}`);
         if (verbose && error instanceof PlapiError && error.url) {
-          console.error(`       URL: ${error.url}`);
+          log.error(`       URL: ${error.url}`);
         }
       }
       process.exit(EXIT_CODE.GENERAL);
@@ -569,7 +569,7 @@ export async function runProgram(
       if (isAgent()) {
         outputJsonError("unexpected_error", error.message);
       } else {
-        console.error(red(`error: ${error.message}`));
+        log.error(error.message);
       }
       process.exit(EXIT_CODE.GENERAL);
     }
@@ -577,7 +577,7 @@ export async function runProgram(
     if (isAgent()) {
       outputJsonError("unexpected_error", "An unexpected error occurred");
     } else {
-      console.error(red("error: An unexpected error occurred"));
+      log.error("An unexpected error occurred");
     }
     process.exit(EXIT_CODE.GENERAL);
   }
@@ -608,7 +608,7 @@ function outputJsonError(
   };
   if (docsUrl) payload.error.docsUrl = docsUrl;
   if (errors?.length) payload.error.errors = errors;
-  console.error(JSON.stringify(payload));
+  log.raw(JSON.stringify(payload));
 }
 
 /** Extract the error code from a Clerk API JSON response body, if present. */
--- a/packages/cli-core/src/commands/apps/list.ts
+++ b/packages/cli-core/src/commands/apps/list.ts
@@ -32,7 +32,7 @@ export async function list(options: AppsOptions = {}): Promise<void> {
   if (printJson(result.map(stripSecrets), options)) return;
 
   if (result.length === 0) {
-    log.data("No applications found. Create one at https://dashboard.clerk.com");
+    log.warn("No applications found. Create one at https://dashboard.clerk.com");
     return;
   }
 
--- a/packages/cli-core/src/commands/apps/list.ts
+++ b/packages/cli-core/src/commands/apps/list.ts
@@ -3,6 +3,7 @@ import { withApiContext } from "../../lib/errors.ts";
 import { dim, cyan } from "../../lib/color.ts";
 import { withSpinner } from "../../lib/spinner.ts";
 import { stripSecrets, displayName, printJson, type AppsOptions } from "./shared.ts";
+import { log } from "../../lib/log.ts";
 
 const COLUMN_PADDING = 2;
 
@@ -13,13 +14,13 @@ function formatAppsTable(apps: Application[]): void {
     Math.max("APP ID".length, ...apps.map((a) => a.application_id.length)) + COLUMN_PADDING;
 
   const header = `${"NAME".padEnd(nameWidth)}${"APP ID".padEnd(idWidth)}ENVIRONMENTS`;
-  console.log(dim(header));
+  log.data(dim(header));
 
   for (const app of apps) {
     const name = displayName(app).padEnd(nameWidth);
     const id = dim(app.application_id.padEnd(idWidth));
     const envs = app.instances.map((i) => i.environment_type).join(", ");
-    console.log(`${cyan(name)}${id}${envs}`);
+    log.data(`${cyan(name)}${id}${envs}`);
   }
 }
 
@@ -31,12 +32,12 @@ export async function list(options: AppsOptions = {}): Promise<void> {
   if (printJson(result.map(stripSecrets), options)) return;
 
   if (result.length === 0) {
-    console.log("No applications found. Create one at https://dashboard.clerk.com");
+    log.data("No applications found. Create one at https://dashboard.clerk.com");
     return;
   }
 
   formatAppsTable(result);
 
   const count = result.length;
-  console.error(`\n${count} application${count === 1 ? "" : "s"}`);
+  log.info(`\n${count} application${count === 1 ? "" : "s"}`);
 }
--- a/packages/cli-core/src/commands/auth/login.test.ts
+++ b/packages/cli-core/src/commands/auth/login.test.ts
@@ -97,6 +97,10 @@ describe("login", () => {
     }
   });
 
+  function runLogin(options?: Parameters<typeof login>[0]) {
+    return captured.run(() => login(options));
+  }
+
   function mockBunSpawn() {
     try {
       (Bun as any).spawn = mock(() => ({ exited: Promise.resolve(0) }));
@@ -114,7 +118,7 @@ describe("login", () => {
     });
 
     consoleSpy = spyOn(console, "log").mockImplementation(() => {});
-    const result = await login();
+    const result = await runLogin();
 
     expect(result).toEqual({ userId: "user_123", email: "existing@example.com" });
     expect(captured.out).toContain("Logged in as existing@example.com");
@@ -145,7 +149,7 @@ describe("login", () => {
     mockSetAuth.mockResolvedValue(undefined);
 
     consoleSpy = spyOn(console, "log").mockImplementation(() => {});
-    const result = await login();
+    const result = await runLogin();
 
     expect(result).toEqual({ userId: "user_new", email: "new@example.com" });
     expect(mockStartAuthServer).toHaveBeenCalledWith("test-state-value");
@@ -183,7 +187,7 @@ describe("login", () => {
     mockSetAuth.mockResolvedValue(undefined);
 
     consoleSpy = spyOn(console, "log").mockImplementation(() => {});
-    const result = await login();
+    const result = await runLogin();
 
     expect(result).toEqual({ userId: "user_refreshed", email: "refreshed@example.com" });
     expect(mockStartAuthServer).toHaveBeenCalled();
@@ -205,7 +209,7 @@ describe("login", () => {
 
     consoleSpy = spyOn(console, "log").mockImplementation(() => {});
 
-    await expect(login()).rejects.toThrow("Authentication timed out");
+    await expect(runLogin()).rejects.toThrow("Authentication timed out");
     expect(mockServer.stop).toHaveBeenCalled();
   });
 
@@ -234,7 +238,7 @@ describe("login", () => {
     mockSetAuth.mockResolvedValue(undefined);
 
     consoleSpy = spyOn(console, "log").mockImplementation(() => {});
-    const result = await login();
+    const result = await runLogin();
 
     expect(result).toEqual({ userId: "user_brand_new", email: "brandnew@example.com" });
     expect(mockStartAuthServer).toHaveBeenCalled();
@@ -250,7 +254,7 @@ describe("login", () => {
     });
 
     consoleSpy = spyOn(console, "log").mockImplementation(() => {});
-    const result = await login();
+    const result = await runLogin();
 
     expect(result).toEqual({ userId: "user_123", email: "agent@example.com" });
     expect(captured.out).toContain("Logged in as agent@example.com");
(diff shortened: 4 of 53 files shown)


packages/cli-core/src/commands/whoami/index.ts
```diff
@@ -1,19 +1,20 @@
...
     return;
   }
 
   try {
     const userInfo = await withSpinner("Fetching account info...", () => fetchUserInfo(token));
-    console.log(userInfo.email);
+    log.data(userInfo.email);
   } catch {
-    console.log("Session expired. Run `clerk auth login` to re-authenticate");
+    log.data("Session expired. Run `clerk auth login` to re-authenticate");
```
Railly: These end up on stdout, so `EMAIL=$(clerk whoami)` would capture the error message as the email value. Also exits 0 on both paths.

packages/cli-core/src/commands/doctor/index.ts
Railly: `log.info()` writes to stderr while the spinner is animating on the same stream, first check result gets appended to the spinner line without a newline:
```
◐  Running...✓ Check 1
```

wyattjoh:
### Code review

Found 2 issues:

1. Integration test `auth-lifecycle` asserts "Not logged in" on stdout, but commit `e884a7c` (on this branch) routes that message through `CliError` to stderr. Both `human` and `agent` `test.each` variants will fail since `clerk()` (strict mode) throws on the resulting non-zero exit before the stdout assertion is even reached.

https://github.com/clerk/cli/blob/e884a7cec7cefca02e1dc5a3a84294f07c70d0fd/packages/cli-core/src/test/integration/auth-lifecycle.test.ts#L13-L33

2. `outputJsonError` writes agent JSON via `log.raw`, which goes through `writeln` -> `shouldWrite` and is subject to the stderr spam throttle. If the same JSON error fires twice within `THROTTLE_WINDOW_MS` (1s), the duplicate is silently dropped, breaking the machine-readable contract for `--mode agent` consumers. Either bypass `shouldWrite` for `log.raw`, or treat it as data-channel output.

https://github.com/clerk/cli/blob/e884a7cec7cefca02e1dc5a3a84294f07c70d0fd/packages/cli-core/src/cli-program.ts#L585-L592

https://github.com/clerk/cli/blob/e884a7cec7cefca02e1dc5a3a84294f07c70d0fd/packages/cli-core/src/lib/log.ts#L78-L142

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with thumbs up. Otherwise, react with thumbs down.</sub>

packages/cli-core/src/lib/log.ts
```diff
@@ -0,0 +1,211 @@
...
+
+let lastLogKey = "";
+let lastLogCount = 0;
+let lastLogTime = 0;
+let throttleTimer: ReturnType<typeof setTimeout> | undefined;
+
+function flushThrottle() {
+  if (lastLogCount > THROTTLE_MIN) {
+    const repeated = lastLogCount - THROTTLE_MIN;
+    writeln(
```
wyattjoh, line 70: **Stack overflow on throttle flush.** `flushThrottle()` emits its summary via `writeln()`, which re-enters `shouldWrite()`. Because `throttleTimer` and the last-message state are cleared *after* the write, `shouldWrite()` re-enters the "flush pending throttle summary first" branch and recursively calls `flushThrottle()` until the stack overflows.

Repro: log the same stderr line 6 times — 5 print, then `RangeError: Maximum call stack size exceeded`.

Fix: reset throttle state before emitting the summary, or write the summary directly to the stream without going back through `shouldWrite()`. Worth adding a throttle test (`log.test.ts` covers levels/capture/tags/highlight but nothing in the throttle path, which is why this slipped through).

packages/cli-core/src/cli-program.ts
```diff
@@ -583,7 +587,7 @@ function outputJsonError(
   };
   if (docsUrl) payload.error.docsUrl = docsUrl;
   if (errors?.length) payload.error.errors = errors;
-  console.error(JSON.stringify(payload));
+  log.raw(JSON.stringify(payload));
```
wyattjoh, line 618: **Throttle can swallow agent JSON errors.** `log.raw` goes through `writeln` → `shouldWrite`, which throttles duplicate stderr messages within `THROTTLE_WINDOW_MS`. If two identical error payloads emit within ~1s in `--mode agent`, the second is silently dropped — breaking the machine-readable contract. Either bypass `shouldWrite` from `log.raw`, or emit agent JSON via a dedicated path that skips throttling.

packages/cli-core/src/lib/log.ts
```diff
@@ -0,0 +1,211 @@
...
+  return true;
+}
+
+// ── Inline highlighting ──────────────────────────────────────────────────
+
+/**
+ * Highlights `backtick` spans in cyan within a message.
+ */
+function highlight(msg: string): string {
+  return msg.replace(/`([^`]+)`/g, (_, content) => cyan(`\`${content}\``));
```
wyattjoh: **Inline highlighting breaks surrounding ANSI styles.** `highlight()` runs after callers apply formatting, so injecting `cyan(...)\x1b[0m` resets the outer style. Visible in `doctor/format.ts:23`: after the first highlighted backtick segment, the rest of the dimmed remedy is no longer dim. Either highlight before applying outer styling, or preserve and reapply the active ANSI state.

packages/cli-core/src/lib/log.ts
```diff
@@ -0,0 +1,211 @@
...
+
+const LEVEL_VALUE: Record<LogLevel, number> = {
+  silent: 0,
+  error: 1,
+  warn: 2,
+  info: 3,
+  debug: 4,
+};
+
+let currentLevel: LogLevel = "info";
```
wyattjoh, line 16: **Log level is process-global and only ever raised.** `cli-program.ts:49` and `commands/deploy/index.ts:92` both bump it to `debug` and never restore. Any in-process caller that runs once with `--verbose` (or `deploy --debug`) keeps debug logging on for later invocations. Reset to `info` at the start of each run, or scope log level to an invocation context.

packages/cli-core/src/lib/log.ts
```diff
@@ -0,0 +1,211 @@
...
+// ── Tagged child logger ──────────────────────────────────────────────────
+
+export type Logger = typeof log & {
+  withTag(tag: string): Logger;
+};
+
+function createLogger(tag?: string): Logger {
+  function formatTag(msg: string): string {
+    if (!tag) return msg;
+    return `${dim(`[${tag}]`)} ${msg}`;
```
wyattjoh: Nit: tagged colored logs lose their outer color after the dim `[tag]`. `log.withTag("api").error("broken")` renders red `error: `, dim `[api]`, then plain `broken` because the inner `dim()` reset cancels the outer red. Reapply the outer color after the tag if tagged warnings/errors are meant to stay fully colored.

packages/cli-core/src/lib/log.ts
```diff
@@ -0,0 +1,211 @@
...
+      if (!isLevelEnabled("debug")) return;
+      writeln(process.stderr, "stderr", applyPrefix(dim(formatTag(msg))));
+    },
+    /** Blank line to stderr. */
+    blank() {
+      const captured = captureStorage.getStore();
+      if (captured) {
+        captured.stderr.push("");
+      } else {
+        process.stderr.write("\n");
```
wyattjoh: Nit / question: `log.blank()` no longer preserves the `intro()`/`outro()` pipe — inside a prefixed flow it writes a literal blank line instead of `│`. `doctor.printResults()` now leaves empty gaps inside bracketed output. Intentional?

wyattjoh (commented):
Combined findings from a Claude + Codex pass on this branch. Two correctness blockers, a couple of design regressions, and a couple of nits — leaving as comments rather than blocking, happy to dig into any of these.

**Out-of-diff finding (can't anchor inline):** `packages/cli-core/src/test/integration/auth-lifecycle.test.ts` lines 13-33 still asserts `"Not logged in"` on `stdout`, but `e884a7c` routes that message through `CliError` to `stderr`. The test uses the strict `clerk()` helper which throws on non-zero exit, so both `human` and `agent` `test.each` variants fail. Switch to `clerk.raw()` (or equivalent) and assert on `stderr`.

packages/cli-core/src/commands/whoami/index.ts
alexcarpenter: Already addressed in e884a7c

packages/cli-core/src/commands/doctor/index.ts
alexcarpenter: Also addressed in e884a7c — the spinner clears the line before returning from `withSpinner`, so `printResults()` starts on a fresh line. Let me know if you can still repro after that commit.

packages/cli-core/src/lib/log.ts
alexcarpenter: Fixed in cec5895 — throttle state is now reset *before* the `writeln()` call in `flushThrottle()`, breaking the re-entrancy cycle. Added a test for the raw/throttle bypass in `log.test.ts`.

packages/cli-core/src/cli-program.ts
alexcarpenter: Fixed in cec5895 — `log.raw()` now bypasses `writeln`/`shouldWrite` entirely, writing directly to the stream (or capture store in tests). Added a dedicated test asserting duplicate raw payloads are never dropped.

packages/cli-core/src/lib/log.ts
alexcarpenter: Fixed in cec5895 — `highlight()` now uses targeted foreground color codes (\`\\x1b[36m\`/\`\\x1b[39m\`) instead of `cyan()` which emits a full \`\\x1b[0m\` reset. Added a test asserting no full reset appears in the highlight region.

packages/cli-core/src/lib/log.ts
alexcarpenter: Fixed in cec5895 — log level is now reset to `"info"` at the top of the `preAction` hook, before checking `--verbose`. Each command invocation starts clean.

packages/cli-core/src/lib/log.ts
alexcarpenter: Fixed in cec5895 — `formatTag()` now uses targeted dim on/off (\`\\x1b[2m\`/\`\\x1b[22m\`) instead of `dim()` which emits a full reset. The outer color survives the tag. Added a test asserting no full reset between the tag and message text.

packages/cli-core/src/lib/log.ts
alexcarpenter: Not intentional — fixed in cec5895. `blank()` now routes through `applyPrefix("")` which returns the dim bar when `prefixDepth > 0`. Added a test asserting the pipe character appears when inside an intro/outro flow.

alexcarpenter:
Re: the out-of-diff finding on `auth-lifecycle.test.ts` — fixed in cec5895. Switched the not-logged-in cases to `clerk.raw()` and asserted on `stderr` + non-zero exit code.

Railly:
@alexcarpenter do you find useful to document a bit how to use this for agents (AGENST.md / CLAUDE.md) so they can pick the rules automatically?

packages/cli-core/src/commands/auth/login.test.ts
```diff
@@ -331,12 +341,8 @@ describe("login", () => {
...
 
-    await login({ showNextSteps: false });
+    await runLogin({ showNextSteps: false });
 
-    expect(
-      consoleErrorSpy.mock.calls.some(
-        (c) => typeof c[0] === "string" && (c[0] as string).includes("Next steps:"),
-      ),
-    ).toBe(false);
+    expect(captured.err).not.toContain("Next steps:");
```
jfoshee, line 346: so much better

packages/cli-core/src/commands/deploy/index.ts
```diff
@@ -85,40 +86,41 @@ Refer to the Clerk Platform API docs for detailed request/response schemas.`;
...
-  const debug = options.debug ? (...args: unknown[]) => console.log("[debug]", ...args) : () => {};
+  if (options.debug) {
+    const { setLogLevel } = await import("../../lib/log.ts");
+    setLogLevel("debug");
+  }
 
-  console.log(
-    yellow("[mock] This command uses mocked data and is not yet wired up to real APIs.") + "\n",
-  );
+  log.data("[mock] This command uses mocked data and is not yet wired up to real APIs.\n");
```
jfoshee, line 98: I've seen a couple places where we use these square bracket prefixes. `[mock]`, `[dry-run]`, ...? Maybe that should be another feature of the `log` for a future PR.

packages/cli-core/src/commands/init/heuristics.ts
```diff
@@ -100,15 +95,15 @@ function formatAction(action: FileAction): string {
...
 
   for (const action of plan.actions) {
-    console.log(formatAction(action));
+    log.info(formatAction(action));
   }
 
   printNextSteps(plan.postInstructions);
   printFindings(findings);
-  console.log();
+  log.blank();
```
jfoshee, line 106: I guess this blank line is meant to skip a line after the info above, so maybe it should be `log.info("")`?

packages/cli-core/src/commands/init/preview.ts
```diff
@@ -14,20 +15,20 @@ function formatAction(action: FileAction): string {
...
-    console.log(dim("\n  After scaffolding, you'll need to:"));
+    log.info(dim("\n  After scaffolding, you'll need to:"));
     for (const instr of plan.postInstructions) {
-      console.log(dim(`  • ${instr}`));
+      log.info(dim(`  • ${instr}`));
     }
   }
 
-  console.log();
+  log.blank();
```
jfoshee, line 31: again, should this be log.info("")?

packages/cli-core/src/lib/log.test.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    withCapturedLogs(cap, () => {
+      log.debug("hidden");
+      log.info("hidden");
+      log.warn("visible");
+      log.error("visible");
+    });
+
+    expect(cap.stderr.length).toBe(2);
+  });
```
jfoshee, line 120: More specifically,
```suggestion
      log.warn("a warning");
      log.error("an error");
    });

    expect(cap.stderr).toEqual(["a warning", "an error"]);
```

packages/cli-core/src/lib/log.test.ts
```diff
@@ -0,0 +1,276 @@
...
+    });
+
+    expect(cap.stdout).toEqual(["always visible"]);
+  });
+});
+
+describe("withTag", () => {
+  test("prefixes messages with dim tag", () => {
+    const cap = createCapture();
+    const tagged = log.withTag("api");
```
jfoshee, line 151: oh now I see you do support this! we can circle back to use apply `withTag` where it would be useful.

CLAUDE.md
```diff
@@ -84,6 +84,73 @@ When creating or modifying a command, evaluate whether it needs an agent mode. C
...
+Use `captureLog()` from `packages/cli-core/src/test/lib/stubs.ts`:
+
+```ts
+import { captureLog } from "../../test/lib/stubs.ts";
+
+let captured: ReturnType<typeof captureLog>;
+beforeEach(() => {
+  captured = captureLog();
+});
+
```
jfoshee: Is it required that we tear down the log?

jfoshee (approved):
Obviously in the future we should have some linter rule around using console.log

packages/cli-core/src/commands/apps/list.ts
```diff
@@ -42,17 +43,17 @@ export async function list(options: AppsListOptions = {}): Promise<void> {
...
 
   if (options.json || isAgent()) {
-    console.log(JSON.stringify(stripSecrets(result), null, 2));
+    log.data(JSON.stringify(stripSecrets(result), null, 2));
     return;
   }
 
   if (result.length === 0) {
-    console.log("No applications found. Create one at https://dashboard.clerk.com");
+    log.data("No applications found. Create one at https://dashboard.clerk.com");
```
wyattjoh:
```suggestion
    log.warn("No applications found. Create one at https://dashboard.clerk.com");
```

packages/cli-core/src/commands/auth/logout.ts
```diff
@@ -1,8 +1,9 @@
 import { deleteToken } from "../../lib/credential-store.ts";
 import { clearAuth } from "../../lib/config.ts";
+import { log } from "../../lib/log.ts";
 
 export async function logout(): Promise<void> {
   await deleteToken();
   await clearAuth();
-  console.log("Logged out successfully");
+  log.data("Logged out successfully");
```
wyattjoh:
```suggestion
  log.success("Logged out successfully");
```

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -93,7 +94,7 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
   const existingSession = await withSpinner("Checking session...", () => getExistingSession());
 
   if (existingSession && !isHuman()) {
-    console.log(`Logged in as ${existingSession.email}`);
+    log.info(`Logged in as ${existingSession.email}`);
```
wyattjoh:
```suggestion
    log.success(`Logged in as ${existingSession.email}`);
```

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -111,7 +112,7 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
   const userInfo = await performOAuthFlow();
 
   bar();
-  console.log(`Logged in as ${userInfo.email}`);
+  log.info(`Logged in as ${userInfo.email}`);
```
wyattjoh:
```suggestion
  log.success(`Logged in as ${userInfo.email}`);
```

packages/cli-core/src/commands/config/pull.ts
```diff
@@ -26,8 +27,8 @@ export async function configPull(options: ConfigPullOptions): Promise<void> {
 
   if (options.output) {
     await Bun.write(options.output, json + "\n");
-    console.error(`Config written to ${options.output}`);
+    log.info(`Config written to ${options.output}`);
```
wyattjoh:
```suggestion
    log.success(`Config written to ${options.output}`);
```

packages/cli-core/src/commands/config/push.ts
```diff
@@ -116,8 +117,8 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
         "Failed to push config",
       ),
   );
-  console.log(JSON.stringify(result, null, 2));
-  console.error("Config pushed successfully");
+  log.data(JSON.stringify(result, null, 2));
+  log.info("Config pushed successfully");
```
wyattjoh:
```suggestion
  log.success("Config pushed successfully");
```

packages/cli-core/src/commands/config/schema.ts
```diff
@@ -24,8 +24,8 @@ export async function configSchema(options: ConfigSchemaOptions): Promise<void>
 
   if (options.output) {
     await Bun.write(options.output, json + "\n");
-    console.error(`Schema written to ${options.output}`);
+    log.info(`Schema written to ${options.output}`);
```
wyattjoh:
```suggestion
    log.success(`Schema written to ${options.output}`);
```

packages/cli-core/src/commands/doctor/index.ts
```diff
@@ -42,17 +43,15 @@ async function runChecks(ctx: DoctorContext, options: DoctorOptions): Promise<Ch
...
+  for (const result of results) {
+    if (!options.spotlight || result.status !== "pass") {
+      log.info(formatCheckResult(result, options.verbose ?? false));
     }
-    console.log("");
   }
-
-  return results;
+  log.blank();
 }
```
wyattjoh, line 55: Would be neat to do this so the results can stream in as they're ready:

```ts
async function runChecks(ctx: DoctorContext, options: DoctorOptions): Promise<CheckResult[]> {
  return Promise.all(
    CHECKS.map(async (check) => {
      try {
        return await check(ctx);
      } catch (error) {
        return {
          name: "Unknown check",
          status: "fail" as const,
          message: `Check crashed: ${errorMessage(error)}`,
        };
      }
    }).map(async (promise) => {
      const result = await promise;
      if (!options.spotlight || result.status !== "pass") {
        log.info(formatCheckResult(result, options.verbose ?? false));
      }
      return result;
    }),
  );
}

```

packages/cli-core/src/commands/link/index.ts
```diff
@@ -108,7 +109,7 @@ export async function link(options: LinkOptions = {}): Promise<void> {
   });
 
   const label = app.name || app.application_id;
-  console.log(`Linked to ${cyan(label)} in ${dim(displayPath)}`);
+  log.info(`Linked to ${cyan(label)} in ${dim(displayPath)}`);
```
wyattjoh:
```suggestion
  log.success(`Linked to ${cyan(label)} in ${dim(displayPath)}`);
```

packages/cli-core/src/commands/switch-env/index.ts
```diff
@@ -16,15 +16,16 @@ import {
...
 export async function switchEnv(environment: string | undefined): Promise<void> {
   const available = getAvailableEnvs();
 
   // No argument: print current environment
   if (!environment) {
     const current = getCurrentEnvName();
-    console.log(`Current environment: ${current}`);
-    console.log(`Available environments: ${available.join(", ")}`);
+    log.data(`Current environment: ${current}`);
+    log.data(`Available environments: ${available.join(", ")}`);
```
wyattjoh: Thinking these could be info instead of data?

packages/cli-core/src/commands/deploy/index.ts
alexcarpenter: handled withTag will handle in future PR

packages/cli-core/src/commands/switch-env/index.ts
alexcarpenter:
```suggestion
    log.info(`Current environment: ${current}`);
    log.info(`Available environments: ${available.join(", ")}`);
```

PR #115, opened 2026-04-07
ci: schedule biweekly e2e fixture refresh workflow

wyattjoh:
## Summary
- Add `.github/workflows/refresh-fixtures.yml` that runs on the 1st and 15th of each month (plus `workflow_dispatch`) to regenerate E2E fixtures and open a PR with any drift.
- Uses `peter-evans/create-pull-request` with `CLERK_COOKIE_PAT` so the resulting PR triggers normal CI (the default `GITHUB_TOKEN` cannot).
- Runs on the repo's standard Blacksmith runner; checks out `main` explicitly so manual dispatch from any branch still refreshes against main; PR-creation action pinned by SHA.

## Prereq
`CLERK_COOKIE_PAT` must be available to `clerk/cli` as a repo or org secret before the first scheduled run. Until then, scheduled runs will fail at the PR-creation step.

## Test plan
- [ ] Merge and wait for a scheduled run, or trigger manually via the Actions UI after the secret is in place.
- [ ] Verify the opened refresh PR triggers the existing ``test-e2e`` job.


.github/workflows/refresh-fixtures.yml
```diff
@@ -0,0 +1,54 @@
...
+  refresh:
+    runs-on: blacksmith-2vcpu-ubuntu-2404
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          # Always refresh from main, even on workflow_dispatch runs that
+          # select a different ref, so manual triggers cannot smuggle
+          # unrelated changes from a feature branch into the refresh PR.
+          ref: main
```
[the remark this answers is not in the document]
wyattjoh: Addressed in b06154f. Added `if: github.ref == 'refs/heads/main'` at the job level (line 23) plus the rationale comment on lines 19-22. The dispatchable workflow now no-ops on any non-main ref, so a modified YAML on a feature branch cannot reach the `CLERK_COOKIE_PAT` step. Went with the job-level guard rather than splitting into a reusable workflow since it is a single privileged step and the guard is sufficient on its own.

.github/workflows/refresh-fixtures.yml
```diff
@@ -0,0 +1,59 @@
...
+      - uses: oven-sh/setup-bun@v2
+
+      - run: bun install --frozen-lockfile
+
+      - name: Refresh fixtures
+        run: bun run e2e:refresh-fixtures
+
+      - name: Create or update pull request
+        # Pinned to v8.1.0 by SHA because this step handles the CLERK_COOKIE_PAT.
+        uses: peter-evans/create-pull-request@c0f553fe549906ede9cf27b5156039d195d2ece0 # v8.1.0
```
jfoshee, line 43: Thanks for being security conscious. 👍 
Is the PAT only scoped to this repo?

jfoshee (approved):

.github/workflows/refresh-fixtures.yml
wyattjoh: Correct!

PR #116, opened 2026-04-07
fix(init): handle skill installer failures and stub in tests

wyattjoh:
## Summary

- Wrap the `Bun.spawn(["npx", "skills", "add", ...])` call inside `installSkills` in try/catch so a missing `npx` binary on a stripped-down CI image can't crash init *after* the scaffold has already succeeded.
- Add an `installSkills` spy to `init/index.test.ts`. Without it, running the unit test would shell out to the real `npx skills add` and pollute the working tree with `.augment/`, `.claude/`, `skills/clerk-*`, etc. on every test run.
- Document the existing `--no-skills` flag in both the root README and `init/README.md`.

## Test plan

- [ ] `bun run test` passes (57 passed)
- [ ] `bun run format` and `bun run lint` (cli-core) clean
- [ ] Manual: run `bun test packages/cli-core/src/commands/init/index.test.ts` and confirm no stray `skills/clerk-*` directories are written

### DIFF
--- a/packages/cli-core/src/commands/init/index.test.ts
+++ b/packages/cli-core/src/commands/init/index.test.ts
@@ -14,6 +14,9 @@ import * as previewMod from "./preview.ts";
 import * as formatMod from "./format.ts";
 import * as scanMod from "./scan.ts";
 import * as heuristics from "./heuristics.ts";
+// installSkills must be stubbed — without this spy, init() would shell out
+// to a real `npx skills add` and write real files into the test cwd.
+import * as skillsMod from "./skills.ts";
 import { init } from "./index.ts";
 
 describe("init command", () => {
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/commands/init/index.test.ts
```diff
@@ -14,6 +14,9 @@ import * as previewMod from "./preview.ts";
 import * as formatMod from "./format.ts";
 import * as scanMod from "./scan.ts";
 import * as heuristics from "./heuristics.ts";
+// installSkills must be stubbed — without this spy, init() would shell out
+// to a real `npx skills add` and write real files into the test cwd.
```
rafa-thayto:
```suggestion
```

rafa-thayto (approved):

Railly (approved):

PR #117, opened 2026-04-07
refactor(init): short-circuit on fully-clean re-run

wyattjoh:
## Summary

- When `clerk init` runs against a project where Clerk is already fully set up (supported framework, SDK installed, every scaffold action is a skip, no postInstructions), init used to print "Clerk is already set up" inside `scaffoldAndWrite` but then kept running env pull, the skills prompt, and the outro. The message lied about init completing.
- Refactor `detectAndInstall` and `scaffoldAndWrite` to return `{ alreadySetUp: boolean }`. `init()` owns the green message and short-circuits the entire post-scaffold flow when `alreadySetUp` is true.
- The "framework detected but unsupported" path (Express/Fastify/Expo) still falls through because postInstructions are useful guidance the user should see every time.
- Add a test that stubs a Next.js context with an empty scaffold plan and asserts `pull`, `printKeylessInfo`, and `installSkills` are NOT called.

Stacked on #116.

## Test plan

- [ ] `bun run test` passes (57 passed)
- [ ] New test `short-circuits env pull and skills install when already set up` passes
- [ ] Manual: run `clerk init` twice in a scaffolded Next.js project and confirm the second run exits after the "already set up" message without prompting for skills

### DIFF
--- a/packages/cli-core/src/commands/init/index.test.ts
+++ b/packages/cli-core/src/commands/init/index.test.ts
@@ -71,4 +71,29 @@ describe("init command", () => {
     expect(pullMod.pull).toHaveBeenCalled();
     expect(heuristics.printKeylessInfo).not.toHaveBeenCalled();
   });
+
+  test("short-circuits env pull and skills install when already set up", async () => {
+    setup({ email: "test@test.com" });
+
+    // Override the default null gatherContext with a minimal supported-framework
+    // context so detectAndInstall reaches the alreadySetUp path.
+    spies.find((s) => s === (context.gatherContext as unknown))?.mockRestore();
+    spyOn(context, "gatherContext").mockResolvedValue({
+      cwd: "/tmp/fake",
+      framework: { name: "Next.js", dep: "next", sdk: "@clerk/nextjs", publishableKeyEnv: "x" },
+      deps: { next: "15.0.0" },
+      packageManager: "bun",
+      typescript: true,
+      srcDir: false,
+      existingClerk: true,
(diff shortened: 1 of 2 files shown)


wyattjoh:
**Stack: init-already-set-up**

- **#117 👈 this PR**

packages/cli-core/src/commands/init/index.test.ts
```diff
@@ -71,4 +71,29 @@ describe("init command", () => {
     expect(pullMod.pull).toHaveBeenCalled();
     expect(heuristics.printKeylessInfo).not.toHaveBeenCalled();
   });
+
+  test("short-circuits env pull and skills install when already set up", async () => {
+    setup({ email: "test@test.com" });
+
+    // Override the default null gatherContext with a minimal supported-framework
+    // context so detectAndInstall reaches the alreadySetUp path.
+    spies.find((s) => s === (context.gatherContext as unknown))?.mockRestore();
```
Railly: nit: this find-by-identity feels a bit fragile. Cleaner if `setup() `returned the `gatherContext` spy as a named handle so you could `.mockResolvedValueOnce(...)` here directly. non-blocking.

Railly (approved):
One broader thought, not for this PR: `setup()` is getting heavy (19 spies) and the tests are drifting toward asserting on implementation rather than behavior. Might be worth a follow-up to split the helper by concern, or lean on a small integration test against a `tmpdir` fixture. 

LGTM

PR #118, opened 2026-04-07
feat(lib): add package runner detection helpers

wyattjoh:
## Summary

- Add `lib/runners.ts` with a `Runner` type, `RUNNERS` constant (bunx, npx, pnpm dlx, yarn dlx), `detectAvailableRunners()` (uses `Bun.which`), and `preferredRunner()` (picks the runner matching the project's package manager).
- Add a `select<T>` wrapper to `lib/prompts.ts` mirroring the existing `confirm` wrapper for piped-stdin safety.
- Use the new helpers in `init/skills.ts`: `installSkills` now picks a runner instead of hardcoding `npx`, prompts the user to choose when multiple runners are available (with the project's pm runner labelled `(detected)` as the default), and threads `packageManager` through from `init/index.ts`.

Originally stacked on #116, now rebased onto `main` after #116 merged.

## Test plan

- [ ] `bun run test` passes (59 passed)
- [ ] `bun test packages/cli-core/src/lib/runners.test.ts` passes (16 tests covering RUNNERS shape, runnerCommand, preferredRunner, detectAvailableRunners)
- [ ] Manual: on a bun-only machine (no `npx` on PATH), run `clerk init` and confirm `installSkills` falls back to `bunx` instead of failing

### DIFF
--- a/packages/cli-core/src/commands/init/skills.ts
+++ b/packages/cli-core/src/commands/init/skills.ts
@@ -11,8 +11,9 @@
  * so it runs unattended with global scope and auto-detected agents.
  */
 
-import { dim, cyan, yellow } from "../../lib/color.js";
+import { dim } from "../../lib/color.js";
 import { isHuman } from "../../mode.js";
+import { log } from "../../lib/log.js";
 import { confirm, select } from "../../lib/prompts.js";
 import {
   type Runner,
@@ -88,11 +89,10 @@ export async function installSkills(
   const available = detectAvailableRunners();
   if (available.length === 0) {
     const suggested = runnerForPackageManager(packageManager);
-    console.log(
-      yellow(
-        "\nNo package runner found on PATH (looked for bunx, npx, pnpm, yarn). " +
-          `Install one and run \`${suggested.display} skills add ${SKILLS_SOURCE}\` manually.`,
-      ),
+    log.blank();
+    log.warn(
+      "No package runner found on PATH (looked for bunx, npx, pnpm, yarn). " +
+        `Install one and run \`${suggested.display} skills add ${SKILLS_SOURCE}\` manually.`,
     );
     return;
   }
@@ -103,10 +103,9 @@ export async function installSkills(
     // preferredRunner should always find something. This guards against any
     // future change that decouples the two.
     const suggested = runnerForPackageManager(packageManager);
-    console.log(
-      yellow(
-        `\nCould not select a package runner. Run \`${suggested.display} skills add ${SKILLS_SOURCE}\` manually.`,
-      ),
+    log.blank();
+    log.warn(
+      `Could not select a package runner. Run \`${suggested.display} skills add ${SKILLS_SOURCE}\` manually.`,
     );
     return;
   }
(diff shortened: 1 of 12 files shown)


wyattjoh:
**Stack: init-skills**

- **#118 👈 this PR**
  - #124
    - #125
      - #126
        - #127

packages/cli-core/src/commands/init/skills.ts
```diff
@@ -67,38 +84,69 @@ export async function installSkills(
...
+  let runner = preferred;
+  if (isHuman() && !skipPrompt && available.length > 1) {
+    runner = await select<Runner>({
+      message: "Which package runner should install the skills?",
+      choices: available.map((r) => ({
+        name: r.id === preferred.id ? `${r.display} ${dim("(detected)")}` : r.display,
+        value: r,
+      })),
+      default: preferred,
+    });
```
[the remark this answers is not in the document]
wyattjoh: This is already handled at the top level. `cli-program.ts:509` catches `ExitPromptError` and treats it identically to `UserAbortError` (clean exit, code 0):

```ts
if (error instanceof UserAbortError || error instanceof ExitPromptError) {
  process.exit(EXIT_CODE.SUCCESS);
}
```

Wrapping the `select<Runner>(...)` call locally would be dead code, and inconsistent with every other prompt site in the codebase, including the `confirm()` call directly above it on line 80. Not changing.

packages/cli-core/src/commands/init/skills.ts
```diff
@@ -67,38 +84,69 @@ export async function installSkills(
     if (!install) return;
   }
 
-  console.log(`\nInstalling skills: ${cyan(skillList)}`);
+  // Detect runners after the user accepts — no point probing PATH if they decline.
+  const available = detectAvailableRunners();
+  if (available.length === 0) {
+    const suggested = runnerForPackageManager(packageManager);
+    console.log(
```
jfoshee: Can you update the console.log commands to use the new `log`?
I think we will need to add a lint rule.

Railly:
@wyattjoh could you update the console.logs with the log utility Carp created? 👀

packages/cli-core/src/lib/runners.test.ts
```diff
@@ -0,0 +1,226 @@
...
+  });
+
+  test("excludes yarn when `yarn dlx --help` exits non-zero (Yarn Classic)", () => {
+    mockWhich(new Set(["bunx", "npx", "pnpm", "yarn"]));
+    mockSpawnSync(1);
+    const result = detectAvailableRunners();
+    expect(result.map((r) => r.id)).toEqual(["bunx", "npx", "pnpm"]);
+  });
+
+  test("includes yarn when `yarn dlx --help` exits 0 (Yarn Berry)", () => {
```
jfoshee, line 210: I did not know about these yarn flavors!

packages/cli-core/src/lib/runners.ts
```diff
@@ -0,0 +1,138 @@
...
+  readonly binary: string;
+  readonly prefixArgs: readonly string[];
+  readonly display: string;
+};
+
+/**
+ * Known runners in preference order. When no project package manager is
+ * provided, the first available runner from this list wins.
+ */
+export const RUNNERS: readonly Runner[] = [
```
jfoshee: Minor: might be more clear if this was named `KNOWN_RUNNERS`

jfoshee (approved):
I like how this is implemented and tested. I was thinking it would be pretty painful to do a real integration test of this, so I think mocking `which` and `spawnSync` is a good strategy. That said, you point out they may not always be mockable.

So, something to consider as time marches on, this might be one of those times we should add a level of indirection. If there are a handful of system/shell functions that we use, we could wrap them up into an interface like:

```
interface ShellCommands {
  which(string): string;
  ...
}
```

And that would have a trivial `bun` implementation. And would also be trivially mocked without fear the tests aren't running on some platforms.

PR #120, opened 2026-04-07
fix(deps): forward-port playwright patch and add check-patches guard

wyattjoh:
## Summary

- Forward-ports the local `playwright-core` patch from 1.58.2 to 1.59.1. Dependabot clerk/cli#109 bumped playwright without updating the `patchedDependencies` key, so Bun silently stopped applying the workaround for [oven-sh/bun#6892](https://github.com/oven-sh/bun/issues/6892). The upstream file at the patched lines is byte-identical in 1.59.1; this re-keys the patch and renames the file.
- Adds `scripts/check-patches.ts`, a small checker that would have caught clerk/cli#109 on the Dependabot PR. It verifies for every `patchedDependencies` entry that the patch file exists, the declared version matches what's installed in `node_modules`, and the hunks are currently applied byte-for-byte to the installed file (via `git apply --reverse --check`). It also sweeps `patches/*.patch` for files no entry references.
- Wires `bun run check:patches` into the existing `Test` CI job so it runs on every PR, including Dependabot's (the `test-e2e` job intentionally skips Dependabot because it cannot access secrets, which is the reason clerk/cli#109 slipped through).
- Updates the test runner pattern to pick up `scripts/**/*.test.ts`. As a side effect this enables `scripts/sign-macos.test.ts`, which was previously orphaned.

## Why a single bundled PR

The checker is motivated by exactly the regression the patch fix addresses, and the two together make the current main branch safe to merge from Dependabot again. Reviewing them together makes the connection obvious.

## Test plan

- [ ] `bun run format:check` clean
- [ ] `bun run lint` no new errors (3 pre-existing warnings in `next-steps.ts`, `spinner.ts`, `config/push.test.ts` are unchanged)
- [ ] `bun run check:patches` prints `✓ checked 1 patch: all consistent and applied`
- [ ] `bun run test` passes all test files
- [ ] Local E2E run against the patched `playwright-core` passes (`bun run test:e2e:op -- --filter react` confirmed auth flow through `_parseSetCookieHeader`)
- [ ] Synthetic drift scenario (manually edit `patchedDependencies` key to a wrong version) correctly fails with exit 1 and a message naming both versions
- [ ] Synthetic orphan scenario (`touch patches/orphan.test.patch`) correctly fails with exit 1 and a message naming the file
- [ ] Synthetic content-drift scenario (revert one patched line in `node_modules/playwright-core/lib/server/fetch.js`) correctly fails with exit 1 and includes the raw `git apply` diagnostic

## Notes for reviewers

- **`GIT_DIR=/dev/null` env override in Check 4**: without this, `git apply --reverse --check` discovers the parent project's `.git` when run from inside `node_modules/<pkg>/` and operates in the project's repo context, where `node_modules/` is gitignored. The check would then pass vacuously even when the file was corrupted. This was caught during manual synthetic verification, not by the happy-path test, which is why an isolated regression test with a minimal real unified diff was added alongside the fix.
- **Why not also add the check to `test-e2e`?** The `Test` job is a strict superset of when this check needs to fire (Dependabot PRs run it, `test-e2e` doesn't). Duplicating would slow CI without catching more cases.
- **Why not a `postinstall`/`prepare` hook?** Out of scope for this change. Easy to add later if drift turns out to happen outside of Dependabot flows.


patches/playwright-core@1.58.2.patch
```diff
@@ -1,11 +1,3 @@
-# Workaround for a Bun/Node divergence: Bun populates IncomingMessage.url on
-# client-side responses (with the request path or redirect target), whereas
-# Node leaves it as "". Playwright's `response.url || url.toString()` fallback
-# therefore picks up a relative path under Bun and crashes in `new URL(...)`
-# inside `_parseSetCookieHeader`. Tracked upstream at:
-#   https://github.com/oven-sh/bun/issues/6892
-# Once Bun stops leaking a non-empty `url` onto client IncomingMessages, this
-# patch can be removed.
```
Railly, line 8: was it intentionally deleted?

wyattjoh: Nope, that was not intentional 😢

Railly (approved):

PR #121, opened 2026-04-07
fix(init): write env vars to existing .env instead of creating .env.local

rafa-thayto:
## Summary

- Fixes `clerk init` creating a new `.env.local` when the project already has a `.env` file
- Expands env file detection to cover the full Next.js/Vite dev load order: `.env.development.local` → `.env.local` → `.env.development` → `.env`
- Passes `ctx.envFile` to the internal `env pull` call so both the scaffolding write and the API key write target the same file

## Env file selection priority (new behavior)

Matches the official Next.js/Vite load order for `NODE_ENV=development`:

| Priority | File | Notes |
|---|---|---|
| 1 (highest) | `.env.development.local` | Always gitignored; local dev overrides |
| 2 | `.env.local` | Always gitignored; local overrides |
| 3 | `.env.development` | Dev-specific shared config |
| 4 | `.env` | Base defaults |
| fallback | `framework.envFile` | Created when none of the above exist |

## Changes

- **`lib/dotenv.ts`** — adds `ENV_FILE_CANDIDATES` constant and `findExistingEnvFile(cwd, fallback)` shared utility
- **`init/context.ts`** — uses `findExistingEnvFile` for scaffolding (all 4 candidates)
- **`env/pull.ts`** — adds `.env.development.local` as highest-priority check for secrets; intentionally skips `.env`/`.env.development` to avoid writing secrets to potentially git-tracked files

## Test plan

- [ ] `bun test packages/cli-core/src/commands/init/` — all pass
- [ ] `bun test packages/cli-core/src/commands/env/pull.test.ts` — all pass
- [ ] Next.js project with only `.env`: env vars written to `.env`, no `.env.local` created
- [ ] Next.js project with `.env.development.local`: env vars written to `.env.development.local`
- [ ] `.env.development.local` takes priority over `.env.local` when both exist
- [ ] Vite project with no env files: `.env.local` created (framework preference)

### DIFF
--- a/packages/cli-core/src/commands/env/pull.ts
+++ b/packages/cli-core/src/commands/env/pull.ts
@@ -1,12 +1,7 @@
 import { join } from "node:path";
 import { resolveAppContext } from "../../lib/config.ts";
 import { fetchApplication } from "../../lib/plapi.ts";
-import {
-  parseEnvFile,
-  mergeEnvVars,
-  serializeEnvFile,
-  ENV_FILE_CANDIDATES,
-} from "../../lib/dotenv.ts";
+import { parseEnvFile, mergeEnvVars, serializeEnvFile } from "../../lib/dotenv.ts";
 import {
   detectPublishableKeyName,
   detectSecretKeyName,
@@ -15,6 +10,8 @@ import {
 import { CliError, ERROR_CODE, withApiContext } from "../../lib/errors.ts";
 import { withSpinner } from "../../lib/spinner.ts";
 import { dim } from "../../lib/color.ts";
+
+const DEV_LOCAL_ENV_FILE = ".env.development.local";
 
 interface EnvPullOptions {
   app?: string;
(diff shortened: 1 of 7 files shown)


packages/cli-core/src/commands/env/pull.ts
```diff
@@ -32,6 +37,11 @@ async function resolveTargetFile(
 ): Promise<string> {
   if (flag) return join(cwd, flag);
 
+  // .env.development.local is the highest-priority file in the Next.js/Vite dev load order
+  // and is always gitignored, so it's safe for secrets.
+  const devLocal = join(cwd, ENV_FILE_CANDIDATES[0]);
+  if (await Bun.file(devLocal).exists()) return devLocal;
```
wyattjoh: Seems strange that the preferred file isn't actually the top file checked

wyattjoh (approved):

PR #123, opened 2026-04-07
feat(lib): add cross-platform browser launcher

wyattjoh:
## Problem

`auth/login.ts` and `deploy/index.ts` both invoked the browser via a hardcoded `Bun.spawn(["open", url])`. That binary only exists on macOS. On headless Linux (and WSL) the spawn silently fails, which for the `clerk auth login` flow means the user never reaches the OAuth callback and the login hangs until it times out. On Windows it's just broken.

## Fix

New `packages/cli-core/src/lib/open.ts` helper exporting a single `openBrowser(url)` function that returns a discriminated `OpenResult`:

```ts
type OpenResult =
  | { ok: true; launcher: string }
  | { ok: false; reason: "no-launcher" | "spawn-failed" };
```

It picks per-platform candidate launchers and probes them with `Bun.which`:

| Platform | Candidates (in order) |
|---|---|
| macOS | `open` |
| Windows / Cygwin | `start` (invoked via `cmd.exe /c start "" <url>`, since `start` is a cmd builtin and isn't on PATH) |
| Linux | `wslview`, `xdg-open`, `gnome-open`, `kde-open`, `sensible-browser` |
| Other Unix | `xdg-open` |

`wslview` is checked first on Linux because inside WSL `xdg-open` usually fails or opens nothing, whereas `wslview` correctly hands the URL to the Windows host browser.

The helper is fire-and-forget (`proc.exited` is not awaited, with a no-op `.catch()` so a non-zero launcher exit doesn't surface as an unhandled rejection), so the CLI can exit immediately after launch.

Both call sites now consume the `OpenResult`:

- `auth/login.ts` prints a formatted fallback block with the URL and failure reason if `ok` is false. Critical because the OAuth callback can't complete without the user reaching the URL.
- `deploy/index.ts` prints a dim one-line hint with the URL on failure.

## Tests

Existing login tests cover the call site via the global `Bun.spawn` mock; full unit suite is green.

### DIFF
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -10,11 +10,8 @@ import { isHuman } from "../../mode.ts";
 import { throwUserAbort } from "../../lib/errors.ts";
 import { intro, outro, bar, withSpinner } from "../../lib/spinner.ts";
 import { NEXT_STEPS } from "../../lib/next-steps.ts";
-
-const BROWSER_COMMANDS: Partial<Record<NodeJS.Platform, string>> = {
-  darwin: "open",
-  win32: "start",
-};
+import { openBrowser } from "../../lib/open.ts";
+import { cyan, dim } from "../../lib/color.ts";
 
 interface LoginOptions {
   showNextSteps?: boolean;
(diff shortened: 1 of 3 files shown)


wyattjoh:
**Stack: browser-launcher**

- **#123 👈 this PR**

Railly (approved):

PR #124, opened 2026-04-07
refactor(init): use runner detection in format.ts

wyattjoh:
## Summary

- Replaces hardcoded `["npx", "prettier", ...]` arrays in `init/format.ts` with the runner detection from `lib/runners.ts`. A bun project with no `npx` installed (e.g. Bun via Homebrew but no Node) now falls back to `bunx` instead of silently failing to format.
- Wraps the `Bun.spawn` call in try/catch, formatting is best-effort and shouldn't break init even if the runner subprocess crashes.
- Signature change: `runFormatters` now takes the full `ProjectContext` instead of just `cwd`, so it can read `packageManager`. Updates the lone caller in `init/index.ts`.

Stacked on #118.

## Test plan

- [ ] `bun run test` passes (59 passed)
- [ ] Manual: on a bun-only machine without `npx`, run `clerk init` in a Prettier-using project and confirm formatting still runs (via `bunx`)

### DIFF
--- a/packages/cli-core/src/commands/init/format.test.ts
+++ b/packages/cli-core/src/commands/init/format.test.ts
@@ -1,0 +1,256 @@
+import { test, expect, describe, beforeEach, afterEach } from "bun:test";
+import { join } from "node:path";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { runFormatters } from "./format.ts";
+import type { ProjectContext } from "./frameworks/types.js";
+
+type SpawnArgs = readonly string[];
+
+const origSpawn = Bun.spawn;
+const origWhich = Bun.which;
+const origSpawnSync = Bun.spawnSync;
+
+type SpawnImpl = (
+  cmd: SpawnArgs,
+  opts?: { cwd?: string; stdout?: unknown; stderr?: unknown },
+) => { exited: Promise<number> };
+
+function setSpawn(impl: SpawnImpl) {
+  try {
+    (Bun as unknown as { spawn: SpawnImpl }).spawn = impl;
+  } catch {
+    // Bun.spawn may not be writable on some runtimes
+  }
+}
+function restoreSpawn() {
+  try {
+    (Bun as unknown as { spawn: typeof Bun.spawn }).spawn = origSpawn;
+  } catch {
+    // ignore
+  }
+}
+
+function mockWhich(present: ReadonlySet<string>) {
+  try {
+    (Bun as unknown as { which: (bin: string) => string | null }).which = (bin) =>
+      present.has(bin) ? `/usr/local/bin/${bin}` : null;
+  } catch {
+    // ignore
+  }
+}
+function restoreWhich() {
+  try {
+    (Bun as unknown as { which: typeof Bun.which }).which = origWhich;
+  } catch {
+    // ignore
+  }
+}
+
+function mockSpawnSync(yarnDlxExitCode: number) {
+  try {
+    (Bun as unknown as { spawnSync: (cmd: string[]) => { exitCode: number } }).spawnSync = (
+      cmd,
+    ) => {
+      if (cmd[0] === "yarn" && cmd[1] === "dlx") return { exitCode: yarnDlxExitCode };
+      return { exitCode: 0 };
+    };
+  } catch {
+    // ignore
+  }
+}
+function restoreSpawnSync() {
+  try {
+    (Bun as unknown as { spawnSync: typeof Bun.spawnSync }).spawnSync = origSpawnSync;
+  } catch {
+    // ignore
+  }
+}
+
+/** Minimal ProjectContext suitable for driving runFormatters. */
+function makeCtx(overrides: Partial<ProjectContext> & { cwd: string }): ProjectContext {
+  return {
+    framework: {
+      name: "next",
+      sdk: "@clerk/nextjs",
+      dep: "next",
+      envFile: ".env.local",
+    } as ProjectContext["framework"],
+    typescript: true,
+    srcDir: false,
+    packageManager: "bun",
+    existingClerk: false,
+    deps: {},
+    envFile: ".env.local",
+    ...overrides,
+  };
+}
+
+describe("runFormatters", () => {
+  let tempDir: string;
+  let spawnCalls: SpawnArgs[];
+
+  beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


wyattjoh:
**Stack: init-skills**

- **#124 👈 this PR**
  - #125
    - #126
      - #127

rafa-thayto (dismissed):

packages/cli-core/src/commands/init/format.test.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+function restoreSpawn() {
+  try {
+    (Bun as unknown as { spawn: typeof Bun.spawn }).spawn = origSpawn;
+  } catch {
+    // ignore
+  }
+}
+
+function mockWhich(present: ReadonlySet<string>) {
```
rafa-thayto, line 34: ### Silent mock failures make tests untrustworthy

The try/catch blocks silently swallow mock-assignment failures. If Bun makes these properties non-writable in a future version, every test runs against real globals with confusing results rather than failing fast.

Same issue in `mockWhich` (line 42) and `mockSpawnSync` (line 55) and all three `restore*` functions.

**Suggestion — drop all try/catch blocks, verify the assignment took effect:**

```ts
function setSpawn(impl: SpawnImpl) {
  (Bun as unknown as { spawn: SpawnImpl }).spawn = impl;
  if (Bun.spawn !== (impl as unknown)) {
    throw new Error("Failed to mock Bun.spawn — property may be non-writable");
  }
}

function restoreSpawn() {
  (Bun as unknown as { spawn: typeof Bun.spawn }).spawn = origSpawn;
}
```

**Bonus — extract the cast once to reduce boilerplate** (repeated 6× as-is):

```ts
const bunOverrides = Bun as unknown as {
  spawn: SpawnImpl;
  which: (bin: string) => string | null;
  spawnSync: (cmd: string[]) => { exitCode: number };
};

function setSpawn(impl: SpawnImpl) {
  bunOverrides.spawn = impl;
}
function restoreSpawn() {
  bunOverrides.spawn = origSpawn as unknown as SpawnImpl;
}
// ... etc
```

packages/cli-core/src/commands/init/format.ts
```diff
@@ -1,35 +1,58 @@
 import { readDeps } from "./context.js";
+// Pulls in the same runner detection skills.ts uses, so a bun project with
+// no `npx` on PATH (entirely possible if the user installed Bun via Homebrew
+// but never installed Node) will fall back to bunx instead of silently failing.
+import { detectAvailableRunners, preferredRunner, runnerCommand } from "../../lib/runners.js";
```
rafa-thayto, line 5: This reads like commit-message rationale rather than a code comment. The function-level docstring already covers the best-effort / silent-failure contract, and the import names (`detectAvailableRunners`, `preferredRunner`) are self-documenting.

Suggestion: delete this comment block entirely — the PR description is the right place for the "why we're switching" context.

rafa-thayto (commented):
Some comments

packages/cli-core/src/commands/init/format.ts
```diff
@@ -1,35 +1,58 @@
...
+ * Format scaffolded files with prettier or biome (whichever the project uses).
+ *
+ * Best-effort: failures are silent (stdio ignored, spawn errors swallowed)
+ * because formatting is purely cosmetic and shouldn't break init.
+ */
+export async function runFormatters(ctx: ProjectContext, files: string[]): Promise<void> {
   if (files.length === 0) return;
 
-  const deps = await readDeps(cwd);
+  const deps = ctx.deps && Object.keys(ctx.deps).length > 0 ? ctx.deps : await readDeps(ctx.cwd);
```
rafa-thayto, line 34: Two things on this line:

**1. `ctx.deps &&` is dead code** — `deps` is typed `Record<string, string>` (not optional), and `gatherContext` always sets it to `deps ?? {}`. An empty object is truthy, so the `&&` never short-circuits. Simplify to:

```ts
const deps = Object.keys(ctx.deps).length > 0 ? ctx.deps : await readDeps(ctx.cwd);
```

**2. The disk fallback itself is redundant in practice** — `gatherContext()` already calls `readDeps(cwd)` and stores the result. If it returned `null`, `ctx.deps` becomes `{}`. Re-calling `readDeps` here will return the same `null`. The only scenario this fallback helps is when someone constructs a `ProjectContext` manually with `deps: {}` while a real `package.json` exists — which only happens in the test for this very code path.

Consider simplifying to just trust `ctx.deps`:

```ts
export async function runFormatters(ctx: ProjectContext, files: string[]): Promise<void> {
  if (files.length === 0) return;
  if (Object.keys(ctx.deps).length === 0) return;

  const matchingFormatters = FORMATTERS.filter((f) => f.pkg in ctx.deps);
  ...
```

Then update the `"reads deps from disk when ctx.deps is empty"` test to pass `deps: { prettier: "3.0.0" }` directly, and drop the `"no-op when ctx.deps is empty and package.json is missing"` test (already covered by `"no-op when no supported formatter is in deps"`).

packages/cli-core/src/commands/init/format.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    });
+    const ctx = makeCtx({
+      cwd: tempDir,
+      packageManager: "bun",
+      deps: { prettier: "3.0.0" },
+    });
+    await runFormatters(ctx, ["x.ts"]);
+    expect(seenCwd).toBe(tempDir);
+  });
+});
```
rafa-thayto, line 256: ### Missing test: non-zero formatter exit code

Tests cover spawn _throwing_ (swallowed correctly) and spawn succeeding (exit 0), but there's no test for a formatter _exiting_ non-zero. The current code silently ignores exit codes (`await proc.exited` with no check), which is the right best-effort behavior — worth locking in with a test:

```ts
test("ignores non-zero exit code from formatter (best-effort)", async () => {
  setSpawn((cmd) => {
    spawnCalls.push(cmd);
    return { exited: Promise.resolve(1) };
  });
  const ctx = makeCtx({
    cwd: tempDir,
    packageManager: "bun",
    deps: { prettier: "3.0.0", "@biomejs/biome": "1.9.0" },
  });
  await runFormatters(ctx, ["x.ts"]);
  // Both formatters attempted despite prettier exiting non-zero
  expect(spawnCalls).toHaveLength(2);
});
```

PR #125, opened 2026-04-07
feat(lib): add NonEmptyArray helper, tighten preferredRunner

wyattjoh:
## Summary

- Adds `lib/helpers/arrays.ts` with a `NonEmptyArray<T>` type alias and an `isNonEmpty<T>` type-guard predicate.
- Tightens `preferredRunner` to require a `NonEmptyArray<Runner>` input and return `Runner` (no longer `| undefined`).
- Updates the two call sites (`init/skills.ts`, `init/format.ts`) to use `isNonEmpty` for the non-empty check, and drops the now-provably-unreachable defensive `if (!preferred)` branch in `skills.ts`. The type system enforces the invariant that runtime code previously checked.

Stacked on #124.

## Test plan

- [ ] `bun run test` passes (60 passed)
- [ ] `bun test packages/cli-core/src/lib/helpers/arrays.test.ts` covers the type guard

PR #126, opened 2026-04-07
feat: Clerk CLI Skill

wyattjoh:
## Summary

Bundles the `clerk` skill into the CLI binary, pins it to the version of the binary that installed it, and teaches the skill how to invoke `clerk` using whatever runner the project prefers.

### How it's bundled

- Markdown files at `<repo-root>/skills/clerk/` are pulled into `skills.ts` as [text imports](https://bun.com/docs/bundler/loaders#text) (`import md from "./SKILL.md" with { type: "text" }`). These resolve live during `bun run dev` and get embedded by `bun build --compile`, so the skill content always matches the binary running it.
- At install time, the staged bundle is handed to `<runner> skills add <tmpdir> --copy`. `--copy` is required: the default symlink mode would point each agent's skill dir at the temp dir we delete right after the install.
- `skills-lock.json` records the install with `sourceType: "local"`, correctly excluding it from `skills update`. The skill can only change when the CLI is upgraded.

### How it pins

- Every asset is piped through a new `renderSkillVersionPlaceholder(content, version)` helper that substitutes `{{CLI_VERSION}}` at staging time. Release builds pin to the shipped version; dev builds (`0.0.0-dev`) resolve to `latest`.
- `DEV_CLI_VERSION` and `resolveCliVersion()` live in a new `lib/version.ts` so the three sites that care about the dev sentinel (the `--version` fallback, the skill templater, and `scripts/build.ts`) share one source of truth and can't drift.

### How the skill invokes `clerk`

A new "Invoking the CLI" section in `SKILL.md` teaches Claude to prefer a globally installed `clerk` binary first, and fall back to a pinned `bunx` / `npx -y` / `pnpm dlx` / `yarn dlx` in lockfile-preferred order. Mirrors the CLI's own `preferredRunner` logic.

### Skill correctness

- Removes a wrong claim that `clerk init --prompt` prints a framework-specific integration guide. It prints a short agent handoff telling the agent to run `clerk init -y`.
- Resyncs the skill against the current CLI surface: adds `init`, `apps create`, `open`, `completion`, and `skill install` to the Core commands table; adds `--destructive` to `config patch`/`put`; strengthens the Prerequisites section so `clerk doctor --json` is the mandatory session-start check; documents agent-mode behavior for `apps create` and `clerk open`.
- Adds `--name` (with `--starter`) to the `init` row and `--secret-key` to the `api` row in the Core commands table, and corrects the `clerk api ls --platform apps` example (dropped an erroneous `--`).
- Repositions `clerk <command> --help` as the source of truth for flags so the skill stays a hint rather than a spec, reducing drift surface.
- Ports useful additions from https://github.com/clerk/skills/pull/29 (verified against CLI source): documents the `clerk auth login`/`logout` aliases (`signup`/`signin`/`sign-in`, `signout`/`sign-out`) and the top-level `clerk login`/`clerk logout` shortcuts; notes that `config` commands authenticate via the Platform API and ignore `--secret-key`; extends the `--destructive` explanation to `config patch` (same semantics as `config put`); adds a table mapping each failing `clerk doctor` check to the manual remediation command (`auth login` / `link` / `env pull`) so agents can remediate without `--fix` (which is disabled in agent mode).

### Skill rename

The bundled skill is named `clerk` (directory: `skills/clerk/`, frontmatter `name: clerk`). CLI-level identifiers that ship to user state (macOS `KEYCHAIN_SERVICE`, `envPaths` config/cache dir) keep the `clerk-cli` name to avoid orphaning existing installs on upgrade.

### Installer continuity

- Upstream `clerk/skills` continues to install via default symlink mode; the two installer calls share runner detection and fail independently.
- `buildSkillsArgs` gains a `copy` parameter and threads it through `runSkillsAdd` so only the clerk skill install gets `--copy`.

Stacked on #125.

## Test plan

- [ ] `bun run test` passes (unit tests cover `buildSkillsArgs` `--copy`, `renderSkillVersionPlaceholder` edge cases, and `withStagedClerkSkill` staging + cleanup + version rendering)
- [ ] Manual: run `clerk init` in a sandbox; confirm `clerk` and framework-pattern installs both succeed; check `skills-lock.json` records `clerk` with `sourceType: "local"`
- [ ] Manual: confirm `.claude/skills/clerk/` contains real files (not a broken symlink) after init
- [ ] Manual: confirm the installed `SKILL.md` no longer contains the literal `{{CLI_VERSION}}` after install


wyattjoh:
**Stack: init-skills**

- ~~#125~~
- **#126 👈 this PR**
  - #127
  - #156

packages/cli-core/src/commands/init/skills.ts
```diff
@@ -1,34 +1,32 @@
...
-  runnerForPackageManager,
-} from "../../lib/runners.js";
-import { isNonEmpty } from "../../lib/helpers/arrays.js";
+import { confirm } from "../../lib/prompts.js";
 import type { ProjectContext } from "./frameworks/types.js";
+import { installClerkSkillCore, resolveSkillsRunner, runSkillsAdd } from "../skill/install.js";
 
-/** Skills installed regardless of framework. */
+/** Upstream skills installed regardless of framework. */
 const BASE_SKILLS = ["clerk", "clerk-setup"];
```
rafa-thayto: The upstream `clerk/skills` repo still ships a skill named `clerk`, and it's installed *after* the bundled one — so it overwrites `.claude/skills/clerk/` and the `{{CLI_VERSION}}` pinning is silently lost.

```suggestion
const BASE_SKILLS = ["clerk-setup"];
```

If upstream still needs to ship `clerk` for some reason, flipping the install order so the bundled one runs last would work too.

packages/cli-core/src/commands/init/skills.ts
```diff
@@ -42,41 +40,25 @@ const FRAMEWORK_SKILL_MAP: Record<string, string> = {
...
 export async function installSkills(
   cwd: string,
   frameworkDep: string | undefined,
   packageManager: ProjectContext["packageManager"] | undefined,
   skipPrompt: boolean,
 ): Promise<void> {
-  const skills = resolveSkills(frameworkDep);
-  const skillList = skills.join(", ");
+  const upstreamSkills = resolveUpstreamSkills(frameworkDep);
+  const skillList = ["clerk", ...upstreamSkills].join(", ");
```
rafa-thayto, line 61: `upstreamSkills` already starts with `"clerk"` (from `BASE_SKILLS`), so the prompt renders as `Install agent skills? (clerk, clerk, clerk-setup, …)`.

If `"clerk"` is dropped from `BASE_SKILLS` above, this becomes correct naturally:

```suggestion
  const skillList = ["clerk", ...upstreamSkills].join(", ");
```

(Same line, but the underlying list no longer duplicates.) Otherwise dedupe explicitly:

```ts
const skillList = [...new Set(["clerk", ...upstreamSkills])].join(", ");
```

packages/cli-core/src/lib/version.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * The string printed by `clerk --version` when the compile-time `CLI_VERSION`
+ * global is undefined (dev builds via `bun run dev`). Used in two places that
+ * must stay in lockstep: the CLI's own version flag fallback, and the skill
+ * installer's detection of "this binary isn't really versioned" so it can tell
+ * the installed skill to pin against `latest` instead of a fake number.
+ */
+export const DEV_CLI_VERSION = "0.0.0-dev";
```
rafa-thayto, line 8: Nice to have one source of truth — but `lib/update-check.ts` still hardcodes the literal in `getCurrentVersion()` and `isDevVersion()`. Worth migrating those too so the sentinel really can't drift:

```ts
// in update-check.ts
import { DEV_CLI_VERSION } from "./version.ts";

export function getCurrentVersion(): string {
  return typeof CLI_VERSION !== "undefined" ? CLI_VERSION : DEV_CLI_VERSION;
}

export function isDevVersion(version: string): boolean {
  return version === DEV_CLI_VERSION;
}
```

packages/cli-core/src/commands/skill/install.ts
```diff
@@ -0,0 +1,254 @@
...
+import {
+  type Runner,
+  detectAvailableRunners,
+  preferredRunner,
+  runnerCommand,
+  runnerForPackageManager,
+} from "../../lib/runners.js";
+import { isNonEmpty } from "../../lib/helpers/arrays.js";
+import type { PackageManager } from "../../lib/package-manager.js";
+import { detectPackageManager } from "../init/context.js";
```
rafa-thayto: Reaching across into another command's module (`commands/init/context.ts`) for a helper that's pure lockfile detection feels a bit tangled — `lib/package-manager.ts` is right there and is the natural home.

```suggestion
import type { PackageManager } from "../../lib/package-manager.js";
import { detectPackageManager } from "../../lib/package-manager.js";
```

…and move `detectPackageManager` into `lib/package-manager.ts`, re-exporting from `context.ts` if anything still imports it from there.

packages/cli-core/src/commands/skill/install.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      expect(skill).not.toContain("{{CLI_VERSION}}");
+    });
+  });
+
+  test("passes undefined through as `latest`", async () => {
+    await withStagedClerkSkill(undefined, async (stageDir) => {
+      const skill = await readFile(join(stageDir, "clerk/SKILL.md"), "utf8");
+      expect(skill).not.toContain("{{CLI_VERSION}}");
+    });
+  });
```
rafa-thayto, line 136: If `{{CLI_VERSION}}` gets added to one of the reference files and rendering regresses, this test passes silently because it only checks `SKILL.md`. Worth asserting across every staged file:

```ts
test("substitutes CLI_VERSION into every staged file", async () => {
  await withStagedClerkSkill("4.5.6", async (stageDir) => {
    for (const rel of ["clerk/SKILL.md", "clerk/references/auth.md", "clerk/references/recipes.md", "clerk/references/agent-mode.md"]) {
      const content = await readFile(join(stageDir, rel), "utf8");
      expect(content, rel).not.toContain("{{CLI_VERSION}}");
    }
  });
});
```

rafa-thayto (approved):

PR #127, opened 2026-04-07
feat(init): support CLERK_SKILL_SOURCE env override

wyattjoh:
## Summary

Adds a `CLERK_SKILL_SOURCE` env var as an escape hatch for skill authors iterating on the `clerk` skill without rebuilding the CLI. When set, `clerk init` passes the value straight to `<runner> skills add <value>` in place of the bundled source (#126).

Any value the `skills` CLI accepts works:

```sh
# Absolute path to a working-tree skill dir (default symlink install,
# edits to the source are reflected in the installed skill immediately).
CLERK_SKILL_SOURCE="$PWD/skills/clerk" clerk init

# A fork or PR branch on GitHub.
CLERK_SKILL_SOURCE="https://github.com/me/cli/tree/wip/skills/clerk" clerk init

# Shorthand for the default repo (installs from main branch).
CLERK_SKILL_SOURCE="clerk/cli" clerk init
```

The override path skips the staged temp dir and the `--copy` flag introduced in #126, so a local working-tree path installs via the default symlink mode and edits to the source are picked up by already-installed projects without re-running `clerk init`. `init` logs the value being used so there's no surprise about what got installed.

The override has no effect on the upstream skills.

Stacked on #126.

## Test plan

- [ ] `bun run test` passes (unit tests cover env-var precedence: unset, empty/whitespace, trimmed / remote / local variants)
- [ ] Manual: `CLERK_SKILL_SOURCE="$PWD/skills/clerk" bun run dev -- init` in a sandbox, confirm the installer logs the override and `.claude/skills/clerk/` is a symlink into the working tree
- [ ] Manual: edit `skills/clerk/SKILL.md` in the working tree, re-open the installed skill, confirm the edit is visible without re-running init
- [ ] Manual: `CLERK_SKILL_SOURCE="clerk/cli" bun run dev -- init` installs from the remote shorthand and the upstream framework skills still install normally

### DIFF
--- a/packages/cli-core/src/commands/skill/install.test.ts
+++ b/packages/cli-core/src/commands/skill/install.test.ts
@@ -3,7 +3,12 @@ import { existsSync } from "node:fs";
 import { readFile, stat } from "node:fs/promises";
 import { join } from "node:path";
 import YAML from "yaml";
-import { buildSkillsArgs, renderSkillVersionPlaceholder, withStagedClerkSkill } from "./install.ts";
+import {
+  buildSkillsArgs,
+  renderSkillVersionPlaceholder,
+  resolveClerkSkillOverride,
+  withStagedClerkSkill,
+} from "./install.ts";
 
 describe("buildSkillsArgs", () => {
   const skills = ["clerk", "clerk-setup", "clerk-nextjs-patterns"];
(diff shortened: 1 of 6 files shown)


wyattjoh:
**Stack: feat/clerk-cli-skill-local-debug**

- ~~#125~~
- ~~#126~~
  - **#127 👈 this PR**

packages/cli-core/src/commands/skill/install.test.ts
```diff
@@ -168,3 +173,26 @@ describe("renderSkillVersionPlaceholder", () => {
...
+    expect(resolveClerkSkillOverride({ CLERK_SKILL_SOURCE: "clerk/cli" })).toBe("clerk/cli");
+    expect(resolveClerkSkillOverride({ CLERK_SKILL_SOURCE: "  /tmp/my-skill  " })).toBe(
+      "/tmp/my-skill",
+    );
+    expect(
+      resolveClerkSkillOverride({
+        CLERK_SKILL_SOURCE: "https://github.com/me/fork/tree/wip/skills/clerk",
+      }),
+    ).toBe("https://github.com/me/fork/tree/wip/skills/clerk");
+  });
```
rafa-thayto, line 284: The resolver tests are solid! Would be nice to also have a test that verifies `installClerkSkillCore` actually routes through the override when the env var is set (i.e., calls `runSkillsAdd` with the override value and `copy: false`) vs. falling through to `withStagedClerkSkill` when unset. That way the wiring is covered, not just the env parsing.

PR #131, opened 2026-04-09
feat(init): non-interactive bootstrap, auth headers, and keyless support

rafa-thayto:
## Summary

- **Non-interactive bootstrap**: `clerk init` now runs the full flow in agent mode (implies `--yes`). `--prompt` is separated to just print guidance.
- **New CLI options**: `--pm <manager>` overrides package manager detection, `--name <project-name>` sets the project name for `--starter` (both skip prompts).
- **Auth header injection**: During `--starter` bootstrap, all framework scaffolders inject a ready-to-use auth header with `<Show>`, `<SignInButton>`, `<SignUpButton>`, and `<UserButton>` components.
- **Keyless support flag**: `FrameworkInfo.supportsKeyless` marks which SDKs support keyless mode. Frameworks without it force authentication during init.
- **Nuxt keyless mode**: Enabled keyless for Nuxt via a snapshot SDK (`@clerk/nuxt@2.2.0-snapshot.v20260413174426`). Added `sdkInstall` field to `FrameworkInfo` for pinning install specifiers that differ from the package name.
- **Bootstrap registry split**: `BOOTSTRAP_REGISTRY` is now split into `BOOTSTRAP_KEYLESS_REGISTRY` and `BOOTSTRAP_AUTHENTICATED_REGISTRY` so the init flow can distinguish frameworks that support keyless from those requiring API keys.
- **Vue router bootstrap**: New Vue projects without `vue-router` get it installed automatically with a router config and `App.vue` updated to use `<RouterView />`.
- **Nuxt index page**: Bootstrap creates an `app/pages/index.vue` so the root route renders content instead of a blank page.
- **Astro SSR output**: Bootstrap enables `output: "server"` in the Astro config so auth middleware works out of the box.
- **JavaScript publishable key validation**: The vanilla JS scaffold now throws a clear error with actionable instructions when `VITE_CLERK_PUBLISHABLE_KEY` is missing.
- **DRY refactoring**: Unified header block generation (HTML vs JSX) via shared `buildHeaderBlock` helper. Extracted `runPmInstall`, `addBootstrapHeader`, and `resolveKeylessMode` helpers. Simplified `bootstrap.ts` by inlining `resolvePm`/`resolveName`.
- **TypeScript fixes**: Added missing `envFile` to test stubs, fixed strict null errors in regex captures, removed unused imports.

## Keyless support matrix

| Framework      | SDK                           | Keyless? |
|----------------|-------------------------------|----------|
| Next.js        | `@clerk/nextjs`               | Yes      |
| Astro          | `@clerk/astro`                | Yes      |
| Nuxt           | `@clerk/nuxt` (snapshot)      | Yes      |
| TanStack Start | `@clerk/tanstack-react-start` | Yes      |
| React Router   | `@clerk/react-router`         | Yes      |
| React          | `@clerk/react`                | No       |
| Vue            | `@clerk/vue`                  | No       |
| JavaScript     | `@clerk/clerk-js`             | No       |

## Test plan

- [x] All 65 unit/integration tests pass (`bun run test`)
- [x] Zero lint errors (`bun run lint`)
- [x] Format check passes (`bun run format`)
- [x] Manual: scaffolded all 32 framework × PM combinations (`--starter` with bun/pnpm/npm/yarn for all 8 frameworks) — all scaffold successfully
- [x] Manual: all 32 dev servers start and return HTTP 200
- [x] Manual: keyless frameworks (next, react-router, astro, nuxt, tanstack-start) auto-generate temporary dev keys
- [x] Manual: non-keyless frameworks (react, vue, javascript) show clear "requires API keys" message with instructions
- [x] Manual: sign-in/sign-up pages exist and render Clerk components for all frameworks with routing
- [x] Manual: `clerk init --starter --framework next --pm bun --name my-app` creates project non-interactively
- [x] Manual: `clerk init --prompt` prints guidance without running init

### DIFF
--- a/packages/cli-core/src/commands/init/bootstrap.test.ts
+++ b/packages/cli-core/src/commands/init/bootstrap.test.ts
@@ -1,5 +1,6 @@
-import { test, expect, describe } from "bun:test";
+import { test, expect, describe, spyOn, afterEach } from "bun:test";
 import { BOOTSTRAP_REGISTRY } from "./bootstrap-registry.ts";
+import { resolvePackageManager } from "./bootstrap.ts";
 
 function entryFor(dep: string) {
   const entry = BOOTSTRAP_REGISTRY.find((e) => e.dep === dep);
@@ -122,5 +123,46 @@ describe("BOOTSTRAP_REGISTRY", () => {
     const cmd = entryFor("react").buildCommand("pnpm", "app");
     expect(cmd[0]).toBe("pnpm");
     expect(cmd[1]).toBe("dlx");
+  });
+});
+
+describe("resolvePackageManager", () => {
+  let whichSpy: ReturnType<typeof spyOn>;
+
+  afterEach(() => {
+    whichSpy?.mockRestore();
+  });
+
+  test("returns first available PM in priority order", () => {
+    whichSpy = spyOn(Bun, "which").mockImplementation((bin) => {
+      if (bin === "bun") return "/usr/local/bin/bun";
+      return null;
+    });
+
+    expect(resolvePackageManager()).toBe("bun");
+  });
+
+  test("falls back to pnpm when bun is unavailable", () => {
+    whichSpy = spyOn(Bun, "which").mockImplementation((bin) => {
+      if (bin === "pnpm") return "/usr/local/bin/pnpm";
+      return null;
+    });
+
+    expect(resolvePackageManager()).toBe("pnpm");
+  });
+
+  test("falls back to yarn when bun and pnpm are unavailable", () => {
+    whichSpy = spyOn(Bun, "which").mockImplementation((bin) => {
+      if (bin === "yarn") return "/usr/local/bin/yarn";
+      return null;
+    });
+
+    expect(resolvePackageManager()).toBe("yarn");
+  });
+
+  test("falls back to npm as last resort", () => {
+    whichSpy = spyOn(Bun, "which").mockReturnValue(null);
+
+    expect(resolvePackageManager()).toBe("npm");
   });
 });
--- a/packages/cli-core/src/commands/init/bootstrap.test.ts
+++ b/packages/cli-core/src/commands/init/bootstrap.test.ts
@@ -1,4 +1,4 @@
-import { test, expect, describe, spyOn, afterEach } from "bun:test";
+import { test, expect, describe, spyOn, afterAll } from "bun:test";
 import { BOOTSTRAP_REGISTRY } from "./bootstrap-registry.ts";
 import { resolvePackageManager } from "./bootstrap.ts";
 
@@ -127,14 +127,14 @@ describe("BOOTSTRAP_REGISTRY", () => {
 });
 
 describe("resolvePackageManager", () => {
-  let whichSpy: ReturnType<typeof spyOn>;
+  const whichSpy = spyOn(Bun, "which");
 
-  afterEach(() => {
-    whichSpy?.mockRestore();
+  afterAll(() => {
+    whichSpy.mockRestore();
   });
 
   test("returns first available PM in priority order", () => {
-    whichSpy = spyOn(Bun, "which").mockImplementation((bin) => {
+    whichSpy.mockImplementation((bin) => {
       if (bin === "bun") return "/usr/local/bin/bun";
       return null;
     });
@@ -143,7 +143,7 @@ describe("resolvePackageManager", () => {
   });
 
   test("falls back to pnpm when bun is unavailable", () => {
-    whichSpy = spyOn(Bun, "which").mockImplementation((bin) => {
+    whichSpy.mockImplementation((bin) => {
       if (bin === "pnpm") return "/usr/local/bin/pnpm";
       return null;
     });
@@ -152,7 +152,7 @@ describe("resolvePackageManager", () => {
   });
 
   test("falls back to yarn when bun and pnpm are unavailable", () => {
-    whichSpy = spyOn(Bun, "which").mockImplementation((bin) => {
+    whichSpy.mockImplementation((bin) => {
       if (bin === "yarn") return "/usr/local/bin/yarn";
       return null;
     });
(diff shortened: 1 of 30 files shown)


packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -153,9 +170,21 @@ export async function promptAndBootstrap(
...
+    if (valid !== true) throwUsageError(valid);
+  }
+
   const entry = await pickFramework(frameworkOverride);
-  const pm = await pickPackageManager();
-  const projectName = await askProjectName(entry);
+  const pm = pmOverride ?? (skipConfirm ? resolvePackageManager() : await pickPackageManager());
+  const projectName =
+    nameOverride ?? (skipConfirm ? entry.defaultProjectName : await askProjectName(entry));
   const projectDir = join(cwd, projectName);
```
wyattjoh, line 188: In `--yes`/agent mode, if `projectDir` already exists and is non-empty, the downstream generators diverge: `create-next-app` errors out, but `create-vite` and `create-nuxt` drop into interactive prompts, which hang forever without a TTY. Worth a quick stat + `CliError` here before `generateProject` runs:

```ts
if (await exists(projectDir)) {
  throw new CliError(`Directory '${projectName}' already exists. Pick a different name or remove it first.`);
}
```

packages/cli-core/src/commands/init/bootstrap.ts
```diff
@@ -153,9 +170,21 @@ export async function promptAndBootstrap(
...
+  if (skipConfirm && !frameworkOverride) {
+    throwUsageError(
+      "Non-interactive mode requires --framework for new projects. Example: clerk init --starter --framework next",
+    );
+  }
+
+  if (nameOverride) {
+    const valid = validateProjectName(nameOverride);
+    if (valid !== true) throwUsageError(valid);
+  }
```
wyattjoh, line 182: The error surfaces as bare `Project name must be lowercase`, with no indication of which flag value was rejected. Echoing the input makes the usage error self-describing, especially helpful when multiple flags are being passed:

```ts
if (valid !== true) throwUsageError(`Invalid --name "${nameOverride}": ${valid}`);
```

packages/cli-core/src/commands/init/index.ts
```diff
@@ -177,6 +193,37 @@ function printBootstrapNextSteps(
...
 
+function printManualSetupInfo(frameworkName: string): void {
+  const lines = [
+    `\n  ${frameworkName} requires API keys — set them up manually:`,
+    "    clerk auth login",
+    "    clerk link",
+    "    clerk env pull",
+  ];
+  log.info(lines.map(dim).join("\n"));
+}
```
wyattjoh: This helper only runs on the bootstrap + `skipConfirm` path, but the name reads like a generic "manual setup" printer. A future caller is likely to reach for it expecting it to fire anywhere manual setup is needed. Consider `printBootstrapManualSetupInfo` so the call-site invariant is visible at the callsite.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -100,7 +116,7 @@ export async function init(options: InitOptions = {}) {
 
   if (options.skills !== false) {
     bar();
-    await installSkills(ctx.cwd, ctx?.framework.dep, ctx?.packageManager, options.yes ?? false);
+    await installSkills(ctx.cwd, ctx?.framework.dep, ctx?.packageManager, overrides.skipConfirm);
```
wyattjoh: `ctx.cwd` on the same line would have already thrown if `ctx` were null, so the `?.` on the next two arguments is dead code and slightly misleading (a reader might wonder in which path `ctx` could be undefined here). Drop to `ctx.framework.dep` and `ctx.packageManager`.

packages/cli-core/src/commands/init/heuristics.ts
```diff
@@ -11,47 +11,57 @@ import { withSpinner } from "../../lib/spinner.js";
...
-  // only have npm). Fail fast with a useful message rather than a raw ENOENT.
-  const pmBinary = addCmd.split(" ")[0] ?? addCmd;
   if (Bun.which(pmBinary) === null) {
-    log.warn(
-      `${pmBinary} is not installed but the project's lockfile suggests it. ` +
-        `Install ${pmBinary} or run \`${addCmd} ${ctx.framework.sdk}\` manually with another package manager.`,
-    );
+    log.warn(`${pmBinary} is not installed. Install manually: ${manualCmd}`);
     return;
   }
```
wyattjoh, line 30: The old `installSdk` message told users that the lockfile suggested a PM that isn't installed (e.g. `bun.lock` exists because a teammate committed it, but bun isn't on this machine). That nuance is useful precisely in the `installSdk` path where the PM comes from lockfile detection, not user choice. Consider letting `runPmInstall` take an optional context string (or a boolean like `fromLockfile`) so `installSdk` can surface the richer hint while `installDeps` keeps the generic one.

packages/cli-core/src/commands/init/bootstrap.test.ts
```diff
@@ -124,3 +125,44 @@ describe("BOOTSTRAP_REGISTRY", () => {
     expect(cmd[1]).toBe("dlx");
   });
 });
+
+describe("resolvePackageManager", () => {
+  let whichSpy: ReturnType<typeof spyOn>;
+
+  afterEach(() => {
+    whichSpy?.mockRestore();
+  });
```
wyattjoh: `rules/testing.md` prescribes `afterAll` for `mockRestore()`. Functionally this works, but it is inconsistent with the codified convention, and the `whichSpy?.` guard makes the intent ambiguous (the spy is created fresh in each test, so restoring per-test is a no-op for all but the last). Either hoist `whichSpy` and restore once in `afterAll`, or leave restore inside each test.

wyattjoh (commented):
Branch looks great 😃! Clean split of `bootstrap-registry.ts`, good test coverage on the new flags, and `supportsKeyless` is a tidy gate. A handful of small polish comments inline, all non-blocking.

wyattjoh (approved):

PR #135, opened 2026-04-09
chore(ci): remove NODE_AUTH_TOKEN once OIDC trusted publishing is enabled

wyattjoh:
Removes `NODE_AUTH_TOKEN` from all three npm publish jobs now that OIDC trusted publishing is configured for the package on npmjs.com.

> **Do not merge until OIDC trusted publishing is enabled on npmjs.com for this package.**

Affects:
- `publish-npm` (stable release)
- `canary-publish-npm` (canary release)
- `publish` (snapshot release)

PR #134, opened 2026-04-09
refactor(scripts): consolidate releaser into scripts/lib

wyattjoh:
Reorganizes the scripts directory to reduce nesting and improve cohesion.

- Move `scripts/releaser/targets.ts` to `scripts/lib/targets.ts` — already shared by `build.ts` and `sign-macos.ts`, now lives with other shared lib utilities
- Add `run()` and `publish()` to `scripts/lib/npm.ts` alongside the existing `isPublished()` — all npm subprocess logic in one place
- Flatten `scripts/releaser/index.ts` to `scripts/releaser.ts` — the subdirectory only held one file
- Update import paths in `build.ts`, `sign-macos.ts`, and `package.json`

PR #139, opened 2026-04-09
fix(ci): upgrade npm publish jobs to Node.js LTS

wyattjoh:
## Summary

- Switch `node-version` from `"22"` to `"lts/*"` in both `publish-npm` and `canary-publish-npm` jobs
- Remove the `npm install -g npm@11` upgrade step (Node 24 LTS ships npm 11 already)

## Why

`node:22` resolved to Node 22.22.2, which ships with a broken bundled npm (missing `promise-retry` in its internal dependency tree). This caused the upgrade step to fail before any publishing could happen. Node 24 LTS (`lts/*`) does not have this issue and already includes npm 11.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -85,10 +85,8 @@ jobs:
       - uses: oven-sh/setup-bun@v2
       - uses: actions/setup-node@v6
         with:
-          node-version: "22"
+          node-version: "lts/*"
           registry-url: "https://registry.npmjs.org"
-      - name: Upgrade npm for OIDC trusted publishing
-        run: npm install -g npm@11
       - run: bun install --frozen-lockfile
 
       - uses: actions/download-artifact@v8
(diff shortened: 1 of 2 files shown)


wyattjoh:
**Stack: npm-publishing-fixes**

- **#139 👈 this PR**
  - #134
    - #135

alexcarpenter (approved):

.github/workflows/release.yml
```diff
@@ -85,10 +85,8 @@ jobs:
       - uses: oven-sh/setup-bun@v2
       - uses: actions/setup-node@v6
         with:
-          node-version: "22"
+          node-version: "lts/*"
```
dmoerner, line 88: Isn't "lts/krypton" safer, since this will just resolve to whatever is the newest LTS, but eventually that will be Node 26 or whatever.

wyattjoh: This is just to enable the `npm publish` flow, so getting the latest LTS version of that should be ok I'd think!

dmoerner (approved):

PR #149, opened 2026-04-12
feat(scripts): add Homebrew distribution

wyattjoh:
## Summary

- Adds a Homebrew distribution library (`scripts/lib/homebrew.ts`) that renders Ruby formula files from platform binary metadata
- Supports versioned formulas (e.g., `clerk@2.rb`) alongside the latest (`clerk.rb`) for major-version pinning
- Adds the `scripts/homebrew.ts` entry point and a CI job in the release workflow to generate and push formulas to a tap repo
- Documents the Homebrew distribution flow in `README.md` and `docs/releasing.md`

## Test plan

- [ ] `bun test scripts/lib/homebrew.test.ts` passes
- [ ] Formula rendering produces valid Ruby syntax for all platform targets
- [ ] Versioned formula (`clerk@N.rb`) is generated alongside `clerk.rb`

### DIFF
--- a/scripts/homebrew.ts
+++ b/scripts/homebrew.ts
@@ -12,37 +12,28 @@ import {
 } from "./lib/homebrew.ts";
 import { run } from "./lib/npm.ts";
 
-const ARTIFACTS_DIR = process.env.ARTIFACTS_DIR ?? join(import.meta.dir, "../dist/artifacts");
+const DEFAULT_ARTIFACTS_DIR =
+  process.env.ARTIFACTS_DIR ?? join(import.meta.dir, "../dist/artifacts");
 
-function parseCliArgs(): {
-  version: string;
-  artifactsDir: string;
-  tapRepo: string;
-  dryRun: boolean;
-} {
-  const { values } = parseArgs({
-    args: Bun.argv.slice(2),
-    options: {
-      version: { type: "string" },
-      "artifacts-dir": { type: "string" },
-      "tap-repo": { type: "string" },
-      "dry-run": { type: "boolean", default: false },
-    },
-  });
+const { values } = parseArgs({
+  options: {
+    version: { type: "string" },
+    "artifacts-dir": { type: "string" },
+    "tap-repo": { type: "string" },
+    "dry-run": { type: "boolean", default: false },
+  },
+  strict: true,
+});
 
-  if (!values.version) {
-    throw new Error("--version is required");
-  }
-
-  return {
-    version: values.version,
-    artifactsDir: values["artifacts-dir"] ?? ARTIFACTS_DIR,
-    tapRepo: values["tap-repo"] ?? "clerk/homebrew-stable",
-    dryRun: values["dry-run"]!,
-  };
+if (!values.version) {
+  console.error("--version is required.");
+  process.exit(1);
 }
 
-const { version, artifactsDir, tapRepo, dryRun } = parseCliArgs();
+const version = values.version;
+const artifactsDir = values["artifacts-dir"] ?? DEFAULT_ARTIFACTS_DIR;
+const tapRepo = values["tap-repo"] ?? "clerk/homebrew-stable";
+const dryRun = values["dry-run"]!;
 
 console.log(`Homebrew distribution: v${version}${dryRun ? " (dry run)" : ""}`);
 console.log(`Artifacts dir: ${artifactsDir}`);
--- a/scripts/homebrew.ts
+++ b/scripts/homebrew.ts
@@ -6,6 +6,7 @@ import {
   renderFormula,
   createArchive,
   computeChecksum,
+  parseMajorVersion,
   HOMEBREW_TARGETS,
   type FormulaInput,
 } from "./lib/homebrew.ts";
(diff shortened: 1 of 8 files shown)


wyattjoh:
**Stack: homebrew**

- **#149 👈 this PR**

scripts/homebrew.ts
```diff
@@ -0,0 +1,138 @@
...
+console.log(`Work directory: ${workDir}`);
+
+const archivePaths = new Map<string, string>();
+
+for (const target of HOMEBREW_TARGETS) {
+  const binaryPath = join(artifactsDir, `clerk-${target.name}`, "clerk");
+  const archivePath = join(workDir, `homebrew-clerk-${target.name}.tar.gz`);
+  console.log(`Creating archive for ${target.name}...`);
+  createArchive(binaryPath, archivePath);
+  archivePaths.set(target.name, archivePath);
```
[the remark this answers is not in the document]
wyattjoh: This is a false positive. The workflow uses `actions/download-artifact@v8` with `pattern: clerk-{darwin-arm64,darwin-x64,linux-arm64,linux-x64}` and `path: dist/artifacts` (`.github/workflows/release.yml:146-149`), which creates one subdirectory per artifact named after the artifact, yielding `dist/artifacts/clerk-<target>/clerk`. That matches the current code at `scripts/homebrew.ts:48`. The canary and snapshot jobs (lines 271-274, 410-413) also explicitly rename into `clerk-<target>/` before invoking the script, so the shape is consistent across all three channels. No change needed.

scripts/homebrew.ts
```diff
@@ -0,0 +1,138 @@
...
+console.log(versionedFormula);
+
+if (dryRun) {
+  console.log("[dry-run] Skipping tap clone and push.");
+  console.log(`[dry-run] Would write Formula/clerk.rb and Formula/clerk@${major}.rb`);
+} else {
+  const token = process.env.HOMEBREW_TAP_TOKEN;
+  if (!token) {
+    throw new Error("HOMEBREW_TAP_TOKEN env var is required for pushing to tap");
+  }
```
rafa-thayto: This runs after all archives are created and uploaded to the release. If the secret is missing, the job burns CI minutes before failing here.

Moving the check right after arg parsing (after line 36) would fail fast:

```ts
if (!dryRun && !process.env.HOMEBREW_TAP_TOKEN) {
  console.error("HOMEBREW_TAP_TOKEN is required (use --dry-run to skip).");
  process.exit(1);
}
```

Then this block simplifies to:

```suggestion
  const token = process.env.HOMEBREW_TAP_TOKEN!;
```

scripts/homebrew.ts
```diff
@@ -0,0 +1,138 @@
...
+  run(
+    [
+      "git",
+      "remote",
+      "set-url",
+      "origin",
+      `https://x-access-token:${token}@github.com/${tapRepo}.git`,
+    ],
+    { cwd: tapWorkDir },
+  );
```
rafa-thayto: `run()` includes the full command in its error message (`cmd.join(" ")` at `npm.ts:12`). If `set-url` fails, the token ends up in CI logs.

Using `Bun.spawnSync` directly with a sanitized error avoids that:

```suggestion
  const setUrlResult = Bun.spawnSync(
    ["git", "remote", "set-url", "origin", `https://x-access-token:${token}@github.com/${tapRepo}.git`],
    { cwd: tapWorkDir, stdio: ["ignore", "pipe", "pipe"] },
  );
  if (setUrlResult.exitCode !== 0) {
    throw new Error("Failed to configure authenticated remote for tap repo");
  }
```

scripts/lib/homebrew.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+export function renderFormula(input: FormulaInput): string {
+  const { version, checksums, major } = input;
+
+  const className = major != null ? `ClerkAT${major}` : "Clerk";
+  const kegOnly = major != null ? "\n  keg_only :versioned_formula" : "";
+
+  return `class ${className} < Formula
+  desc "The Clerk CLI"
```
rafa-thayto: Homebrew style says no leading article in `desc`. Not enforced for taps, but saves a review round when migrating to core.

```suggestion
  desc "Clerk command-line interface"
```

scripts/lib/homebrew.ts
```diff
@@ -0,0 +1,115 @@
...
+  on_linux do
+    if Hardware::CPU.arm? && Hardware::CPU.is_64_bit?
+      url "${assetUrl(version, "linux-arm64")}"
+      sha256 "${checksums["linux-arm64"]}"
+    end
+    if Hardware::CPU.intel?
+      url "${assetUrl(version, "linux-x64")}"
+      sha256 "${checksums["linux-x64"]}"
+    end
+  end
```
rafa-thayto, line 59: The modern Homebrew DSL uses `on_arm do` / `on_intel do` blocks instead of `if Hardware::CPU.arm?`. Both work, but the block form is what `brew audit --new` prefers. The `is_64_bit?` guard on Linux ARM is also redundant — Homebrew only supports aarch64 on Linux.

```suggestion
  on_macos do
    on_arm do
      url "${assetUrl(version, "darwin-arm64")}"
      sha256 "${checksums["darwin-arm64"]}"
    end
    on_intel do
      url "${assetUrl(version, "darwin-x64")}"
      sha256 "${checksums["darwin-x64"]}"
    end
  end

  on_linux do
    on_arm do
      url "${assetUrl(version, "linux-arm64")}"
      sha256 "${checksums["linux-arm64"]}"
    end
    on_intel do
      url "${assetUrl(version, "linux-x64")}"
      sha256 "${checksums["linux-x64"]}"
    end
  end
```

rafa-thayto (approved):

PR #150, opened 2026-04-13
feat(init): add --app flag to link to a specific application

djgould:
## Summary

- Adds `--app <id>` to `clerk init` so the Clerk dashboard can render a copy-paste command that links a repo to a specific application.
- Tightens the `skipIfLinked` short-circuit in `clerk link` so an explicit `--app` that differs from an existing link falls through to the re-link prompt instead of being silently ignored.

## Behavior

| Repo state | `clerk init --app X` |
|---|---|
| Not linked | Links to `X` silently |
| Linked to `X` already | No-op: prints "Already linked to X" and proceeds |
| Linked to a different app | Prompts "Re-link to X?" (default No) |
| `X` doesn't exist / no access | Errors via `Failed to fetch application` before scaffolding |

Note: `-y` does not currently auto-confirm the re-link prompt — `link()` has no `yes` plumbing. The dashboard copy-paste flow does not use `-y`, so the safety prompt is desired. Threading `yes` through `link()` is a separate follow-up if a future need arises.

## Test plan

- [x] Unit tests in `packages/cli-core/src/commands/link/index.test.ts` cover the three `skipIfLinked` branches (no `--app`, matching `--app`, differing `--app`)
- [x] Unit tests in `packages/cli-core/src/commands/init/index.test.ts` cover `--app` forwarding with and without an existing profile
- [x] Manual: fresh repo with `--app <id>` links + scaffolds (no picker)
- [x] Manual: same app re-run is a silent no-op
- [x] Manual: different app prompts `Re-link to X?` defaulting to No
- [x] Manual: invalid `--app` errors out before scaffolding

### DIFF
--- a/packages/cli-core/src/commands/init/index.ts
+++ b/packages/cli-core/src/commands/init/index.ts
@@ -49,7 +49,7 @@ export async function init(options: InitOptions = {}) {
 
   if (options.prompt || isAgent()) {
     log.data(
-      "Run `clerk init -y` to automatically detect the framework, install the Clerk SDK, and scaffold authentication files without interactive prompts. Pass `--app <id>` to link to a specific Clerk application.",
+      "Run `clerk init -y` to automatically detect the framework, install the Clerk SDK, and scaffold authentication files without interactive prompts.",
     );
     return;
   }
--- a/packages/cli-core/src/commands/init/index.ts
+++ b/packages/cli-core/src/commands/init/index.ts
@@ -49,7 +49,7 @@ export async function init(options: InitOptions = {}) {
 
   if (options.prompt || isAgent()) {
     log.data(
-      "Run `clerk init -y` to automatically detect the framework, install the Clerk SDK, and scaffold authentication files without interactive prompts.",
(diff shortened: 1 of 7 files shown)


djgould:
https://github.com/user-attachments/assets/41640c3b-bc7c-408f-b2a3-1453454a7ec2

packages/cli-core/src/commands/init/index.ts
```diff
@@ -37,6 +37,7 @@ type InitOptions = {
   prompt?: boolean;
   skills?: boolean;
   starter?: boolean;
+  app?: string;
```
wyattjoh, line 46: nit: would be nice for us to get into the habit of adding JSDoc comments to these types, especially as we begin to add a lot of keys it'll get hard to understand what they're used for.

wyattjoh (approved):

packages/cli-core/src/commands/init/index.ts
djgould: good call! added jsdocs for this type

PR #151, opened 2026-04-13
feat(ci): add Slack failure notifications

wyattjoh:
## Summary

- Add `scripts/slack.ts` that posts block-kit formatted failure messages to a Slack incoming webhook
- Extract a reusable `notify-failure.yml` workflow to keep notification steps DRY across CI, stable release, and canary release pipelines

## Requirements

Requires a `SLACK_WEBHOOK_URL` repository secret.

## Test plan

- [ ] Verify `notify-failure.yml` is triggered when a CI job fails
- [ ] Verify Slack message format renders correctly
- [ ] Confirm no regressions in existing CI or release workflows

PR #152, opened 2026-04-13
refactor(ci): extract reusable CI checks and setup-bun composite action

wyattjoh:
## Summary

- Add `.github/actions/setup-bun` composite action (Bun install + optional
  dependency caching + `bun install`). Used by `release.yml`'s `versioning`,
  `canary-version`, and `notify-failure` jobs, all of which run in trusted
  contexts (push to main or workflow_call from push to main).
- Add `snapshot-ci` job in `release.yml` that reuses `ci.yml` via
  `workflow_call`, gating snapshot publishing on CI passing first.
- Gate all release paths (stable, canary, snapshot) on CI checks passing
  before versioning or building.

`ci.yml` keeps inline `oven-sh/setup-bun@v2` + `actions/cache` + `bun install`
in each job rather than adopting the composite action. Switching would trip
CodeQL's "checkout of untrusted code in a privileged context" finding via
the `issue_comment` -> `snapshot-ci` -> `ci.yml` path, since `uses: ./...`
would resolve `action.yml` from the untrusted PR ref.

## Test plan

- [ ] CI workflow runs correctly on a PR (build, lint, test, e2e)
- [ ] Release workflow gates on CI passing before versioning
- [ ] Snapshot workflow gates on CI passing before building
- [ ] Fork PRs and Dependabot PRs correctly skip e2e tests
- [ ] Failure notifications still fire when jobs fail


wyattjoh:
**Stack: slack-notifications**

- ~~#151~~
- **#152 👈 this PR**

.github/workflows/release.yml
```diff
@@ -30,8 +36,7 @@ jobs:
       version: ${{ steps.check.outputs.version }}
     steps:
       - uses: actions/checkout@v6
-      - uses: oven-sh/setup-bun@v2
-      - run: bun install --frozen-lockfile
+      - uses: ./.github/actions/setup-bun
```
rafa-thayto, line 39: `cache` defaults to `"none"` when omitted, so both of these quietly install without the dep cache. Since `versioning` runs after `ci` (which populates the cache via the `build` job's `read-write`), passing `restore` here gets you a warm install basically for free. Same applies to `canary-version`.

```suggestion
      - uses: ./.github/actions/setup-bun
        with:
          cache: restore
```

rafa-thayto:
**`.github/workflows/release.yml`**, snapshot job (lines 371-372):

`versioning` and `canary-version` already moved to the composite — lining up `snapshot` keeps all three identical and saves the next person from wondering why this one is the odd one out.

```diff
      - uses: actions/checkout@v6
        with:
          ref: ${{ steps.pr.outputs.sha }}
-      - uses: oven-sh/setup-bun@v2
-      - run: bun install --frozen-lockfile
+      - uses: ./.github/actions/setup-bun
```

rafa-thayto (approved):

PR #157, opened 2026-04-15
feat: auto-claim keyless applications on login

rafa-thayto:
## Summary

- Add keyless application lifecycle (`keyless.ts`): create accountless apps via BAPI, write framework-specific keys to `.env.local`, manage `.clerk/keyless.json` claim token breadcrumb
- Add autoclaim orchestrator (`autoclaim.ts`): after `clerk auth login`, detect keyless breadcrumb and automatically claim the application via PLAPI — never throws, returns discriminated union so login is never interrupted
- Auto-pull environment variables after successful claim — no manual `clerk env pull` step needed
- Extract shared `linkApp()` from `autolink.ts` for reuse by both autolink (key detection) and autoclaim (keyless claim)
- Add `claimApplication()` PLAPI endpoint (`POST /v1/platform/accountless_applications/claim`)
- Integrate keyless setup into `clerk init` bootstrap when user skips auth
- Integrate autoclaim into `clerk auth login` with contextual warnings and next-steps per failure mode
- Truncate app name to 50 chars to match backend validation limit

## How it works

1. **`clerk init` (skip auth)** → creates accountless app, writes keys to `.env.local`, stores claim token in `.clerk/keyless.json`
2. **Developer builds their app** using temporary dev keys
3. **`clerk auth login`** → after OAuth, reads breadcrumb, calls claim endpoint, links app to account, auto-pulls env vars, clears breadcrumb
4. **Failure handling**: transient errors (5xx) preserve breadcrumb for retry on next login; terminal errors (404/403) clear it; env pull failures warn but don't block login

## Cleanup pass (commit 6cda4f5)

Applied KISS/DRY simplification across the new surface and touched neighboring code where it shared the same patterns. **Net: -121 lines** (107 insertions, 228 deletions).

- **`plapi.ts`**: extracted shared `plapiRequest<T>` helper; 6 near-identical fetch-with-auth blocks (20 lines each) collapsed to 2–3 line call sites. Helper normalizes URL construction, query-param serialization, auth headers, `PlapiError` throw, and JSON parsing.
- **`errors.ts`**: promoted `errorMessage(unknown): string` out of `commands/doctor/checks.ts` so `autoclaim.ts` and `init/index.ts` can reuse it instead of re-implementing `error instanceof Error ? error.message : String(error)` inline.
- **`keyless.ts`**: replaced two TOCTOU `Bun.file(...).exists()` + read patterns with atomic `.text().catch(() => "")` / `.json().catch(...)`; parallelized independent `detectPublishableKeyName` + `detectSecretKeyName` with `Promise.all`; factored `.env.local` / `.clerk` / `keyless.json` into named constants.
- **`autoclaim.ts`**: table-driven `classifyClaimError` — replaced back-to-back `error instanceof PlapiError && error.status === N` branches with a `TERMINAL_BY_STATUS: Record<number, Terminal["status"]>` lookup.
- **Comments**: removed file-level docstring blocks, section-header banners, and WHAT/narration comments from new files per house style (kept non-obvious WHY comments).
- **Tests**: `as never` → `as Profile`, dropped the `spies[]` array in favor of explicit per-spy `mockRestore()`, deleted redundant setup commentary.

## Test plan

- [x] 9 unit tests for autoclaim (all status paths, breadcrumb preservation, env pull success/failure)
- [x] 15 unit tests for keyless (BAPI calls, env writing, breadcrumb I/O, token parsing)
- [x] Login tests pass with autoclaim mocked
- [x] Full test suite passes (73 suites)
- [x] Lint, format, and typecheck clean
- [x] CI green after cleanup pass
- [ ] Requires backend PLAPI claim endpoint to be deployed first (CLI gracefully no-ops if endpoint is unavailable)
- [ ] E2E: `clerk init -y` → skip auth → verify `.env.local` and `.clerk/keyless.json` created → `clerk auth login` → verify app claimed, linked, and env vars refreshed

### DIFF
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -95,6 +95,12 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
 
   if (existingSession && !isHuman()) {
     log.success(`Logged in as ${existingSession.email}`);
+    const claimResult = await handleAutoclaim(process.cwd());
+    if (showNextSteps) {
+      outro(await loginNextSteps(claimResult));
+    } else {
+      outro("Done");
+    }
     return existingSession;
   }
 
@@ -127,8 +133,8 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
 
 const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
   not_found:
-    "Claim token is no longer valid — the application may have been claimed from the dashboard.",
-  already_claimed: "Unable to claim — your account does not have an active organization.",
+    "Claim token is no longer valid - the application may have been claimed from the dashboard.",
+  no_organization: "Unable to claim - your account does not have an active organization.",
   failed: "Auto-claim failed due to a temporary error. It will be retried on your next login.",
 };
 
@@ -138,11 +144,6 @@ async function handleAutoclaim(cwd: string): Promise<AutoclaimResult> {
   if (result.status === "claimed") {
     const label = result.app.name || result.app.application_id;
     log.success(`Claimed and linked application: \`${label}\``);
-    if (!result.envPulled) {
-      log.warn(
-        "Could not refresh environment variables automatically. Run `clerk env pull` manually.",
-      );
-    }
   }
 
   const warning = CLAIM_WARNINGS[result.status];
@@ -156,7 +157,7 @@ async function loginNextSteps(result: AutoclaimResult): Promise<readonly string[
     return result.envPulled ? NEXT_STEPS.AUTOCLAIMED : NEXT_STEPS.AUTOCLAIMED_NO_ENV;
   }
   if (result.status === "failed") return NEXT_STEPS.AUTOCLAIM_RETRY;
-  if (result.status === "not_found" || result.status === "already_claimed") {
+  if (result.status === "not_found" || result.status === "no_organization") {
     return NEXT_STEPS.AUTOCLAIM_MANUAL_LINK;
   }
 
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -160,7 +160,6 @@ async function loginNextSteps(result: AutoclaimResult): Promise<readonly string[
     return NEXT_STEPS.AUTOCLAIM_MANUAL_LINK;
   }
 
-  // not_keyless — normal login, check for existing linked profile
   const linked = await resolveProfile(process.cwd());
   if (!linked) return NEXT_STEPS.LOGIN;
 
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -114,10 +114,10 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
   bar();
   log.success(`Logged in as ${userInfo.email}`);
 
-  const claimStatus = await handleAutoclaim(process.cwd());
+  const claimResult = await handleAutoclaim(process.cwd());
 
   if (showNextSteps) {
-    outro(await loginNextSteps(claimStatus));
+    outro(await loginNextSteps(claimResult));
   } else {
     outro("Done");
   }
@@ -132,24 +132,31 @@ const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
   failed: "Auto-claim failed due to a temporary error. It will be retried on your next login.",
 };
 
-async function handleAutoclaim(cwd: string): Promise<AutoclaimResult["status"]> {
+async function handleAutoclaim(cwd: string): Promise<AutoclaimResult> {
   const result = await attemptAutoclaim(cwd);
 
   if (result.status === "claimed") {
     const label = result.app.name || result.app.application_id;
     log.success(`Claimed and linked application: \`${label}\``);
+    if (!result.envPulled) {
+      log.warn(
+        "Could not refresh environment variables automatically. Run `clerk env pull` manually.",
+      );
+    }
   }
 
   const warning = CLAIM_WARNINGS[result.status];
   if (warning) log.warn(warning);
 
-  return result.status;
+  return result;
 }
 
-async function loginNextSteps(claimStatus: AutoclaimResult["status"]): Promise<readonly string[]> {
-  if (claimStatus === "claimed") return NEXT_STEPS.AUTOCLAIMED;
-  if (claimStatus === "failed") return NEXT_STEPS.AUTOCLAIM_RETRY;
-  if (claimStatus === "not_found" || claimStatus === "already_claimed") {
+async function loginNextSteps(result: AutoclaimResult): Promise<readonly string[]> {
+  if (result.status === "claimed") {
+    return result.envPulled ? NEXT_STEPS.AUTOCLAIMED : NEXT_STEPS.AUTOCLAIMED_NO_ENV;
+  }
+  if (result.status === "failed") return NEXT_STEPS.AUTOCLAIM_RETRY;
+  if (result.status === "not_found" || result.status === "already_claimed") {
     return NEXT_STEPS.AUTOCLAIM_MANUAL_LINK;
   }
 
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -10,6 +10,7 @@ import { isHuman } from "../../mode.ts";
 import { throwUserAbort } from "../../lib/errors.ts";
 import { intro, outro, bar, withSpinner } from "../../lib/spinner.ts";
 import { NEXT_STEPS } from "../../lib/next-steps.ts";
+import { attemptAutoclaim, type AutoclaimResult } from "../../lib/autoclaim.ts";
 import { openBrowser } from "../../lib/open.ts";
 import { cyan, dim } from "../../lib/color.ts";
 import { log } from "../../lib/log.ts";
@@ -113,19 +114,52 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
   bar();
   log.success(`Logged in as ${userInfo.email}`);
 
+  const claimStatus = await handleAutoclaim(process.cwd());
+
   if (showNextSteps) {
-    const linked = await resolveProfile(process.cwd());
-    if (linked) {
-      const appLabel = linked.profile.appName
-        ? `\`${linked.profile.appName}\` (${linked.profile.appId})`
-        : `\`${linked.profile.appId}\``;
-      log.success(`Linked to ${appLabel}`);
-      outro(NEXT_STEPS.LOGIN_LINKED);
-    } else {
-      outro(NEXT_STEPS.LOGIN);
-    }
+    outro(await loginNextSteps(claimStatus));
   } else {
     outro("Done");
   }
+
   return userInfo;
+}
+
+const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
+  not_found:
+    "Claim token is no longer valid — the application may have been claimed from the dashboard.",
+  already_claimed: "Unable to claim — your account does not have an active organization.",
+  failed: "Auto-claim failed due to a temporary error. It will be retried on your next login.",
+};
+
+async function handleAutoclaim(cwd: string): Promise<AutoclaimResult["status"]> {
+  const result = await attemptAutoclaim(cwd);
+
+  if (result.status === "claimed") {
+    const label = result.app.name || result.app.application_id;
+    log.success(`Claimed and linked application: \`${label}\``);
+  }
+
+  const warning = CLAIM_WARNINGS[result.status];
+  if (warning) log.warn(warning);
+
+  return result.status;
+}
+
+async function loginNextSteps(claimStatus: AutoclaimResult["status"]): Promise<readonly string[]> {
+  if (claimStatus === "claimed") return NEXT_STEPS.AUTOCLAIMED;
+  if (claimStatus === "failed") return NEXT_STEPS.AUTOCLAIM_RETRY;
+  if (claimStatus === "not_found" || claimStatus === "already_claimed") {
+    return NEXT_STEPS.AUTOCLAIM_MANUAL_LINK;
+  }
+
+  // not_keyless — normal login, check for existing linked profile
+  const linked = await resolveProfile(process.cwd());
+  if (!linked) return NEXT_STEPS.LOGIN;
+
+  const appLabel = linked.profile.appName
+    ? `\`${linked.profile.appName}\` (${linked.profile.appId})`
+    : `\`${linked.profile.appId}\``;
+  log.success(`Linked to ${appLabel}`);
+  return NEXT_STEPS.LOGIN_LINKED;
 }
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -135,7 +135,8 @@ const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
   not_found:
     "Claim token is no longer valid - the application may have been claimed from the dashboard.",
   no_organization: "Unable to claim - your account does not have an active organization.",
-  failed: "Auto-claim failed due to a temporary error. It will be retried on your next login.",
+  failed:
+    "Auto-claim failed due to a temporary error. It will be retried on your next `clerk auth login`.",
 };
 
 async function handleAutoclaim(cwd: string): Promise<AutoclaimResult> {
--- a/packages/cli-core/src/commands/init/heuristics.ts
+++ b/packages/cli-core/src/commands/init/heuristics.ts
@@ -151,9 +151,9 @@ export async function isAuthenticated(): Promise<boolean> {
   return (await getToken()) != null;
 }
 
-export function printKeylessInfo(): void {
+export function printKeylessInfo(envFile: string): void {
   const lines = [
-    "\n  Your app is ready with development keys in .env.local.",
+    `\n  Your app is ready with development keys in ${envFile}.`,
     `  When you're ready, run ${bold("clerk auth login")} and your app will be claimed automatically.\n`,
   ];
   log.info(lines.map(dim).join("\n"));
--- a/packages/cli-core/src/commands/init/heuristics.ts
+++ b/packages/cli-core/src/commands/init/heuristics.ts
@@ -153,12 +153,8 @@ export async function isAuthenticated(): Promise<boolean> {
 
 export function printKeylessInfo(): void {
   const lines = [
-    "\n  Your app will work immediately — Clerk generates temporary dev keys automatically.",
-    `  Look for the ${bold('"Configure your application"')} banner to claim your account.\n`,
-    "  To connect a Clerk account later:",
-    "    clerk auth login",
-    "    clerk link",
-    "    clerk env pull",
+    "\n  Your app is ready with development keys in .env.local.",
+    `  When you're ready, run ${bold("clerk auth login")} and your app will be claimed automatically.\n`,
   ];
   log.info(lines.map(dim).join("\n"));
 }
--- a/packages/cli-core/src/commands/init/index.ts
+++ b/packages/cli-core/src/commands/init/index.ts
@@ -7,6 +7,12 @@ import { throwUserAbort, CliError } from "../../lib/errors.js";
 import { lookupFramework, type FrameworkInfo } from "../../lib/framework.js";
 import { resolveProfile } from "../../lib/config.js";
 import { log } from "../../lib/log.js";
+import {
+  createAccountlessApp,
+  writeKeysToEnvFile,
+  parseClaimToken,
+  writeKeylessBreadcrumb,
+} from "../../lib/keyless.js";
 import { printNextSteps } from "../../lib/next-steps.js";
 import { gatherContext, hasPackageJson } from "./context.js";
 import { scaffold, enrichProjectContext } from "./scaffold.js";
@@ -117,7 +123,7 @@ export async function init(options: InitOptions = {}) {
   } else if (!keyless) {
     await pull({ file: ctx.envFile });
   } else {
-    printKeylessInfo();
+    await setupKeylessApp(ctx.cwd, ctx.framework.dep);
   }
 
   if (options.skills !== false) {
@@ -274,6 +280,28 @@ async function authenticateAndLink(cwd: string, app: string | undefined): Promis
   }
 
   await link({ skipIfLinked: true, app, cwd });
+}
+
+// --- Keyless app setup ---
+
+async function setupKeylessApp(cwd: string, frameworkDep: string): Promise<void> {
+  try {
+    const app = await withSpinner("Creating development application...", () =>
+      createAccountlessApp(frameworkDep),
+    );
+
+    await writeKeysToEnvFile(cwd, {
+      publishableKey: app.publishable_key,
+      secretKey: app.secret_key,
+    });
+
+    await writeKeylessBreadcrumb(cwd, parseClaimToken(app.claim_url));
+  } catch (error) {
+    const msg = error instanceof Error ? error.message : String(error);
+    log.debug(`Could not create accountless app: ${msg}`);
+  }
+
+  printKeylessInfo();
 }
 
 // --- Detect & install ---
--- a/packages/cli-core/src/commands/init/index.ts
+++ b/packages/cli-core/src/commands/init/index.ts
@@ -3,7 +3,7 @@ import { link } from "../link/index.js";
 import { pull } from "../env/pull.js";
 import { isAgent } from "../../mode.js";
 import { dim, bold } from "../../lib/color.js";
-import { throwUserAbort, CliError } from "../../lib/errors.js";
+import { throwUserAbort, CliError, errorMessage } from "../../lib/errors.js";
 import { lookupFramework, type FrameworkInfo } from "../../lib/framework.js";
 import { resolveProfile } from "../../lib/config.js";
 import { log } from "../../lib/log.js";
@@ -297,8 +297,7 @@ async function setupKeylessApp(cwd: string, frameworkDep: string): Promise<void>
 
     await writeKeylessBreadcrumb(cwd, parseClaimToken(app.claim_url));
   } catch (error) {
-    const msg = error instanceof Error ? error.message : String(error);
-    log.debug(`Could not create accountless app: ${msg}`);
+    log.debug(`Could not create accountless app: ${errorMessage(error)}`);
   }
 
   printKeylessInfo();
--- a/packages/cli-core/src/commands/init/index.ts
+++ b/packages/cli-core/src/commands/init/index.ts
@@ -123,7 +123,7 @@ export async function init(options: InitOptions = {}) {
   } else if (!keyless) {
     await pull({ file: ctx.envFile });
   } else {
-    await setupKeylessApp(ctx.cwd, ctx.framework.dep);
+    await setupKeylessApp(ctx.cwd, ctx.framework.dep, ctx.envFile);
   }
 
   if (options.skills !== false) {
@@ -284,7 +284,7 @@ async function authenticateAndLink(cwd: string, app: string | undefined): Promis
 
 // --- Keyless app setup ---
 
-async function setupKeylessApp(cwd: string, frameworkDep: string): Promise<void> {
+async function setupKeylessApp(cwd: string, frameworkDep: string, envFile: string): Promise<void> {
   try {
     const app = await withSpinner("Creating development application...", () =>
       createAccountlessApp(frameworkDep),
@@ -296,11 +296,13 @@ async function setupKeylessApp(cwd: string, frameworkDep: string): Promise<void>
     });
 
     await writeKeylessBreadcrumb(cwd, parseClaimToken(app.claim_url));
+    printKeylessInfo(envFile);
   } catch (error) {
     log.debug(`Could not create accountless app: ${errorMessage(error)}`);
+    log.warn(
+      "Could not set up development keys. Run `clerk auth login` then `clerk link` to connect your app manually.",
+    );
   }
-
-  printKeylessInfo();
 }
 
 // --- Detect & install ---
--- a/packages/cli-core/src/commands/init/index.ts
+++ b/packages/cli-core/src/commands/init/index.ts
@@ -299,8 +299,12 @@ async function setupKeylessApp(cwd: string, frameworkDep: string, envFile: strin
     printKeylessInfo(envFile);
   } catch (error) {
     log.debug(`Could not create accountless app: ${errorMessage(error)}`);
+    const isTimeout = error instanceof Error && error.name === "AbortError";
+    const prefix = isTimeout
+      ? "Could not reach api.clerk.com within 15s."
+      : "Could not set up development keys.";
     log.warn(
-      "Could not set up development keys. Run `clerk auth login` then `clerk link` to connect your app manually.",
+      `${prefix} Run \`clerk auth login\` then \`clerk link\` to connect your app manually.`,
     );
   }
 }
--- a/packages/cli-core/src/lib/autoclaim.ts
+++ b/packages/cli-core/src/lib/autoclaim.ts
@@ -1,4 +1,4 @@
-import { basename } from "node:path";
+import { basename, join } from "node:path";
 import { readKeylessBreadcrumb, clearKeylessBreadcrumb } from "./keyless.ts";
 import { claimApplication, type Application } from "./plapi.ts";
 import { PlapiError, errorMessage } from "./errors.ts";
@@ -7,7 +7,7 @@ import { pull } from "../commands/env/pull.ts";
 import { log } from "./log.ts";
 
 type Claimed = { status: "claimed"; app: Application; envPulled: boolean };
-type Terminal = { status: "not_found" | "already_claimed" };
+type Terminal = { status: "not_found" | "no_organization" };
 type Failed = { status: "failed"; error: Error };
 type Skipped = { status: "not_keyless" };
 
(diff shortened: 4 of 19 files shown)


wyattjoh (commented):
## Code Review — PR #157

### Blocker

**B1. `.clerk/keyless.json` stores a live claim token with no `.gitignore` protection** _(codex confirmed)_
`packages/cli-core/src/lib/keyless.ts:88-106` writes `claimToken` under `.clerk/` in the user's project. Nothing in this PR (or `clerk init` today) appends `.clerk/` to the project's `.gitignore`. A developer can `git add .` and commit the token to a public repo, at which point any third party can claim the app before the original developer logs in, hijacking an app that will later hold production workloads.

Suggestion: in `setupKeylessApp` or `writeKeylessBreadcrumb`, append `.clerk/` to `.gitignore` if not already present (mirroring what `create-next-app` and similar tools do). Add a test that verifies the `.gitignore` update.

**B2. Keyless init writes `.env.local` but post-claim `pull` writes the framework default, producing two env files with drifting keys**
This is a cross-file interaction with existing code rather than a change in this PR alone, but the PR is where the regression becomes user-visible:
- `packages/cli-core/src/lib/keyless.ts:57-76` hardcodes `const targetFile = join(cwd, ".env.local")`.
- `packages/cli-core/src/lib/framework.ts` (unchanged) declares `envFile: ".env"` for Next.js, Astro, and Nuxt.
- `packages/cli-core/src/commands/env/pull.ts:30-46` (`resolveTargetFile`) prefers the framework-detected file.

Net effect on a Next.js project: `.env.local` written at init, `.env` written by pull after claim, same keys, different values.

Suggestion: use `detectEnvFile(cwd)` in `writeKeysToEnvFile` so there is one canonical target. Also update the JSDoc that references `.env.local`.

**B3. Missing changeset**
`git diff main..HEAD -- .changeset/` is empty. This is a `feat:` PR touching `@clerk/cli-core` runtime. The `Enforce Changeset` workflow will block merge.

### Major

**M1. App name sent to backend uses `basename(cwd).slice(0, 50)` with no sanitization** _(codex confirmed)_
`packages/cli-core/src/lib/autoclaim.ts:32`:
- Leaks potentially-sensitive directory names to the Clerk platform API.
- `.slice(50)` truncates on UTF-16 code units; an emoji or non-BMP char at the boundary produces an orphaned surrogate (400 on strict UTF-8 validators).
- Backend-invalid chars are not sanitized.

Suggestion: derive the name from `package.json#name` or the git repo name with `basename(cwd)` as a fallback, and sanitize + byte-length-truncate before sending.

**M2. `attemptAutoclaim(cwd)` threads `cwd` but `tryPullEnv` ignores it**
`autoclaim.ts:37-50` accepts a `cwd` argument for the breadcrumb but calls `pull({})`, which uses `process.cwd()` internally (`pull.ts:53`). Today's single call site passes `process.cwd()` so it is benign, but this is fragile. Either thread `cwd` through `pull` options or assert at entry.

**M3. `createAccountlessApp` has no timeout or abort; `clerk init` hangs on captive portal / slow BAPI** _(codex confirmed)_
`packages/cli-core/src/lib/keyless.ts:35-52` and `packages/cli-core/src/commands/init/index.ts:272-287`: `fetch(...)` has no `AbortController`. On slow/broken networks, `withSpinner` hangs indefinitely. `log.debug` is only visible with `--verbose`, so users see nothing.

**M4. `setupKeylessApp` swallows all errors while success copy still renders** _(codex confirmed)_
`commands/init/index.ts:270-287` wraps the body in `try { ... } catch { log.debug(...) }`. Even on filesystem errors, `parseClaimToken` throws, or the fetch fails, `printKeylessInfo` still tells the user "Your app is ready with development keys in .env.local". Distinguish expected network failures from unexpected bugs and suppress the success message on failure.

**M5. `classifyClaimError` semantic mismatch and aggressive breadcrumb-clear**
`autoclaim.ts:62-74` maps 403 to `already_claimed`, but the message at `login.ts:128-132` says "your account does not have an active organization". Name and message disagree; rename the status (e.g. `no_organization`). Also, 404 unconditionally clears the breadcrumb; a transient DNS blip that returns 404 would destroy the retry signal.

### Minor

- `readKeylessBreadcrumb` (`keyless.ts:108-119`) has no schema validation; a malformed-but-JSON file yields `{claimToken: undefined}` and `autoclaim.ts:33` will POST `{token: undefined}`.
- `parseClaimToken` uses `https://placeholder.com` as the synthetic URL base (`keyless.ts:81-87`); prefer `https://example.invalid` (RFC 6761).
- Em-dashes added in `login.ts:130-131,163` and `autoclaim.ts:7` (global style rule).
- `handleAutoclaim` prints `log.success` then `log.warn` then `outro` (`login.ts:117-124,135-148`); noisy ordering.
- No retry for 5xx / 429 despite the design claim that "transient failures preserve breadcrumb for retry on next login" (`autoclaim.ts:28-36`); a short in-process retry with backoff would save users a manual re-login.

### Nits

- `CLAIM_WARNINGS` could be an explicit object-literal map of the three warning statuses.
- `log.info("Environment variables written to .env.local")` will be wrong once B2 is fixed.
- Test `expect(parsed.name).toBeTruthy()` (`autoclaim.test.ts:164`) is too loose; assert the exact truncation.
- CI refactor (composite action, `workflow_call`) is bundled into a `feat:` PR; splitting would simplify bisect.

### Positives

`AutoclaimResult` discriminated union is cleanly typed and enables exhaustive handling in `loginNextSteps`. `attemptAutoclaim` never throws (the type system enforces it). `linkApp` extraction in `autolink.ts` is a good refactor. Test coverage for the orchestrator is thorough: all status paths, breadcrumb clear/preserve semantics, env-pull success/failure, `linkApp` invocation, and request body shape.

packages/cli-core/src/commands/init/heuristics.ts
```diff
@@ -138,12 +138,8 @@ export async function getAuthenticatedEmail(): Promise<string | null> {
 
 export function printKeylessInfo(): void {
   const lines = [
-    "\n  Your app will work immediately — Clerk generates temporary dev keys automatically.",
-    `  Look for the ${bold('"Configure your application"')} banner to claim your account.\n`,
-    "  To connect a Clerk account later:",
-    "    clerk auth login",
-    "    clerk link",
-    "    clerk env pull",
+    "\n  Your app is ready with development keys in .env.local.",
```
wyattjoh: **[HIGH] printKeylessInfo hardcodes .env.local, but Next.js/Astro/Nuxt write to .env**

The new message says `Your app is ready with development keys in .env.local.` But `writeKeysToEnvFile` (in `lib/keyless.ts`) resolves the target via `detectEnvFile(cwd)`, which returns `.env` for Next.js, Astro, and Nuxt per `FRAMEWORK_MAP`. A Next.js user following this output will look in the wrong file.

Consider accepting the resolved `envFile` as a parameter (the init flow already has it in `ctx.envFile`) and interpolating it into the message. At minimum, replace the hardcoded `.env.local` with a generic phrase.

heuristics.ts:
  "\n  Your app is ready with development keys in .env.local.",

framework.ts FRAMEWORK_MAP:
  { dep: "next",   envFile: ".env",       supportsKeyless: true },
  { dep: "astro",  envFile: ".env",       supportsKeyless: true },
  { dep: "nuxt",   envFile: ".env",       supportsKeyless: true },

</details>

packages/cli-core/src/commands/env/pull.ts
```diff
@@ -50,7 +51,7 @@ async function resolveTargetFile(
 
 export async function pull(options: EnvPullOptions): Promise<void> {
   const ctx = await resolveAppContext(options);
-  const cwd = process.cwd();
+  const cwd = options.cwd ?? process.cwd();
```
wyattjoh: **[HIGH] pull({ cwd }) threads cwd partially — resolveAppContext still uses process.cwd()**

The new `cwd` option is honored for `detectEnvFile`, `resolveTargetFile`, and `detectPublishableKeyName`, but `resolveAppContext(options)` at line 52 still calls `resolveProfile(process.cwd())` internally (see `lib/config.ts:287`). If a future caller passes a `cwd` that differs from `process.cwd()`, profile lookup silently uses the wrong directory.

Today the only consumer (`autoclaim.attemptAutoclaim`) passes `process.cwd()`, so there is no live bug — but the API is misleading. Either plumb `cwd` through `resolveAppContext` → `resolveProfile`, or drop the `cwd` option entirely and document that `pull` uses `process.cwd()`. The lean fix is the latter since the only caller happens to be same-cwd.

pull.ts:
  export async function pull(options: EnvPullOptions): Promise<void> {
    const ctx = await resolveAppContext(options);   // <-- options.cwd ignored here
    const cwd = options.cwd ?? process.cwd();
    ...

config.ts (resolveAppContext):
  const resolved = await resolveProfile(process.cwd());  // hardcoded

</details>

packages/cli-core/src/lib/autoclaim.ts
```diff
@@ -0,0 +1,94 @@
...
+type Failed = { status: "failed"; error: Error };
+type Skipped = { status: "not_keyless" };
+
+export type AutoclaimResult = Claimed | Terminal | Failed | Skipped;
+
+type ClaimAttempt = { status: "claimed"; app: Application } | Terminal | Failed;
+
+const APP_NAME_MAX_BYTES = 50;
+
+const TERMINAL_BY_STATUS: Record<number, Terminal["status"]> = {
```
wyattjoh, line 20: **[MEDIUM] Non-404/403 PlapiErrors preserve breadcrumb → infinite retry on terminal 4xx**

`classifyClaimError` only treats 404 and 403 as terminal. A 400 (invalid token), 409 (already claimed elsewhere), or 422 (validation) all fall through to `failed`, which preserves the breadcrumb and retries on every subsequent `clerk auth login`. The user sees the same "Auto-claim failed due to a temporary error" warning repeatedly with no path to recovery short of manually deleting `.clerk/keyless.json`.

Broaden `TERMINAL_BY_STATUS` to cover all 4xx except 408/429, or explicitly list 400, 409, 422 alongside 403/404. 5xx and network errors should remain retryable.

const TERMINAL_BY_STATUS: Record<number, Terminal["status"]> = {
  404: "not_found",
  403: "no_organization",
};

function classifyClaimError(error: unknown): Terminal | Failed {
  if (error instanceof PlapiError && error.status in TERMINAL_BY_STATUS) { ... }
  // everything else -> Failed, breadcrumb preserved
  return { status: "failed", error: err };
}

</details>

packages/cli-core/src/lib/keyless.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  await Bun.write(breadcrumbPath(cwd), JSON.stringify(breadcrumb, null, 2) + "\n");
+  await ensureGitignoreEntry(cwd, BREADCRUMB_DIR + "/");
+  log.debug(`Wrote keyless breadcrumb to ${BREADCRUMB_DIR}/${BREADCRUMB_FILE}`);
+}
+
+export async function readKeylessBreadcrumb(cwd: string): Promise<KeylessBreadcrumb | undefined> {
+  try {
+    const data: unknown = await Bun.file(breadcrumbPath(cwd)).json();
+    return isKeylessBreadcrumb(data) ? data : undefined;
```
wyattjoh: **[MEDIUM] Corrupted keyless.json is silently ignored forever**

`readKeylessBreadcrumb` swallows parse errors and shape-validation failures, returning `undefined`. A corrupted `keyless.json` (truncated write, user edit, wrong shape) will cause every subsequent `clerk auth login` to report `not_keyless` silently — the user never learns why auto-claim is skipped.

When `isKeylessBreadcrumb(data)` returns false, call `clearKeylessBreadcrumb(cwd)` (and `log.debug` the reason). That way the file is recoverable by re-running `clerk init --keyless`, and the state doesn't persist indefinitely.

export async function readKeylessBreadcrumb(cwd: string): Promise<KeylessBreadcrumb | undefined> {
  try {
    const data: unknown = await Bun.file(breadcrumbPath(cwd)).json();
    return isKeylessBreadcrumb(data) ? data : undefined;  // malformed file stays on disk
  } catch {
    return undefined;
  }
}

</details>

packages/cli-core/src/commands/init/index.ts
```diff
@@ -261,6 +267,26 @@ async function authenticateAndLink(cwd: string, app: string | undefined): Promis
...
+      publishableKey: app.publishable_key,
+      secretKey: app.secret_key,
+    });
+
+    await writeKeylessBreadcrumb(cwd, parseClaimToken(app.claim_url));
+    printKeylessInfo();
+  } catch (error) {
+    log.debug(`Could not create accountless app: ${errorMessage(error)}`);
+  }
+}
```
wyattjoh, line 310: **[MEDIUM] setupKeylessApp swallows creation failures at debug log level**

When `createAccountlessApp` fails (BAPI down, network error, timeout), the catch block only emits `log.debug`, which is suppressed without `--verbose`. The user sees zero output for the keyless step: no warning, no next-steps, nothing. `printKeylessInfo()` is never reached. The `outro("Done")` still prints. The user then runs `bun dev` and wonders why auth doesn't work.

Escalate the failure case to `log.warn` and print a brief recovery hint (`clerk auth login` then `clerk link`). `log.debug` is fine for the underlying error detail, but the user needs to know the keyless step failed.

async function setupKeylessApp(cwd: string, frameworkDep: string): Promise<void> {
  try {
    const app = await withSpinner("Creating development application...", () => ...);
    await writeKeysToEnvFile(...);
    await writeKeylessBreadcrumb(...);
    printKeylessInfo();
  } catch (error) {
    log.debug(`Could not create accountless app: ${errorMessage(error)}`);  // silent to the user
  }
}

</details>

packages/cli-core/src/lib/keyless.ts
```diff
@@ -0,0 +1,142 @@
...
+      body: body.toString(),
+      signal: controller.signal,
+    });
+  } finally {
+    clearTimeout(timer);
+  }
+
+  if (!response.ok) {
+    const text = await response.text();
+    throw new Error(`Failed to create accountless application (${response.status}): ${text}`);
```
wyattjoh: **[MEDIUM] createAccountlessApp bypasses the project's error-handling protocol**

The BAPI call throws a plain `new Error(...)` with the raw status/body. Per `.claude/rules/errors.md`, API errors should use a typed error class that the global handler recognizes (`BapiError`/`PlapiError`) and be wrapped with `withApiContext` for user-facing context. As written, a non-OK response surfaces as `Failed to create accountless application (500): <raw body>` which is both unstyled and leaks response body directly.

Prefer `BapiError` if one exists (the rules doc references `src/commands/api/bapi.ts`), or wrap with an error class the global handler can format.

if (!response.ok) {
  const text = await response.text();
  throw new Error(`Failed to create accountless application (${response.status}): ${text}`);
}

</details>

packages/cli-core/src/lib/errors.ts
```diff
@@ -43,6 +43,8 @@ export const ERROR_CODE = {
   CATALOG_ERROR: "catalog_error",
   /** Doctor checks found issues. */
   DOCTOR_FAILED: "doctor_failed",
+  /** Autoclaim of a keyless application failed. */
+  AUTOCLAIM_FAILED: "autoclaim_failed",
 } as const;
```
wyattjoh, line 46: **[LOW] AUTOCLAIM_FAILED error code added but never thrown**

`ERROR_CODE.AUTOCLAIM_FAILED` is added to the enum but nothing in `autoclaim.ts` throws a `CliError` with it. The autoclaim design is explicitly "never throws" and uses a discriminated-union return type, so there is no consumer for this code. Dead code — either remove it or wire it into the `failed` branch's Error.

errors.ts:
  /** Autoclaim of a keyless application failed. */
  AUTOCLAIM_FAILED: "autoclaim_failed",

autoclaim.ts never references ERROR_CODE.AUTOCLAIM_FAILED.

</details>

packages/cli-core/src/lib/autoclaim.ts
```diff
@@ -0,0 +1,94 @@
...
+async function deriveAppName(cwd: string): Promise<string> {
+  try {
+    const pkg: { name?: unknown } = await Bun.file(join(cwd, "package.json")).json();
+    if (typeof pkg.name === "string" && pkg.name.trim()) return pkg.name.trim();
+  } catch {
+    // fall through
+  }
+  return basename(cwd);
+}
+
```
wyattjoh, line 34: **[LOW] truncateToByteLength uses byte-slice + lenient decode; consider char-based truncation**

The current impl slices UTF-8 bytes, decodes with `fatal: false`, and strips a trailing U+FFFD replacement char. This works but is brittle: it creates an extra buffer copy and the intent is non-obvious. Per the PR body the backend limit is "50 chars". If that is codepoints (likely) rather than UTF-8 bytes, `Array.from(str).slice(0, 50).join('')` is both simpler and matches the "chars" framing.

Worth confirming with the backend team what the 50 actually counts — bytes, codepoints, or JS-string-length — and matching that exactly. Emoji/CJK app names are the edge case.

const APP_NAME_MAX_BYTES = 50;

function truncateToByteLength(str: string, maxBytes: number): string {
  const encoded = new TextEncoder().encode(str);
  if (encoded.length <= maxBytes) return str;
  return new TextDecoder("utf-8", { fatal: false })
    .decode(encoded.slice(0, maxBytes))
    .replace(/\uFFFD$/, "");
}

</details>

packages/cli-core/src/lib/keyless.ts
```diff
@@ -0,0 +1,142 @@
...
+    [publishableKeyName]: keys.publishableKey,
+    [secretKeyName]: keys.secretKey,
+  });
+
+  await Bun.write(targetFile, serializeEnvFile(merged));
+  log.info(`Environment variables written to ${envFile}`);
+}
+
+export function parseClaimToken(claimUrl: string): string {
+  const base = claimUrl.startsWith("http") ? undefined : "https://example.invalid";
```
wyattjoh, line 95: **[LOW] example.invalid base URL sentinel deserves a one-line comment**

`parseClaimToken` uses `https://example.invalid` as a base URL so WHATWG `URL` can parse relative paths like `/apps/claim?token=abc`. The pattern is correct but the intent is non-obvious to readers who haven't hit the WHATWG URL relative-parsing quirk. A one-liner (`// Need a dummy base because WHATWG URL rejects relative URLs without one`) would save the next reader a confused minute.

export function parseClaimToken(claimUrl: string): string {
  const base = claimUrl.startsWith("http") ? undefined : "https://example.invalid";
  const token = new URL(claimUrl, base).searchParams.get("token");
  ...
}

</details>

packages/cli-core/src/lib/plapi.ts
```diff
@@ -118,56 +112,24 @@ export interface Application {
...
+    search: { include_secret_keys: true },
   });
-
-  if (!response.ok) {
-    const body = await response.text();
-    throw new PlapiError(response.status, body, url.toString());
-  }
-
-  return response.json() as Promise<Application>;
 }
```
wyattjoh, line 135: **[LOW] include_secret_keys=true now serializes as the string "true"**

`fetchApplication` previously called `url.searchParams.set("include_secret_keys", "true")`. The refactored version passes `search: { include_secret_keys: true }`, and `applySearch` handles `value === true` by writing `"true"`. Behavior is identical to the old code (both produce `?include_secret_keys=true`), but worth a targeted spot-check against a live PLAPI instance to confirm the backend parses the query string identically after the refactor — especially for the `destructive` flag in `sendInstanceConfig`, which is only set when truthy and now skipped via the `value === undefined || value === false` short-circuit in `applySearch`.

function applySearch(url: URL, params: Record<string, SearchValue>): void {
  for (const [key, value] of Object.entries(params)) {
    if (value === undefined || value === false) continue;
    ...
    url.searchParams.set(key, value === true ? "true" : value);
  }
}

</details>

wyattjoh (commented):
## Code review

Substantial new surface (keyless.ts, autoclaim.ts) plus a clean plapi.ts refactor. Core design — discriminated AutoclaimResult, never-throws contract, breadcrumb-preserve-on-transient — is sound. A handful of issues worth tightening before merge: one user-facing inaccuracy (`.env.local` message), one misleading API parameter (`pull({ cwd })`), and a few error-handling edges (retry loops, corrupted breadcrumb recovery, silent keyless creation failures).

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -113,19 +114,58 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
   bar();
   log.success(`Logged in as ${userInfo.email}`);
 
+  const claimResult = await handleAutoclaim(process.cwd());
```
wyattjoh, line 123: `handleAutoclaim` only runs after `performOAuthFlow()` completes. But above this point in `login()`:

- In non-human mode (agent), an existing valid session returns early at `return existingSession;`.
- In human mode, if the user declines the re-auth confirm, `throwUserAbort()` fires.

In both cases, a `.clerk/keyless.json` breadcrumb written by a prior `clerk init -y` will never be consumed. The user (or agent) keeps running on temporary keyless dev keys until they manually delete the breadcrumb or somehow force a full OAuth flow again.

The intended flow (`clerk init -y` then first-time `clerk auth login`) works fine, but all of these common situations fall through:
- User already has a session from a previous project and runs `clerk init -y` in a new directory, then `clerk auth login` there.
- An AI agent is OAuth'd and runs `clerk init -y` + `clerk auth login` in sequence.
- User runs `clerk auth login` twice and declines re-auth the second time.

Consider running `attemptAutoclaim(process.cwd())` on the existing-session branches too (before `return existingSession;` around line 97, and, if you want to be generous, when the user declines re-auth rather than aborting outright). At minimum the agent-mode existing-session path should trigger auto-claim, since that's the primary bootstrap target.

packages/cli-core/src/lib/keyless.ts
```diff
@@ -0,0 +1,147 @@
...
+  };
+
+  const body = new URLSearchParams({ source: "cli" });
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), CREATE_TIMEOUT_MS);
+
+  let response: Response;
+  try {
+    response = await fetch(url, {
+      method: "POST",
```
wyattjoh, line 52: `createAccountlessApp` calls `fetch(url, ...)` directly. The project rule in `.claude/rules/debug-logging.md` is explicit: "All outbound HTTP in library code uses `loggedFetch` from `src/lib/fetch.ts`" and "Never call `fetch()` directly in library code." This means `--verbose` will not show the BAPI request URL or the error body when accountless creation fails, which is exactly the diagnostic channel we want when a user reports `clerk init -y` not writing keys.

Swap to:

```ts
const response = await loggedFetch(url, {
  tag: "bapi",
  method: "POST",
  headers,
  body: body.toString(),
  signal: controller.signal,
});
```

The abort-timer wrapping stays the same. `loggedFetch` already emits the request and non-ok response body at debug level, so the call site does not need its own `log.debug`.

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -113,19 +114,58 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
...
+
+async function handleAutoclaim(cwd: string): Promise<AutoclaimResult> {
+  const result = await attemptAutoclaim(cwd);
+
+  if (result.status === "claimed") {
+    const label = result.app.name || result.app.application_id;
+    log.success(`Claimed and linked application: \`${label}\``);
+    if (!result.envPulled) {
+      log.warn(
+        "Could not refresh environment variables automatically. Run `clerk env pull` manually.",
```
wyattjoh: When `status === "claimed"` and `envPulled === false`, `handleAutoclaim` emits:

1. `log.success("Claimed and linked application: ...")`
2. `log.warn("Could not refresh environment variables automatically. Run `clerk env pull` manually.")`

Then `loginNextSteps` returns `NEXT_STEPS.AUTOCLAIMED_NO_ENV`, which prints:

- `Run `clerk env pull` to refresh your environment variables`
- `Run `clerk doctor` to verify your setup`

So the user sees the "run env pull" guidance twice back-to-back, once as a warning and once as a next-step. Pick one: either drop the inline `log.warn` (next-steps already covers it) or drop the `Run clerk env pull` line from `AUTOCLAIMED_NO_ENV` and keep the warn for the attention-grabbing case.

packages/cli-core/src/lib/autoclaim.ts
```diff
@@ -0,0 +1,98 @@
...
+    await pull({});
+    return true;
+  } catch (error) {
+    log.debug(`Auto env pull failed: ${errorMessage(error)}`);
+    return false;
+  }
+}
+
+function classifyClaimError(error: unknown): Terminal | Failed {
+  if (error instanceof PlapiError) {
```
wyattjoh:
```ts
if (error.status >= 400 && error.status < 500 && error.status !== 408 && error.status !== 429) {
  log.debug(`Claim returned non-retryable ${error.status}: treating as not_found`);
  return { status: "not_found" };
}
```

Two problems:

1. `400 Bad Request` on a malformed token payload, `401 Unauthorized` from a stale OAuth token, and `409 Conflict` from a race all get bucketed as `not_found` and cause the breadcrumb to be cleared. 401 in particular is recoverable by re-login; silently treating it as "app not found" and wiping the breadcrumb means the user loses claim state without ever learning their session was expired.
2. The status name is a lie once the catch-all fires on e.g. 400/401. Users then see the `"Claim token is no longer valid"` warning from `CLAIM_WARNINGS.not_found`, which is wrong for those cases.

Recommend narrowing the whitelist of explicit terminal codes (404, 403, maybe 410) and letting everything else fall through to `failed` so the breadcrumb is preserved for the next login. If you want a broader "some 4xx means give up" guard, introduce a third `unknown_error` terminal with a neutral warning rather than pretending it is 404.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -276,6 +282,29 @@ async function authenticateAndLink(cwd: string, app: string | undefined): Promis
   await link({ skipIfLinked: true, app });
```
wyattjoh: `clerk init` now creates an accountless application via BAPI and writes a `.clerk/keyless.json` breadcrumb in the keyless skip-auth path; that is a new API call and a new on-disk artifact. The project rule in `.claude/rules/commands.md` ("When modifying a command's behavior, options, or API calls, update its README to match") applies here. Please update `packages/cli-core/src/commands/init/README.md` to mention the `POST /v1/accountless_applications` call and the `.clerk/keyless.json` breadcrumb, and do the same for the login README if one exists for `auth/login` (new `POST /v1/platform/accountless_applications/claim` endpoint + breadcrumb lifecycle).

packages/cli-core/src/lib/autoclaim.ts
```diff
@@ -0,0 +1,98 @@
...
+export type AutoclaimResult = Claimed | Terminal | Failed | Skipped;
+
+type ClaimAttempt = { status: "claimed"; app: Application } | Terminal | Failed;
+
+const APP_NAME_MAX_CHARS = 50;
+
+const TERMINAL_BY_STATUS: Record<number, Terminal["status"]> = {
+  404: "not_found",
+  403: "no_organization",
+};
```
wyattjoh, line 23:
```ts
function truncateToChars(str: string, max: number): string {
  const chars = Array.from(str);
  return chars.length <= max ? str : chars.slice(0, max).join("");
}
```

`Array.from(str)` splits on code points (emoji-safe), but still splits grapheme clusters: a base character + combining mark can end up with the base kept and the mark dropped, producing a visually different character. For a 50-char package name this is extremely unlikely to bite in practice, so this is pure nitpick, but if it matters, `Intl.Segmenter` with `granularity: "grapheme"` is the right primitive. Also worth asking whether the truncation should happen server-side via BAPI validation rather than being mirrored in the CLI; the comment about matching the backend validation limit is the classic duplicated-invariant smell.

wyattjoh (commented):
Solid feature overall. The main concerns are an existing-session path that bypasses auto-claim entirely, a raw `fetch()` call that skips the project's `loggedFetch` wrapper, a success-log UX wart where both a success line and a `Run env pull manually` warn can print together, and an overly broad 4xx-is-terminal classifier. See inline comments.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -276,6 +282,29 @@ async function authenticateAndLink(cwd: string, app: string | undefined): Promis
   await link({ skipIfLinked: true, app, cwd });
 }
 
+// --- Keyless app setup ---
+
+async function setupKeylessApp(cwd: string, frameworkDep: string, envFile: string): Promise<void> {
+  try {
+    const app = await withSpinner("Creating development application...", () =>
+      createAccountlessApp(frameworkDep),
```
wyattjoh, line 290: The new `setupKeylessApp` path calls `createAccountlessApp(frameworkDep)` with no way for tests to stub it in-tree. `packages/cli-core/src/commands/init/index.test.ts` (unchanged in this PR, but exercised by the new code) has a test `"blank dir with keyless framework auto-selects keyless when unauthenticated"` that reaches this path. Running `bun test packages/cli-core/src/commands/init/index.test.ts` against this branch:

1. Issues a real POST to `api.clerk.com/v1/accountless_applications` (a new accountless app is minted on Clerk production on every test run),
2. Writes `pk_test_...`/`sk_test_...` to `/tmp/test/.env.local`,
3. Leaves `/tmp/test/.clerk/keyless.json` with a live claim token on the host.

This is flaky (needs network + healthy BAPI), costly (creates real apps), and a minor secrets-on-disk hygiene problem. Fix options: (a) export a seam so the test can `spyOn(keylessMod, "createAccountlessApp")` and `spyOn(keylessMod, "writeKeylessBreadcrumb")`, or (b) wrap the call in a helper that tests can mock, or (c) add `stubFetch(...)` to the existing init test suite setup. Given the file-per-subprocess test isolation, a top-level `mock.module("../../lib/keyless.ts", ...)` in `index.test.ts` is the cleanest fix.

packages/cli-core/src/lib/autoclaim.ts
```diff
@@ -0,0 +1,96 @@
...
+  const rawName = await deriveAppName(cwd);
+  const appName = truncateToChars(rawName, APP_NAME_MAX_CHARS);
+  const result = await tryClaim(breadcrumb.claimToken, appName);
+
+  if (result.status === "failed") return result;
+
+  await clearKeylessBreadcrumb(cwd);
+
+  if (result.status === "claimed") {
+    await linkApp(result.app, cwd);
```
wyattjoh: The docstring says "Orchestrates post-login claim of a keyless app. Never throws." but `linkApp(result.app, cwd)` on line 59 is not wrapped in try/catch. `linkApp` calls `setProfile` → `writeConfig`, which performs disk I/O and can throw (EACCES, ENOSPC, malformed config, etc.). If it throws at this point, the server-side claim has succeeded **and the breadcrumb has already been cleared** on line 56. The login command has no catch — the thrown error bubbles up and the user sees a failed login despite the claim having actually landed on the server, with no recovery path (the breadcrumb is gone).

Either wrap `linkApp` and `clearKeylessBreadcrumb` inside a try/catch that still returns `{status: "claimed", envPulled: false}` plus a warning, or clear the breadcrumb only after `linkApp` succeeds (keeps the invariant: breadcrumb cleared → client state consistent with server).

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -113,19 +120,53 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
...
   }
+
   return userInfo;
 }
+
+const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
+  not_found:
+    "Claim token is no longer valid - the application may have been claimed from the dashboard.",
+  no_organization: "Unable to claim - your account does not have an active organization.",
+  failed: "Auto-claim failed due to a temporary error. It will be retried on your next login.",
```
wyattjoh: `CLAIM_WARNINGS.failed` says "It will be retried on your next login." but `handleAutoclaim` is run on *every* invocation of `clerk auth login` that reaches the end of the flow, including the agent-mode existing-session branch. A user who already has a valid session and keeps getting transient 5xxs from the claim endpoint will see this message on every re-run, which is confusing — they are already "logged in". Consider "It will be retried on your next `clerk auth login`." or drop the claim about when it retries.

packages/cli-core/src/lib/keyless.ts
```diff
@@ -0,0 +1,149 @@
...
+  await Bun.write(breadcrumbPath(cwd), JSON.stringify(breadcrumb, null, 2) + "\n");
+  log.debug(`Wrote keyless breadcrumb to ${BREADCRUMB_DIR}/${BREADCRUMB_FILE}`);
+}
+
+export async function readKeylessBreadcrumb(cwd: string): Promise<KeylessBreadcrumb | undefined> {
+  try {
+    const data: unknown = await Bun.file(breadcrumbPath(cwd)).json();
+    if (isKeylessBreadcrumb(data)) return data;
+    log.debug("Keyless breadcrumb has wrong shape; clearing it to allow fresh setup");
+    await clearKeylessBreadcrumb(cwd);
```
wyattjoh, line 135: When `readKeylessBreadcrumb` finds a file it can parse as JSON but that fails `isKeylessBreadcrumb`, it silently calls `clearKeylessBreadcrumb` and returns undefined. If a future version adds new required fields (or an older CLI wrote a v0 shape), this "auto-repair" deletes state before the user notices. Two alternatives:

- Log at `log.warn` (not `log.debug`) so users running without `--verbose` see the repair happened.
- Leave the file in place and return undefined; the next keyless init overwrites it anyway.

The round-trip test (`read returns undefined and clears file when breadcrumb has wrong shape`) codifies the current behavior. If kept, upgrade the debug log to warn.

packages/cli-core/src/lib/keyless.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  const headers: Record<string, string> = {
+    "Content-Type": "application/x-www-form-urlencoded",
+    Accept: "application/json",
+    ...(framework && { "Clerk-Framework": framework }),
+  };
+
+  const body = new URLSearchParams({ source: "cli" });
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), CREATE_TIMEOUT_MS);
```
wyattjoh, line 46: `createAccountlessApp` uses a 15-second `AbortController` timeout. When it fires, the abort error bubbles up through `setupKeylessApp` and is caught as a generic warning: "Could not set up development keys. Run `clerk auth login` then `clerk link`...". The user has no signal that their network was the problem versus BAPI being down. A dedicated branch on `AbortError` (or `error.name === "AbortError"`) that says "Could not reach api.clerk.com within 15s" would make the failure mode actionable.

wyattjoh (approved):

PR #165, opened 2026-04-15
fix(init): register --app option shown in help examples

wyattjoh:
## Summary

`clerk init --app <id>` was advertised in the help examples and wired through to `authenticateAndLink`, but the option was never registered with Commander. Running it errored with `unknown option '--app'` instead of linking to the supplied app.

This PR:
- Registers `--app <id>` on the `init` command, matching the description already used by `clerk link`
- Adds two regression guards at different layers:
  - A completion-tree test asserting `--app` is part of `init`'s registered options
  - An integration test that invokes `clerk init --app app_123 --prompt` through the real Commander program and fails if the option isn't registered (verified by temporarily reverting the fix)
- Syncs the `init` command README and the root README's `clerk init` help block (which was also missing the `--app app_123` example)

No behavioral change beyond making the advertised flag actually reach the already-wired runtime path.

## Test plan

- [ ] `bun run src/cli.ts init --help` lists `--app <id>` in Options and Examples
- [ ] `bun run test` passes (both regression tests included)
- [ ] Manual: `clerk init --app app_xxx` in a test project skips the interactive picker

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ clerk init
   --framework <name>     Framework to set up (skips auto-detection)
   --pm <manager>         Package manager to use (skips prompt/auto-detection)
   --name <project-name>  Project name for --starter (skips prompt)
+  --app <id>             Application ID to link (skips interactive picker)
   --starter              Bootstrap a new project from a starter template
   --prompt               Output a prompt for an AI agent to integrate Clerk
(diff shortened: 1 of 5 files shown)


wyattjoh:
**Stack: init-app-option**

- **#165 👈 this PR**

dmoerner (approved):

PR #167, opened 2026-04-15
feat(changeset): enrich changelog entries with PR and author links

wyattjoh:
## Summary

Port the changelog generator from clerk/javascript so generated CHANGELOG.md sections include PR references, commit links, and contributor handles alongside each release line.

The generator uses the GitHub GraphQL API with a small concurrency limiter and in-memory cache to avoid secondary rate limits during `changeset version` runs in the Version Packages PR. `GITHUB_TOKEN` is already passed by `changesets/action` in `release.yml`, so no workflow changes are required.

## Test plan

- [ ] Merge a PR with a changeset and confirm the next Version Packages PR writes `(#NNN) by @user` style entries to CHANGELOG.md

### DIFF
--- a/.changeset/changelog.js
+++ b/.changeset/changelog.js
@@ -1,0 +1,253 @@
+const repo = "clerk/cli";
+const [owner, repoName] = repo.split("/");
+
+// Cache to avoid duplicate fetches for the same commit/PR
+const cache = new Map();
+
+// Simple concurrency limiter to avoid hitting GitHub secondary rate limits
+const MAX_CONCURRENT = 6;
+let active = 0;
+const queue = [];
+
+function withLimit(fn) {
+  return (...args) =>
+    new Promise((resolve, reject) => {
+      const run = async () => {
+        active++;
+        try {
+          resolve(await fn(...args));
+        } catch (e) {
+          reject(e);
+        } finally {
+          active--;
+          if (queue.length > 0) queue.shift()();
+        }
+      };
+      if (active < MAX_CONCURRENT) run();
+      else queue.push(run);
+    });
+}
+
+async function graphql(query) {
+  const token = process.env.GITHUB_TOKEN;
+  if (!token) {
+    throw new Error("GITHUB_TOKEN environment variable is required");
+  }
+
+  const res = await fetch("https://api.github.com/graphql", {
+    method: "POST",
+    headers: {
+      Authorization: `Token ${token}`,
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({ query }),
+  });
+
+  if (!res.ok) {
+    throw new Error(`GitHub API responded with ${res.status}: ${await res.text()}`);
+  }
+
+  const json = await res.json();
+  if (json.errors) {
+    throw new Error(`GitHub GraphQL error: ${JSON.stringify(json.errors, null, 2)}`);
+  }
+  if (!json.data) {
+    throw new Error(`Unexpected GitHub response: ${JSON.stringify(json)}`);
+  }
+  return json.data;
+}
+
+// Fetches commit info with a single small GraphQL query per commit
+const fetchCommitInfo = withLimit(async (commit) => {
+  const key = `commit:${commit}`;
+  if (cache.has(key)) return cache.get(key);
+
+  const data = await graphql(`query {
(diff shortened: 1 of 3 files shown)


wyattjoh:
**Stack: feat/changelog-with-pr-links**

- **#167 👈 this PR**

.changeset/changelog.js
```diff
@@ -0,0 +1,287 @@
+const repo = "clerk/cli";
```
rafa-thayto, line 1: The config passes `{ "repo": "clerk/cli" }` as options, but this file hardcodes `repo` here and never reads the options param. Might want to either wire up `options.repo` in `getReleaseLine` / `getDependencyReleaseLine`, or drop the options object from `config.json` so the config doesn't look like it does something it doesn't.

.changeset/changelog.js
```diff
@@ -0,0 +1,287 @@
...
+          reject(e);
+        } finally {
+          active--;
+          if (queue.length > 0) queue.shift()();
+        }
+      };
+      if (active < MAX_CONCURRENT) run();
+      else queue.push(run);
+    });
+}
```
rafa-thayto, line 76: If someone runs `changeset version` locally without a token this will throw an unhandled error with a stack trace. Might be worth catching at the top level and falling back to plain entries (no links) — that way local runs still work and only CI gets the enriched output.

.changeset/changelog.js
```diff
@@ -0,0 +1,287 @@
...
+
+  if (!res.ok) {
+    throw new Error(`GitHub API responded with ${res.status}: ${await res.text()}`);
+  }
+
+  const json = await res.json();
+  if (json.errors) {
+    throw new Error(`GitHub GraphQL error: ${JSON.stringify(json.errors, null, 2)}`);
+  }
+  if (!json.data) {
```
rafa-thayto: Works fine today since `owner`/`repoName` are hardcoded, but if you ever read `repo` from the options object (see other comment) this becomes a GraphQL injection surface. Safer to use query variables:
```js
const data = await graphql(
  `query($owner: String!, $name: String!) {
    repository(owner: $owner, name: $name) { ... }
  }`,
  { owner, name: repoName }
);
```
Would need a small tweak to `graphql()` to accept a `variables` param — not urgent but worth considering.

.changeset/changelog.js
```diff
@@ -0,0 +1,287 @@
...
+    });
+}
+
+async function graphql(query) {
+  const token = process.env.GITHUB_TOKEN;
+  if (!token) {
+    throw new Error("GITHUB_TOKEN environment variable is required");
+  }
+
+  const controller = new AbortController();
```
rafa-thayto: No retry on transient failures (502, timeout, etc.) — a single blip would fail the whole Version Packages step. A one-shot retry with a short backoff on 5xx/abort errors would make this more resilient in CI without adding much complexity.

.changeset/changelog.js
```diff
@@ -0,0 +1,287 @@
...
+    .replace(/^\s*(?:pr|pull|pull\s+request):\s*#?(\d+)/im, (_, pr) => {
+      let num = Number(pr);
+      if (!isNaN(num)) prFromSummary = num;
+      return "";
+    })
+    .replace(/^\s*commit:\s*([^\s]+)/im, (_, commit) => {
+      commitFromSummary = commit;
+      return "";
+    })
+    .replace(/^\s*(?:author|user):\s*@?([^\s]+)/gim, (_, user) => {
```
rafa-thayto, line 305: nit: `trimRight()` is deprecated — `trimEnd()` is the modern equivalent.

rafa-thayto (commented):
Nice work! The cache + concurrency limiter combo is really well done — in-flight dedup on the promise is a great touch. Left a few suggestions below.

.changeset/changelog.js
```diff
@@ -0,0 +1,287 @@
+const repo = "clerk/cli";
+const [owner, repoName] = repo.split("/");
```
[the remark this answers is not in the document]
wyattjoh: Addressed by dropping the unused `{ repo: ... }` from `.changeset/config.json` in 19d824cb (the alternative the prior reviewer also suggested). The hardcoded value is intentional now that the config no longer advertises configurability.

rafa-thayto (approved):

PR #172, opened 2026-04-15
ci: pin GITHUB_TOKEN to contents:read in all workflows

wyattjoh:
## Summary

- Add top-level `permissions: contents: read` to `ci.yml`, `build-binaries.yml`, `smoke-test.yml`, and `sign-macos.yml`.
- Matches the pin already on `release.yml` and the job-level pin on `enforce-changeset.yml`.
- Closes 8 CodeQL `actions/missing-workflow-permissions` alerts.

## Test plan

- [ ] CI runs green on this PR (workflows still only need read access to repo contents)

PR #171, opened 2026-04-15
fix(pkce): use rejection sampling to remove modulo bias

wyattjoh:
## Summary

- Replace `byte % CHARSET.length` in `generateCodeVerifier` with rejection sampling. The 66-character PKCE charset does not divide evenly into 256, so the naive modulo biased the first 58 characters to appear ~33% more often than the last 8.
- Restores uniform distribution and full RFC 7636 entropy without affecting the public API.
- Flagged by CodeQL: `js/biased-cryptographic-random`.

## Test plan

- [x] `bun test packages/cli-core/src/lib/pkce.test.ts`
- [x] New test asserts per-character counts stay within 10% of uniform over 2000 verifiers
- [x] `bun run typecheck`

PR #173, opened 2026-04-15
ci(smoke-test): pass workflow inputs through env vars

wyattjoh:
## Summary

- Move `inputs.preset`, `inputs.version`, `matrix.docker`, and `matrix.bin` from inline `${{ ... }}` template substitution in `run:` blocks to `env:` bindings, so shell interpolation happens after expansion rather than before.
- Defense-in-depth: the upstream gate in `release.yml` already restricts `issue_comment` triggers to `MEMBER`/`OWNER` actors, but the pattern is cheap to follow and eliminates the rule entirely.
- Closes CodeQL `actions/code-injection/critical` alert on the smoke-test version check.

## Test plan

- [ ] Smoke-test workflow runs green against an existing release (or canary) job

### DIFF
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -28,10 +28,12 @@ jobs:
       matrix: ${{ steps.resolve.outputs.matrix }}
     steps:
       - id: resolve
+        env:
+          PRESET: ${{ inputs.preset }}
         run: |
-          case "${{ inputs.preset }}" in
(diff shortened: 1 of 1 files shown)


wyattjoh:
**Stack: security-codeql-fixes**

- ~~#171~~
- ~~#172~~
- **#173 👈 this PR**

alexcarpenter (approved):

PR #174, opened 2026-04-16
fix(auth): authorize with default scopes

jfoshee:
because the CLI at this time is not in a position to know what scopes to request

When we request an empty (or missing) scope parameter the Authorization Server will use a default set of scopes appropriate for the CLI.

This is important because we want to enumerate permissions being granted in the Consent screen

CleanShot 2026-04-16 at 12 42 29@2x


rafa-thayto:
!snapshot

wyattjoh (approved):

PR #175, opened 2026-04-16
fix(init): simplify auth/keyless flow and fix --starter

rafa-thayto:
## Summary

Four `clerk init` fixes on this branch:

- **Skip `askSkipAuth` and keyless flow when signed in.** If the user has an OAuth token, `clerk init` goes straight to the authenticated flow (auth + link + env pull).
- **Count `CLERK_PLATFORM_API_KEY` as authenticated.** Non-interactive / CI runs with an API key no longer fall through to keyless mode. A new `isAuthenticated()` helper covers both OAuth tokens and the env var.
- **Remove the "Skip authentication for now?" prompt.** The prompt added a step without meaningful choice — the unauthenticated path can't pull keys either way. Now the flow just follows auth state: signed in → authenticated flow + env pull; not signed in → keyless automatically (the keyless info message already points users to `clerk auth login` for later).
- **Keep `clerk init --starter` interactive without `--framework`.** Running `--starter` without `-y` no longer errors with "Non-interactive mode requires --framework" — it only implies "yes, bootstrap," not "skip every prompt."

Also suppresses the "keyless not supported" framework message outside the bootstrap flow.

## Test plan
- [x] `bun run test` — init suite covers the permutations (auth vs. not, API key vs. OAuth, bootstrap vs. existing repo, `--starter` with and without `-y`)
- [x] `bun run lint` / `bun run typecheck`
- [x] CI E2E suite passes (was failing on Astro / Next / TanStack / React Router before the API-key fix)
- [x] Manual: `clerk init` on a keyless framework while signed in → authenticated flow, no keyless prompt
- [x] Manual: `clerk init` on a keyless framework while signed out → keyless flow, no prompt
- [x] Manual: `clerk init` in an existing repo with keyless framework while signed in → no keyless prompt, env pulled
- [x] Manual: `clerk init --starter` (no `-y`) → interactive framework picker instead of an error

### DIFF
--- a/.changeset/more-init-fixes.md
+++ b/.changeset/more-init-fixes.md
@@ -5,5 +5,5 @@
 Fix `clerk init` prompt flow:
 
 - When you are signed in (OAuth or `CLERK_PLATFORM_API_KEY`), `clerk init` skips straight to the authenticated flow — no more "Skip authentication for now?" prompt.
-- When you are not signed in on a keyless-capable framework, `clerk init` now goes keyless automatically (previously prompted) and points you to `clerk auth login` for later.
+- When you are not signed in **during bootstrap** (new projects) on a keyless-capable framework, `clerk init` now goes keyless automatically (previously prompted) and points you to `clerk auth login` for later. Re-runs in an existing project still fall through to the authenticated flow so real keys can be pulled.
 - Keep `clerk init --starter` fully interactive — it no longer fails with "Non-interactive mode requires --framework" when running without `-y`.
--- a/.changeset/more-init-fixes.md
+++ b/.changeset/more-init-fixes.md
@@ -4,5 +4,5 @@
 
 Fix `clerk init` prompt flow:
 
-- Skip the keyless / skip-auth confirmation when you are already signed in.
+- Skip the keyless / skip-auth confirmation when you are already signed in (including via `CLERK_PLATFORM_API_KEY`).
 - Keep `clerk init --starter` fully interactive — it no longer fails with "Non-interactive mode requires --framework" when running without `-y`.
--- a/.changeset/more-init-fixes.md
+++ b/.changeset/more-init-fixes.md
@@ -1,0 +1,8 @@
+---
+"clerk": patch
+---
+
+Fix `clerk init` prompt flow:
+
+- Skip the keyless / skip-auth confirmation when you are already signed in.
+- Keep `clerk init --starter` fully interactive — it no longer fails with "Non-interactive mode requires --framework" when running without `-y`.
--- a/.changeset/more-init-fixes.md
+++ b/.changeset/more-init-fixes.md
@@ -4,5 +4,6 @@
 
 Fix `clerk init` prompt flow:
 
-- Skip the keyless / skip-auth confirmation when you are already signed in (including via `CLERK_PLATFORM_API_KEY`).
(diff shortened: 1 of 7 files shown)


rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

wyattjoh (commented):
## Code Review — PR #175

### Major

**M1. Removing the `bootstrap != null` guard broadens keyless auto-select to existing projects** _(codex partial)_
`packages/cli-core/src/commands/init/index.ts:215-225` now returns `keyless = true` for any unauthenticated user on a keyless-capable framework, not only during bootstrap. Downstream effects:

1. `authenticateAndLink` is skipped when users re-run `clerk init` signed out on an existing Clerk project.
2. `ctx.keyless = true` at `frameworks/helpers.ts:204-221,461-467` toggles permissive `clerkMiddleware()` (no route protection). Idempotency via `hasClerkImport` should prevent overwrites, but a scaffold-idempotency test for `existingClerk: true + keyless: true` would make the invariant load-bearing rather than incidental.
3. `env pull` is skipped for the keyless branch.

The changeset does note keyless auto-select behavior, but the existing-project qualifier is not spelled out. Consider either restoring the `if (!bootstrap) return false` guard, or explicitly documenting the existing-project case in README and changeset.

**M2. `isAuthenticated()` conflates "no credentials" with "network/API failure"** _(codex partial)_
`packages/cli-core/src/commands/init/heuristics.ts:139-147` delegates to `getAuthenticatedEmail()` which catches all errors and returns `null`. Expired tokens, Clerk 5xx, or offline laptops all silently demote the user into keyless (for keyless-capable frameworks) or skip-auth. Previously `askSkipAuth` gave the user a decision point. Consider a lighter credential-presence check (`getToken()` + `CLERK_PLATFORM_API_KEY`) for flow selection, reserving `getAuthenticatedEmail()` for the display label.

### Minor

- **M3.** `isAuthenticated()` is called twice on the authenticated + keyless-capable path (`index.ts:90,94`); resolve once and pass the boolean.
- **M4.** `packages/cli-core/src/commands/init/README.md:48-52` still documents the removed "Continue with temporary keys" prompt. Per `.claude/rules/commands.md`, the README must match behavior.
- **M5.** New test `"--starter without -y runs bootstrap interactively..."` (`index.test.ts:424-436`) fully mocks `promptAndBootstrap` and never exercises the real guard at `bootstrap.ts:173-177` that was the actual bug. Add a direct test for the guard.

- **Second README inconsistency.** `init/README.md:86` says non-keyless frameworks "always force authentication even with `--yes`", but `index.ts:93-99` skips auth for unauthenticated users when `-y` is set.

### Nits

- `isAuthenticated` docstring (`heuristics.ts:139-143`) still references prompts that no longer exist.
- `BootstrapOverrides.implicitBootstrap` doc (`bootstrap.ts:135-140`) doesn't note that `skipConfirm` supersedes it.
- `handleStarter` (`index.ts:153-163`) dropping `skipConfirm: true` changes interactive `--starter` from `previewPlan()` to `previewAndConfirm()`. The changeset mentions `--starter` is "fully interactive", so this looks intentional; worth a bullet for reviewer transparency.

### Positives

`implicitBootstrap` cleanly expresses "user opted into bootstrapping but did not opt out of prompts", better than overloading `skipConfirm`. Test additions cover the authentication x keyless x bootstrap/existing matrix well. `isAuthenticated` correctly short-circuits on `CLERK_PLATFORM_API_KEY` before network calls.

wyattjoh (approved):

PR #176, opened 2026-04-16
feat(prompts): add scroll indicators and interactive listage module

rafa-thayto:
## Summary

- Add custom `select`/`search` prompts in `lib/listage.ts` built on `@inquirer/core`'s `createPrompt` that show `↑ N more above` / `↓ N more below` scroll indicators when choices overflow the visible page
- Consolidate the piped-stdin TTY fallback (previously only in `lib/prompts.ts` for `select`/`confirm`, missing for `search`) and the duplicated `filterChoices` helper into the shared listage module
- Add interactive environment picker to `clerk switch-env` when no argument is given in human mode

### Why custom prompts?

`@inquirer/select` v5 and `@inquirer/search` v4 removed the old `helpMode` theme option, and `usePagination` returns a plain string with no metadata about items above/below the viewport. There's no theme hook or callback to inject scroll indicators without reimplementing the render function via `createPrompt`.

### Commands updated

| Command | Prompt type | What it picks |
|---|---|---|
| `clerk link` | `search` | Clerk application |
| `clerk init` | `search` | Framework, package manager |
| `clerk api` (interactive) | `select` | API category, endpoint |
| `clerk deploy` | `select` | Domain type, OAuth flow |
| `clerk switch-env` | `select` (new) | Environment (when no arg given) |

### Other improvements
- `search` prompts now have the TTY fallback for piped stdin (previously missing — `link` and `init/bootstrap` imported `search` directly from `@inquirer/prompts`)
- Removed duplicated `filterChoices` helper from `init/bootstrap.ts` (now shared from `listage.ts`)

## Test plan
- [ ] Run `clerk api` with no args — verify scroll indicators appear on the endpoint list when it exceeds page size
- [ ] Run `clerk link` — verify app picker shows scroll indicators with many apps
- [ ] Run `clerk switch-env` with no argument — verify interactive picker appears
- [ ] Run `clerk switch-env staging` — verify direct argument still works
- [ ] Pipe stdin (`echo | clerk switch-env`) — verify fallback to text output
- [ ] Verify all CI checks pass (typecheck, lint, tests)

### DIFF
--- a/packages/cli-core/src/commands/switch-env/index.ts
+++ b/packages/cli-core/src/commands/switch-env/index.ts
@@ -36,6 +36,10 @@ export async function switchEnv(environmentArg: string | undefined): Promise<voi
         })),
         default: current,
       });
+    } else if (available.length <= 1) {
+      log.info(`Current environment: ${current}`);
+      log.info("Only one environment configured — nothing to switch to.");
+      return;
     } else {
       log.info(`Current environment: ${current}`);
       log.info(`Available environments: ${available.join(", ")}`);
--- a/packages/cli-core/src/commands/switch-env/index.ts
+++ b/packages/cli-core/src/commands/switch-env/index.ts
@@ -20,14 +20,15 @@ import { log } from "../../lib/log.ts";
 import { isHuman } from "../../mode.ts";
 import { select } from "../../lib/listage.ts";
 
-export async function switchEnv(environment: string | undefined): Promise<void> {
+export async function switchEnv(environmentArg: string | undefined): Promise<void> {
   const available = getAvailableEnvs();
   const current = getCurrentEnvName();
 
   // No argument: show interactive picker (human) or print info (non-interactive)
-  if (!environment) {
+  let target = environmentArg;
+  if (!target) {
     if (isHuman() && available.length > 1) {
-      environment = await select<string>({
+      target = await select<string>({
         message: "Switch to environment:",
         choices: available.map((env) => ({
           name: env === current ? `${env} (current)` : env,
@@ -42,28 +43,28 @@ export async function switchEnv(environment: string | undefined): Promise<void> 
     }
   }
 
-  if (!isValidEnv(environment)) {
+  if (!isValidEnv(target)) {
     throw new CliError(
-      `Unknown environment "${environment}". Available environments: ${available.join(", ")}`,
+      `Unknown environment "${target}". Available environments: ${available.join(", ")}`,
     );
   }
 
   const previousEnv = getCurrentEnvName();
 
-  if (previousEnv === environment) {
-    log.data(`Already on ${environment} environment.`);
+  if (previousEnv === target) {
+    log.data(`Already on ${target} environment.`);
     return;
   }
 
   // Update the in-memory state and persist
-  setCurrentEnv(environment);
-  await setEnvironment(environment);
+  setCurrentEnv(target);
+  await setEnvironment(target);
 
-  log.data(`Switched from ${previousEnv} to ${environment}.`);
+  log.data(`Switched from ${previousEnv} to ${target}.`);
 
   // Check if there's a stored token for the target environment
   const token = await getToken();
   if (!token) {
-    log.data(`No credentials found for ${environment}. Run \`clerk auth login\` to authenticate.`);
+    log.data(`No credentials found for ${target}. Run \`clerk auth login\` to authenticate.`);
   }
 }
--- a/packages/cli-core/src/commands/switch-env/index.ts
+++ b/packages/cli-core/src/commands/switch-env/index.ts
@@ -17,16 +17,29 @@ import {
 } from "../../lib/environment.ts";
 import { CliError } from "../../lib/errors.ts";
 import { log } from "../../lib/log.ts";
+import { isHuman } from "../../mode.ts";
+import { select } from "../../lib/listage.ts";
 
 export async function switchEnv(environment: string | undefined): Promise<void> {
   const available = getAvailableEnvs();
+  const current = getCurrentEnvName();
 
-  // No argument: print current environment
+  // No argument: show interactive picker (human) or print info (non-interactive)
   if (!environment) {
-    const current = getCurrentEnvName();
-    log.info(`Current environment: ${current}`);
-    log.info(`Available environments: ${available.join(", ")}`);
-    return;
+    if (isHuman() && available.length > 1) {
+      environment = await select<string>({
+        message: "Switch to environment:",
+        choices: available.map((env) => ({
+          name: env === current ? `${env} (current)` : env,
+          value: env,
+        })),
+        default: current,
+      });
+    } else {
+      log.info(`Current environment: ${current}`);
+      log.info(`Available environments: ${available.join(", ")}`);
+      return;
+    }
   }
 
   if (!isValidEnv(environment)) {
--- a/packages/cli-core/src/commands/switch-env/index.ts
+++ b/packages/cli-core/src/commands/switch-env/index.ts
@@ -27,7 +27,7 @@ export async function switchEnv(environmentArg: string | undefined): Promise<voi
   // No argument: show interactive picker (human) or print info (non-interactive)
   let target = environmentArg;
   if (!target) {
-    if (isHuman() && available.length > 1) {
+    if (isHuman() && available.length > 1 && process.stdin.isTTY) {
       target = await select<string>({
         message: "Switch to environment:",
         choices: available.map((env) => ({
@@ -36,6 +36,10 @@ export async function switchEnv(environmentArg: string | undefined): Promise<voi
         })),
         default: current,
       });
+    } else if (isHuman() && available.length > 1 && !process.stdin.isTTY) {
+      throw new CliError(
+        "No interactive terminal available — pass an environment name explicitly: `clerk switch-env <name>`",
+      );
     } else if (available.length <= 1) {
       log.info(`Current environment: ${current}`);
       log.info("Only one environment configured — nothing to switch to.");
--- a/packages/cli-core/src/lib/listage.test.ts
+++ b/packages/cli-core/src/lib/listage.test.ts
@@ -1,5 +1,5 @@
-import { test, expect, describe } from "bun:test";
-import { scrollBounds, withScrollIndicators, filterChoices } from "./listage.ts";
+import { test, expect, describe, beforeEach } from "bun:test";
+import { scrollBounds, withScrollIndicators, filterChoices, ttyContext } from "./listage.ts";
 
 describe("scrollBounds", () => {
   test("returns zeros when all items fit on page", () => {
@@ -26,10 +26,18 @@ describe("scrollBounds", () => {
     expect(scrollBounds(20, 19, 5)).toEqual({ above: 15, below: 0 });
   });
 
-  test("above + below + pageSize = totalItems", () => {
+  test("above + below + pageSize = totalItems (pageSize=5)", () => {
     for (let active = 0; active < 20; active++) {
       const { above, below } = scrollBounds(20, active, 5);
       expect(above + below + 5).toBe(20);
+    }
+  });
+
+  test("above + below + pageSize = totalItems (pageSize=7, odd)", () => {
+    // Odd pageSize may drift by ±1 at boundaries but must never be catastrophically wrong
+    for (let active = 0; active < 20; active++) {
+      const { above, below } = scrollBounds(20, active, 7);
+      expect(above + below + 7).toBe(20);
     }
   });
 });
@@ -94,5 +102,30 @@ describe("filterChoices", () => {
 
   test("returns empty array when nothing matches", () => {
     expect(filterChoices(choices, "angular")).toEqual([]);
+  });
+});
+
+describe("ttyContext", () => {
+  const originalIsTTY = process.stdin.isTTY;
+
+  beforeEach(() => {
+    process.stdin.isTTY = originalIsTTY;
+  });
+
+  test("returns undefined when stdin is a TTY", () => {
+    process.stdin.isTTY = true;
+    expect(ttyContext()).toBeUndefined();
+  });
+
+  test("returns context with input and close when stdin is not a TTY", () => {
+    process.stdin.isTTY = false;
+    const ctx = ttyContext();
+    // On macOS/Linux with /dev/tty available, this should return a context
+    if (ctx) {
+      expect(ctx.input).toBeDefined();
+      expect(typeof ctx.close).toBe("function");
+      ctx.close();
+    }
+    // On CI/Docker without a TTY, ttyContext may return undefined — both are valid
   });
 });
--- a/packages/cli-core/src/lib/listage.test.ts
+++ b/packages/cli-core/src/lib/listage.test.ts
@@ -1,0 +1,98 @@
+import { test, expect, describe } from "bun:test";
+import { scrollBounds, withScrollIndicators, filterChoices } from "./listage.ts";
+
+describe("scrollBounds", () => {
+  test("returns zeros when all items fit on page", () => {
+    expect(scrollBounds(5, 0, 7)).toEqual({ above: 0, below: 0 });
+    expect(scrollBounds(7, 3, 7)).toEqual({ above: 0, below: 0 });
+  });
+
+  test("at the top of a long list", () => {
+    // 20 items, active=0, pageSize=5 → first 5 visible
+    expect(scrollBounds(20, 0, 5)).toEqual({ above: 0, below: 15 });
+    expect(scrollBounds(20, 1, 5)).toEqual({ above: 0, below: 15 });
+  });
+
+  test("in the middle of a long list", () => {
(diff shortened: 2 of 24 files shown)


wyattjoh (commented):
## Code Review — PR #176

### High

**H1. TTY fallback does not handle `/dev/tty` or `CONIN$` open failures** _(codex confirmed)_
`packages/cli-core/src/lib/listage.ts:42-45` unconditionally calls `createReadStream(TTY_PATH)` when stdin is not a TTY, without attaching an error listener. In Docker without `--tty`, detached CI runners, or Windows sessions without `CONIN$`, the stream emits an async `error` event that is unhandled. Previously the same flaw existed in `lib/prompts.ts` but is now extended to `search` callsites (`link`, `init/bootstrap`).

**H2. Scroll indicator math ignores rendered-line wrapping** _(codex confirmed)_
`scrollBounds()` (`listage.ts:68-95`) counts items, but `@inquirer/core`'s `usePagination` paginates on rendered lines post-`breakLines()`. Long `clerk api` labels or narrow terminals produce wrapped choices, which means the "more above/below" counts and visible-range inference diverge from what is actually shown.

### Medium

- **M1.** Prompt height jumps 1 line at edges (`listage.ts:104-107,115-117`): only the non-empty indicator row renders at top/bottom.
- **M2.** `Math.max(pageSize - 2, 3)` plus indicator rows (`listage.ts:294,508`): callers passing `pageSize: 3` or `4` get 4-5 lines.
- **M3.** Zero direct tests for `lib/listage.ts`. Every command mocks it out; `scrollBounds`, `withScrollIndicators`, `normalizeChoices`, `isSelectable`, the TTY fallback, keypress handlers, and search race-cancellation are all unverified. Add at minimum a `lib/listage.test.ts` for the pure helpers.
- **M4.** `ValidationError` thrown inside `useMemo` (`listage.ts:216-223`) is not guaranteed to be caught by `@inquirer/core`'s validation path; prefer validating up-front in the `select()` wrapper.
- **M5.** `search` `active` can be `-1` with empty results (`listage.ts:417-423,461`): the `useState<number>()` destructure default only applies when the value is strictly `undefined`, not for the `bounds.first = -1` case.
- **M6.** `cursorHide` is appended on every select render (`listage.ts:340`) with no matching `cursorShow` on the done branch (`listage.ts:319-321`). Commands that chain prompts (`deploy`, `api/interactive`) can render subsequent prompts with a hidden cursor.
- **M7.** Duplicate `select` API surface: old `lib/prompts.ts:36-46` still exports `select`, and `commands/init/skills.ts:17,106-113` still uses it. Two implementations with different appearance and TTY-fallback paths.
- **M8.** `packages/cli-core/src/commands/switch-env/README.md` is not updated despite the new interactive picker behavior (per `.claude/rules/commands.md`).

- **All-disabled list deadlock.** `listage.ts:217-220` checks `isNavigable` rather than `isSelectable`; an all-disabled list never throws and produces a prompt with no selectable item and no exit path.
- **Stuck loading state on search error.** `listage.ts:427-453` sets `status` to `"loading"` before fetch but the error path never resets it to `"idle"`, leaving the prompt loading indefinitely and blocking arrow-key nav (`listage.ts:485-494`).

### Nits

- `isSelectable<T>` narrows to `{ disabled?: false }` but runtime accepts any falsy `disabled` (`listage.ts:124-126`).
- `rl.clearLine(0)` on arrow keys clobbers the type-ahead buffer (`listage.ts:251-260`).
- Type-ahead `startsWith` lacks Unicode normalization (`listage.ts:275-278`).
- `SelectConfig` / `SearchConfig` types not exported (`listage.ts:185-191,378-390`).
- `•` bullet in `keysHelpTip` may not render in all terminals (`listage.ts:201,401`).
- `switch-env/index.ts:30` reassigns the function parameter.
- `@inquirer/ansi` and `@inquirer/figures` added as direct deps though already transitive.

### Positives

Clear motivation in the PR body (upstream removed `helpMode` / `usePagination` metadata). Abort-controller wiring in the search `useEffect` correctly cancels stale results. The `defaultApplied.current` ref is a subtle bug fix that is easy to miss. Deduplication of `filterChoices` from `init/bootstrap.ts` into the shared module is a net cleanup. Changeset scoped as `minor` matches the surface-visible change.

wyattjoh (commented):
## Code review

Net positive change: a small custom select/search on top of @inquirer/core that adds scroll indicators, fills in the piped-stdin TTY fallback that search was missing, and ships an interactive env picker for switch-env. Implementation is careful and the scrollBounds test coverage is good. A few things worth tightening before merge.

packages/cli-core/src/lib/listage.ts
```diff
@@ -0,0 +1,583 @@
...
+
+  const middle = Math.floor(pageSize / 2);
+  const spaceBelow = totalItems - active;
+
+  let firstVisible: number;
+  if (spaceBelow < pageSize - middle) {
+    // Near the bottom — window slides to show the last pageSize items.
+    firstVisible = totalItems - pageSize;
+  } else if (active <= middle) {
+    // Near the top — window starts at 0.
```
wyattjoh, line 102: **[HIGH] scrollBounds may drift from usePagination for odd pageSize**

The inline comment says this mirrors usePagination's usePointerPosition, but @inquirer/core v11's pagination only slides the window when the active item crosses out of the visible range, not when it hits the `middle` boundary. For pageSize=7 (middle=3), scrollBounds(20, 4, 7) reports {above: 1, below: 12}, but the actual rendered window may still be items 0..6, giving true above=0. All the unit tests use pageSize=5 (middle=2) and happen to align; a test that walks active from 0..19 with pageSize=7 would surface the drift. Either mirror usePagination's exact algorithm or acknowledge imprecision in the indicator text.

const middle = Math.floor(pageSize / 2);
...
if (spaceBelow < pageSize - middle) {
  firstVisible = totalItems - pageSize;
} else if (active <= middle) {
  firstVisible = 0;
} else {
  firstVisible = active - middle;
}

</details>

packages/cli-core/src/lib/listage.ts
```diff
@@ -0,0 +1,583 @@
...
+// ---------------------------------------------------------------------------
+
+const TTY_PATH = process.platform === "win32" ? "CONIN$" : "/dev/tty";
+
+function ttyContext(): { input: NodeJS.ReadableStream; close: () => void } | undefined {
+  if (process.stdin.isTTY) return undefined;
+  try {
+    const input = createReadStream(TTY_PATH);
+    // Swallow open errors (Docker without --tty, detached CI runners, Windows
+    // sessions without CONIN$) so the prompt falls back to the default stdin
```
wyattjoh, line 47: **[MEDIUM] ttyContext duplicated between listage.ts and prompts.ts**

lib/prompts.ts still carries its own piped-stdin fallback for confirm, and lib/listage.ts reimplements a nicer version (with an error-swallowing guard for detached CI). Either export ttyContext from one module and share it, or fold confirm into listage.ts. Two copies will drift over time — the listage copy already handles a case the prompts copy doesn't.

// listage.ts
function ttyContext(): ... {
  if (process.stdin.isTTY) return undefined;
  try {
    const input = createReadStream(TTY_PATH);
    input.on("error", () => {});
    return { input, close: () => input.close() };
  } catch { return undefined; }
}

</details>

packages/cli-core/src/test/lib/stubs.ts
```diff
@@ -80,6 +80,25 @@ export const promptsStubs = {
...
+    return choices.filter((c: T) => c.name.toLowerCase().includes(lower));
+  },
+  Separator: class Separator {
+    separator: string;
+    constructor(separator = "──────") {
+      this.separator = separator;
+    }
+    static isSeparator(item: unknown): item is Separator {
+      return item instanceof Separator;
+    }
```
wyattjoh: **[MEDIUM] listageStubs reimplements filterChoices and Separator instead of importing them**

The production filterChoices is a 4-line pure function in lib/listage.ts with no runtime dependency on @inquirer/core — it can be imported directly into stubs. Same for the Separator re-export. Right now the same logic is copy-pasted into both test/lib/stubs.ts and test/integration/lib/harness.ts. If filterChoices ever changes (e.g. diacritic-insensitive matching) the stubs will silently diverge and tests will pass against a ghost implementation.

filterChoices: <T extends { name: string }>(choices: T[], term: string | undefined): T[] => {
  if (!term) return choices;
  const lower = term.toLowerCase();
  return choices.filter((c: T) => c.name.toLowerCase().includes(lower));
},

</details>

packages/cli-core/src/commands/switch-env/index.ts
```diff
@@ -17,40 +17,54 @@ import {
...
-  // No argument: print current environment
-  if (!environment) {
-    const current = getCurrentEnvName();
-    log.info(`Current environment: ${current}`);
-    log.info(`Available environments: ${available.join(", ")}`);
-    return;
+  // No argument: show interactive picker (human) or print info (non-interactive)
+  let target = environmentArg;
+  if (!target) {
+    if (isHuman() && available.length > 1) {
```
wyattjoh: **[MEDIUM] switch-env silently prints status when only one env is configured**

When `available.length === 1` and no argument is given in human mode, the guard `available.length > 1` falls through to the non-interactive branch and prints `Current environment: X / Available: X`. A user who typed `clerk switch-env` expecting a picker will see this as a confusing non-op. Consider printing a short note like `Only one environment configured — nothing to switch to.` or document the one-env behavior in the README.

if (isHuman() && available.length > 1) {
  target = await select<string>({ ... });
} else {
  log.info(`Current environment: ${current}`);
  log.info(`Available environments: ${available.join(", ")}`);
  return;
}

</details>

packages/cli-core/src/lib/listage.ts
```diff
@@ -0,0 +1,583 @@
...
+  active: number,
+  effectivePageSize: number,
+): string {
+  const { above, below } = scrollBounds(totalItems, active, effectivePageSize);
+  const top = above > 0 ? styleText("dim", ` ↑ ${above} more above`) : " ";
+  const bottom = below > 0 ? styleText("dim", ` ↓ ${below} more below`) : " ";
+  return [top, page, bottom].join("\n");
+}
+
+// ---------------------------------------------------------------------------
```
wyattjoh, line 135: **[LOW] Scroll-indicator arrows have no ASCII fallback**

`↑` / `↓` render as `?` on Windows cmd.exe with legacy code pages. `@inquirer/figures` (already imported for the cursor) exposes `figures.arrowUp` / `figures.arrowDown` with ASCII downgrades. Using those would match the cursor icon's platform behavior.

const top = above > 0 ? styleText("dim", ` ↑ ${above} more above`) : " ";
const bottom = below > 0 ? styleText("dim", ` ↓ ${below} more below`) : " ";

</details>

Railly (commented):
Arrow navigation lands on disabled items ([lib/listage.ts:265](https://github.com/clerk/cli/blob/c5743eae694d587f3042e42530f985b5572f649b/packages/cli-core/src/lib/listage.ts#L265)). Uses `isNavigable` instead of `isSelectable`, upstream `@inquirer/select` uses `isSelectable` in the equivalent loop, and the search prompt below ([:502](https://github.com/clerk/cli/blob/c5743eae694d587f3042e42530f985b5572f649b/packages/cli-core/src/lib/listage.ts#L502)) uses it too, so select is the odd one out. Coupled render quirk at [:317-318](https://github.com/clerk/cli/blob/c5743eae694d587f3042e42530f985b5572f649b/packages/cli-core/src/lib/listage.ts#L317-L318) (cursor `❯` renders on active disabled rows) fixes itself once the loop predicate is corrected.

Railly (approved):

packages/cli-core/src/commands/switch-env/index.ts
```diff
@@ -17,40 +17,58 @@ import {
...
+      target = await select<string>({
```
wyattjoh, line 31: When `isHuman()` is true and `available.length > 1`, this branch unconditionally calls `select(...)`. Inside `listage.select`, `ttyContext()` returns `undefined` if `createReadStream("/dev/tty")` throws (Docker without `--tty`, detached CI runners, some SSH contexts). In that case `rawSelect` runs with no `input` override and tries to read from the real `process.stdin` — which may already be at EOF if stdin was piped. The prompt will then either hang or surface a cryptic unhandled error event rather than a usable message.

Consider detecting this case before invoking the picker and falling back to the existing "print current + available environments" branch, or have `listage.select` throw a `CliError` with actionable text ("No interactive terminal available — pass an environment name explicitly.") when it would otherwise block on EOF. Either path keeps `clerk switch-env` (no arg) well-behaved in Docker/CI.

packages/cli-core/src/lib/listage.ts
```diff
@@ -0,0 +1,591 @@
...
+      let selectableIndex = -1;
+      const position = items.findIndex((item) => {
+        if (Separator.isSeparator(item)) return false;
+        selectableIndex++;
+        return selectableIndex === selectedIndex;
+      });
+      const item = items[position];
+      if (item != null && isSelectable(item)) setActive(position);
+      searchTimeoutRef.current = setTimeout(() => rl.clearLine(0), 700);
+    } else if (isBackspaceKey(key)) {
```
wyattjoh, line 280: `separatorCount` is a `let` inside the `createPrompt` body and is incremented inside `renderItem`. `usePagination` only invokes `renderItem` for the visible slice of `items`. When a `Separator` sits above the first visible item, it is never rendered, so `separatorCount` stays at 0 and the `indexLabel` computation `${index + 1 - separatorCount}` reports the wrong human-facing number for items below it.

This is latent today because `indexMode` defaults to `"hidden"` and no caller opts into `"number"`, but it will surprise the first consumer that does. Either pre-compute per-item display indices in one pass over `items` before `usePagination` runs and look them up inside `renderItem`, or — since `indexMode: "number"` is not reachable from the current `SelectConfig` surface anyway — delete the branch until a real caller asks for it.

packages/cli-core/src/lib/prompts.ts
```diff
@@ -6,42 +6,22 @@
...
- * Uses /dev/tty on Unix and CONIN$ on Windows.
+ * Uses the shared ttyContext from lib/listage.ts for consistent error handling.
  */
 
-import { createReadStream } from "node:fs";
-import { confirm as inquirerConfirm, select as inquirerSelect } from "@inquirer/prompts";
-
-/** OS-specific path to the controlling terminal's input stream. */
-const TTY_PATH = process.platform === "win32" ? "CONIN$" : "/dev/tty";
+import { confirm as inquirerConfirm } from "@inquirer/prompts";
```
wyattjoh, line 12: This PR consolidates the piped-stdin TTY fallback into `ttyContext()` and upgrades `search` to use it. But `confirm` continues to be imported directly from `@inquirer/prompts` in `link/index.ts`, `deploy/index.ts`, `init/bootstrap.ts`, `init/preview.ts`, `unlink/index.ts`, and `api/interactive.ts`. In those commands, a piped-stdin invocation will successfully prompt for `select`/`search`/`input` via the wrapper but fail on the next `confirm` call because it reads raw stdin.

Converting those sites to `import { confirm } from "../../lib/prompts.ts"` would close the inconsistency. An oxlint `no-restricted-imports` rule pointing `confirm`/`select`/`search` from `@inquirer/prompts` at the wrapper modules would prevent regressions. Happy to do this in a follow-up if you'd rather keep this PR focused.

packages/cli-core/src/lib/listage.test.ts
```diff
@@ -0,0 +1,98 @@
+import { test, expect, describe } from "bun:test";
```
wyattjoh: `listage.test.ts` covers the pure functions (`scrollBounds`, `withScrollIndicators`, `filterChoices`). The newly added `input.on("error", () => {})` handler in `ttyContext()` exists specifically to prevent a crash when `/dev/tty` open errors out on Docker/Windows, but nothing exercises it. A regression that removed the error handler would not be caught.

Mirroring the two TTY tests from `prompts.test.ts` against `listage.ttyContext` would lock the behaviour in. The test additions to `prompts.test.ts` in this PR already had to add `.on()` to the mock stream to keep the existing tests green, which shows how easy it is to break.

packages/cli-core/src/lib/listage.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  test("in the middle of a long list", () => {
+    // 20 items, active=10, pageSize=5, middle=2 → firstVisible=8
+    const result = scrollBounds(20, 10, 5);
+    expect(result.above).toBe(8);
+    expect(result.below).toBe(7);
+    expect(result.above + result.below + 5).toBe(20);
+  });
+
+  test("near the bottom of a long list", () => {
+    // 20 items, active=19, pageSize=5 → last 5 visible
```
wyattjoh, line 25: The docstring for `scrollBounds` notes a documented ±1 imprecision at odd `pageSize` boundaries and calls it cosmetic. The test `above + below + pageSize = totalItems` only exercises `pageSize=5`. Adding the same assertion at `pageSize=7` would encode the "never catastrophically wrong" invariant so a future refactor can't accidentally drift past the acknowledged ±1.

Railly (approved):

PR #179, opened 2026-04-17
fix(update): PATH-priority installer detection, asdf shims, --all

rafa-thayto:
Fixes `clerk update` silently writing to the wrong installer on machines with multiple `clerk` binaries on PATH. Replaces runtime-based installer detection with PATH-priority-aware resolution, adds asdf shim support, and introduces an opt-in `--all` batch mode.

## Before / after

Reporter's setup (bun + asdf-managed npm + Homebrew):

```
$ which -a clerk
/Users/rafa/.bun/bin/clerk
/Users/rafa/.asdf/shims/clerk
/opt/homebrew/bin/clerk
```

**Before:** `clerk update --channel canary` reported success, but `clerk -v` still showed the old version. The update had landed in `~/.asdf/shims/clerk` (via an npm fallback), not in `~/.bun/bin/clerk` (the one the shell actually resolves).

**After:** the update writes to `~/.bun/install/global/node_modules/` via `bun add -g`, and `clerk -v` reflects the new version. A post-run report lists the asdf and Homebrew installs with their owners; `--all` updates the asdf install too (via `asdf which` → underlying npm, plus `asdf reshim nodejs` after).

## What the new UX looks like

Default (first-on-PATH only, other installs reported):

```
┌  clerk update
◇  Checking for updates
│    Current: 0.8.4
│    Latest:  0.8.5 (canary)
│    Target:  /Users/rafa/.bun/bin/clerk (bun)
│
◇  Updated bun: /Users/rafa/.bun/bin/clerk
│
│  Also found 2 other clerk installs:
│    /Users/rafa/.asdf/shims/clerk (npm)
│    /opt/homebrew/bin/clerk (homebrew) — Homebrew has no canary tap
│  Run `clerk update --all` to update them too.
│
│  If `clerk` still points to the old binary, run `hash -r` or open a new shell.
└  Successfully updated to 0.8.5
```

`--all` (batch + per-install summary):

```
◇  Updated bun: /Users/rafa/.bun/bin/clerk
◇  Updated npm: /Users/rafa/.asdf/shims/clerk
│
│  Summary:
│    ✓ /Users/rafa/.bun/bin/clerk (bun)
│    ✓ /Users/rafa/.asdf/shims/clerk (npm)
│    ⚠ /opt/homebrew/bin/clerk (homebrew) — skipped: no canary tap
```

`install.sh` standalone (refuse, don't guess):

```
│    Target:  /usr/local/bin/clerk (unknown)
│
│  ⚠ Cannot auto-update: unknown installer — not a package-manager-owned binary
│    Reinstall via your preferred method, e.g.:
│      bun add -g clerk@0.8.5
│      npm install -g clerk@0.8.5
│      curl -fsSL …/install.sh | bash
└  Update required manual action
```

## Fixed

- **PATH-priority installer detection.** Resolves the *first* `clerk` on PATH (what the shell actually executes) and runs the installer that owns it. If no known installer owns the primary target, refuses with reinstall guidance instead of silently writing to a different prefix.
- **bun detection.** `detectInstaller()` Stage 2b compared `process.execPath` against `bun pm bin -g` (the shim dir), which never matched the Bun-compiled platform binary under `~/.bun/install/global/node_modules`. Replaced with path-based `ownerOfBinary()` that matches the install dir.
- **asdf shims.** asdf shims are bash scripts (not symlinks), so `realpath` returns the shim itself and nothing matches. Now resolved via `asdf which <name>`; the underlying binary is owner-matched normally, and `asdf reshim <plugin>` runs after install as a safety net for asdf versions that don't auto-reshim.

## Added

- **`--all` flag.** Updates every `clerk` install on PATH in one run. Skips Homebrew on non-stable channels and `null`-owned binaries with per-install warnings; summary printed at the end.
- **Other-installs report.** After every run, lists other `clerk` installs with their owners so shadowing is visible.
- **Shell-hash hint.** Prints `hash -r` (bash/zsh/sh/dash/ksh), `rehash` (tcsh/csh), or nothing (fish auto-rehashes, PowerShell has no cache) based on `$SHELL`.

## Version managers

- **nvm** — works out of the box. Real symlinks; `realpath` chases into `<nvm-version>/lib/node_modules`, which matches active `npm prefix -g`.
- **asdf** — handled via `resolveAsdfShim` + post-install `asdf reshim`. Honors `$ASDF_DATA_DIR`. Falls back to "unknown" (refuse) when asdf isn't installed.

## New exports in `packages/cli-core/src/lib/installer.ts`

| Export | Contract |
| --- | --- |
| `findClerkOnPath(name?)` | Walks `process.env.PATH`, realpaths + dedupes, filters to executable regular files (POSIX X bit / Windows PATHEXT). Shell-agnostic. |
| `getInstallerPackageDirs()` | Returns the packages dir for each PM present (npm, pnpm, yarn, bun). |
| `ownerOfBinary(path, installDirs)` | Pure, synchronous. Homebrew pattern → longest-match PM prefix → `null`. |
| `isAsdfShimPath(path)` | Path-based, doesn't require asdf installed. |
| `resolveAsdfShim(path)` | `asdf which`-based resolution. Returns input unchanged when asdf is absent or can't resolve. |
| `asdfPluginFromPath(path)` | Extracts plugin name from `<asdf>/installs/<plugin>/<ver>/...`. |
| `asdfReshim(plugin)` | Best-effort; swallows errors. |

The existing `detectInstaller()` is preserved for backward compatibility but no longer used by the update command.

## Test plan

- [x] `bun run typecheck` passes
- [x] `bun run lint` passes (oxlint, 0 warnings / 0 errors)
- [x] `bun run test` passes (71/71 files); `installer.test.ts` has 53 cases including 16 new ones covering `ownerOfBinary`, `findClerkOnPath`, `isAsdfShimPath`, `asdfPluginFromPath`, and `resolveAsdfShim`
- [ ] Manual verification on reporter's setup (bun + asdf-npm + Homebrew): `clerk update --channel canary` writes to `~/.bun/install/global/node_modules` and `clerk -v` reflects the new version
- [ ] `--all` updates both bun and asdf-npm (via `asdf which`) and runs `asdf reshim nodejs` after
- [ ] `--all` skips Homebrew with a channel-mismatch warning when `--channel canary` is used
- [ ] `install.sh`-installed clerk (standalone binary) is refused with reinstall guidance
- [ ] `hash -r` hint renders on zsh/bash; `rehash` on tcsh; no hint on fish/pwsh

### DIFF
--- a/packages/cli-core/src/commands/update/index.ts
+++ b/packages/cli-core/src/commands/update/index.ts
@@ -1,4 +1,4 @@
-import { isHuman } from "../../mode.ts";
+import { isAgent, isHuman } from "../../mode.ts";
 import { green, cyan, yellow, dim } from "../../lib/color.ts";
 import { CliError } from "../../lib/errors.ts";
 import {
@@ -144,6 +144,9 @@ function reportOtherInstalls(others: Target[], channel: string): void {
 
 /** Hint for invalidating the current shell's command-hash cache after update. */
 function hashHint(): string | null {
+  // Windows shells (cmd.exe, PowerShell) don't cache command paths, and `$SHELL`
+  // is typically unset — don't emit a POSIX hint in that case.
+  if (process.platform === "win32") return null;
   const shell = (process.env.SHELL ?? "").toLowerCase();
   if (shell.endsWith("/fish") || shell.endsWith("fish.exe")) return null; // auto-rehashes
   if (shell.endsWith("/pwsh") || shell.endsWith("powershell.exe")) return null; // no cache
@@ -152,6 +155,19 @@ function hashHint(): string | null {
   }
   // bash, zsh, sh, dash, ksh all support `hash -r`.
   return "If `clerk` still points to the old binary, run `hash -r` or open a new shell.";
+}
+
+/**
+ * Detect invocation via a package runner (npx/bunx/pnpm dlx). The binary lives
+ * in a runner cache (e.g. `~/.npm/_npx/<hash>/...`) not on PATH, so `ownerOfBinary`
+ * can't identify it — but we can recognize the runner from env vars it sets.
+ */
+function detectPackageRunner(): "npx" | "bunx" | null {
+  const ua = (process.env.npm_config_user_agent ?? "").toLowerCase();
+  const execPath = (process.env.npm_execpath ?? "").toLowerCase();
+  if (execPath.includes("npx") || ua.startsWith("npm/")) return "npx";
+  if (ua.startsWith("bun/")) return "bunx";
+  return null;
 }
 
 // ── Confirmation ─────────────────────────────────────────────────────────────
@@ -202,25 +218,40 @@ export async function update(options: UpdateOptions): Promise<void> {
   // Primary target can't be updated by us: refuse (don't guess a different installer).
   const primarySkip = whyCantUpdate(primary, channel);
   if (primarySkip) {
-    log.warn(`Cannot auto-update: ${primarySkip}`);
-    if (primary.owner === "homebrew") {
-      log.info(`  Run: ${cyan("brew upgrade clerk")}`);
-    } else if (primary.owner === null) {
-      log.info(`  This binary appears to be installed outside any known package manager.`);
-      log.info(`  Reinstall via your preferred method, e.g.:`);
+    const runner = detectPackageRunner();
+    if (primary.owner === null && runner) {
+      // npx/bunx runs land in a runner cache that isn't on PATH. There's
+      // nothing to "update" in place — the user needs a global install.
+      log.warn(`Running via ${runner}; no installed clerk to update. Install globally first:`);
       log.info(`    ${cyan(`bun add -g ${UPDATE_PACKAGE_NAME}@${latest}`)}`);
       log.info(`    ${cyan(`npm install -g ${UPDATE_PACKAGE_NAME}@${latest}`)}`);
-      log.info(
-        `    ${cyan(`curl -fsSL https://raw.githubusercontent.com/clerk/cli/main/install.sh | bash`)}`,
-      );
+    } else {
+      log.warn(`Cannot auto-update: ${primarySkip}`);
+      if (primary.owner === "homebrew") {
+        log.info(`  Run: ${cyan("brew upgrade clerk")}`);
+      } else if (primary.owner === null) {
+        log.info(`  This binary appears to be installed outside any known package manager.`);
+        log.info(`  Reinstall via your preferred method, e.g.:`);
+        log.info(`    ${cyan(`bun add -g ${UPDATE_PACKAGE_NAME}@${latest}`)}`);
+        log.info(`    ${cyan(`npm install -g ${UPDATE_PACKAGE_NAME}@${latest}`)}`);
+        log.info(
+          `    ${cyan(`curl -fsSL https://raw.githubusercontent.com/clerk/cli/main/install.sh | bash`)}`,
+        );
+      }
     }
     reportOtherInstalls(others, channel);
     if (isHuman()) outro("Update required manual action");
     return;
   }
 
-  const autoConfirm = options.yes || !isHuman();
-  const shouldInstall = autoConfirm || (await confirmUpdate(currentVersion, latest));
+  // In agent/non-interactive mode, require explicit `--yes` rather than silently
+  // running a global install the caller didn't confirm.
+  if (isAgent() && !options.yes) {
+    log.info(`Run \`clerk update --yes\` to proceed.`);
+    return;
+  }
+
+  const shouldInstall = options.yes || (await confirmUpdate(currentVersion, latest));
 
   if (!shouldInstall) {
     if (isHuman()) outro("Update cancelled");
@@ -267,7 +298,13 @@ export async function update(options: UpdateOptions): Promise<void> {
     await asdfReshim(plugin);
   }
 
-  await writeUpdateCache({ checkedAt: Date.now(), latest, distTag: channel });
+  // Only refresh the update-notification cache when every attempted install
+  // succeeded — otherwise a partial --all failure would silently mark the
+  // latest version as cached.
+  const anyFailed = results.some((r) => !r.ok);
+  if (!anyFailed) {
+    await writeUpdateCache({ checkedAt: Date.now(), latest, distTag: channel });
+  }
 
   // Summary + skipped installs when --all.
   if (options.all) {
@@ -275,8 +312,9 @@ export async function update(options: UpdateOptions): Promise<void> {
     log.info("Summary:");
     for (const r of results) {
       const icon = r.ok ? green("✓") : yellow("✗");
+      const primaryTag = r.target === primary ? dim(" [primary]") : "";
       const suffix = r.ok ? "" : ` ${yellow(`- ${r.error}`)}`;
-      log.info(`  ${icon} ${formatTarget(r.target)}${suffix}`);
+      log.info(`  ${icon} ${formatTarget(r.target)}${primaryTag}${suffix}`);
     }
     for (const t of others) {
       const skip = whyCantUpdate(t, channel);
@@ -294,7 +332,6 @@ export async function update(options: UpdateOptions): Promise<void> {
   }
 
   if (isHuman()) {
-    const anyFailed = results.some((r) => !r.ok);
     outro(anyFailed ? "Update completed with errors" : `Successfully updated to ${latest}`);
   }
 }
--- a/packages/cli-core/src/commands/update/index.ts
+++ b/packages/cli-core/src/commands/update/index.ts
@@ -2,10 +2,13 @@ import { isHuman } from "../../mode.ts";
 import { green, cyan, yellow, dim } from "../../lib/color.ts";
 import { CliError } from "../../lib/errors.ts";
 import {
+  asdfPluginFromPath,
+  asdfReshim,
   findClerkOnPath,
   getInstallerPackageDirs,
   globalInstallCommand,
   ownerOfBinary,
+  resolveAsdfShim,
   type Installer,
 } from "../../lib/installer.ts";
 import { log } from "../../lib/log.ts";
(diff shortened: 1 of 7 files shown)


wyattjoh (commented):
## Code Review — PR #179

### Major

**M1. `brew upgrade clerk` now auto-runs on stable; docs disagree** _(codex confirmed)_
`packages/cli-core/src/commands/update/index.ts:94-96,117-124,247` runs `brew upgrade clerk` via `runGlobalInstall`, but `packages/cli-core/src/commands/update/README.md:23-31` still says Homebrew "prints `brew upgrade clerk` and exits (no auto-install)". Pick a semantic. If auto-upgrade is intended, update the docs and consider a separate confirmation step; running `brew upgrade` under a silenced spinner hides brew side-effects (dependency upgrades, autoremoves).

**M2. `npx clerk update` / `bunx clerk update` silently regress** _(codex partial)_
The old code had `detectFromUserAgent` returning `"npm"` when `npm_config_user_agent=npm/...`. New `findClerkOnPath()` cannot see the npx cache at `~/.npm/_npx/<hash>/...` (not on PATH), and the fallback to `process.execPath` makes `ownerOfBinary` return `null`, so the command refuses with "outside any known package manager". Users who used to run `npx clerk update` now get a misleading install-sh error. Either preserve user-agent detection for this case, or emit an actionable "install globally first" message.

**M3. Changeset is `patch` but PR adds `--all` and changes behavior**
`.changeset/clerk-update-path-aware.md` bumps `patch`. A new flag plus the Homebrew behavior change lean `minor`.

- **Cache poisoning on partial `--all` failure.** `writeUpdateCache()` runs after the loop in `update/index.ts:240-256,270` even when some targets errored, so a failed run marks the latest version as cached and suppresses future update prompts until staleness.
- **Non-interactive auto-confirm without `--yes`.** `README.md:77` says agent mode requires `--yes`, but `update/index.ts:222-223` auto-confirms whenever `!isHuman()`, regardless of `--yes`.

### Minor

- `process.env.ASDF_DATA_DIR ?? join(...)` (`installer.ts:167-169,191-196`): `??` treats `""` as set. Use `||` or an explicit empty-check.
- `queryBunPackageDir` always returns a path via `safeRealpath` (`installer.ts:156-159`), contradicting the `getInstallerPackageDirs` docstring that says absent PMs are omitted.
- `ownerOfBinary.startsWith` on Windows (`installer.ts:218-230`): realpath case differences on drive letters can cause spurious "outside any known package manager" refusals.
- `resolveTargets` fallback to `process.execPath` (`update/index.ts:64-66`) pairs with M2 to surface misleading errors for package-runner invocations.
- `hashHint` (`update/index.ts:146-155`) falls through to POSIX `hash -r` on Windows when `$SHELL` is unset.
- `--all` summary doesn't distinguish primary-install status; a user whose shell-resolved clerk failed can miss that fact.

### Positives

Excellent unit-test coverage for `findClerkOnPath`, `ownerOfBinary`, and the asdf helpers. The refuse-rather-than-guess ownership model is a real improvement over the prior silent npm fallback. The bun shim-vs-install-dir diagnosis is correct, and the `realpath(mkdtemp(...))` trick to paper over `/var -> /private/var` is exactly the kind of detail that usually gets missed.

wyattjoh (commented):
## Code review

Ambitious rewrite of clerk update's installer resolution. The PATH-walking and ownership-by-install-dir design is solid and the test coverage for ownerOfBinary, findClerkOnPath, and the asdf helpers is thorough. A few issues stood out: the homebrew branch silently misreports success when the tap lags, --all can no-op when the primary is a refuse-case, and the update-cache gate holds primary success hostage to ancillary failures.

packages/cli-core/src/commands/update/index.ts
```diff
@@ -40,6 +91,9 @@ async function runGlobalInstall(installer: Installer, packageSpec: string): Prom
     case "yarn":
       result = await Bun.$`yarn global add ${packageSpec}`.quiet().nothrow();
       break;
+    case "homebrew":
```
wyattjoh, line 98: **[HIGH] `brew upgrade clerk` ignores packageSpec and can silently install the wrong version**

The new homebrew branch runs `brew upgrade clerk`, which upgrades to whatever version the tap currently publishes — not to `packageSpec` (e.g. `clerk@0.8.5`). If the tap lags the npm release (formula-bump PR hasn't merged yet), the spinner reports `Updated homebrew` and the command prints `Successfully updated to 0.8.5`, but `clerk -v` still shows the old version. `brew upgrade` also exits 0 when already-latest, so there's no signal to fall back on. Either verify the installed version after `brew upgrade` and fail when it doesn't match `latest`, or check `brew info --json=v2 clerk` first and warn when the tap is stale.

case "homebrew":
  result = await Bun.$`brew upgrade ${UPDATE_PACKAGE_NAME}`.quiet().nothrow();
  break;

</details>

packages/cli-core/src/commands/update/index.ts
```diff
@@ -57,47 +111,147 @@ async function runGlobalInstall(installer: Installer, packageSpec: string): Prom
...
-  // Homebrew: cannot auto-update — instruct the user instead
-  if (installer === "homebrew") {
-    log.info(`  Installed via Homebrew. To update, run:`);
-    log.info(`  ${cyan("brew upgrade clerk")}`);
-    log.blank();
-    if (isHuman()) outro("Run `brew upgrade clerk` to update");
+  // Primary target can't be updated by us: refuse (don't guess a different installer).
+  const primarySkip = whyCantUpdate(primary, channel);
+  if (primarySkip) {
+    const runner = detectPackageRunner();
```
wyattjoh, line 275: **[HIGH] `--all` refuses when the primary is unrecoverable even if other installs are updatable**

When the primary target is e.g. Homebrew on `--channel canary`, `whyCantUpdate(primary, channel)` returns non-null and the command returns with `Update required manual action` — even if `--all` was passed and there are bun/npm installs further down PATH that are perfectly updatable. Users running `clerk update --all --channel canary` on a mixed setup would expect Homebrew to be skipped (with a warning) and the other installs to proceed. Consider: when `--all` is set and the primary isn't updatable, promote the first updatable `other` to the effective primary (or iterate all updatable targets without a designated primary).

const primarySkip = whyCantUpdate(primary, channel);
if (primarySkip) {
  ...
  reportOtherInstalls(others, channel);
  if (isHuman()) outro("Update required manual action");
  return;
}

</details>

packages/cli-core/src/commands/update/index.ts
```diff
@@ -57,47 +111,147 @@ async function runGlobalInstall(installer: Installer, packageSpec: string): Prom
...
+  return "If `clerk` still points to the old binary, run `hash -r` or open a new shell.";
+}
+
+/**
+ * Detect invocation via a package runner (npx/bunx/pnpm dlx). The binary lives
+ * in a runner cache (e.g. `~/.npm/_npx/<hash>/...`) not on PATH, so `ownerOfBinary`
+ * can't identify it — but we can recognize the runner from env vars it sets.
+ */
+function detectPackageRunner(): "npx" | "bunx" | null {
+  const ua = (process.env.npm_config_user_agent ?? "").toLowerCase();
```
wyattjoh: **[MEDIUM] detectPackageRunner matches plain npm, not just npx**

`ua.startsWith("npm/")` is true for *any* invocation where npm set `npm_config_user_agent` — not just npx. A user who has a globally-installed clerk and runs an `npm run` script that calls `clerk update` will get the `Running via npx; no installed clerk to update` warning instead of the real update flow. Tighten the check to require the `execPath.includes("npx")` signal, and drop the `ua.startsWith("npm/")` disjunct (or scope it to UAs that also contain `npx`).

if (execPath.includes("npx") || ua.startsWith("npm/")) return "npx";

</details>

packages/cli-core/src/lib/installer.ts
```diff
@@ -55,85 +104,156 @@ async function safeRealpath(p: string): Promise<string> {
...
+}
+
+/** Best-effort; swallows errors. */
+export async function asdfReshim(plugin: string): Promise<void> {
+  try {
+    await Bun.$`asdf reshim ${plugin}`.quiet().nothrow();
+  } catch {}
+}
+
+// ── Ownership decision ──────────────────────────────────────────────────────
```
wyattjoh, line 220: **[MEDIUM] ownerOfBinary Windows normalization misses extended-length and UNC paths**

`p.toLowerCase().replace(/\//g, sep)` handles the common case. It doesn't handle `\\?\C:\...` (extended-length) or `\\server\share\...` (UNC) — and Bun's realpath on Windows can return extended-length paths for deep trees. If `getInstallerPackageDirs()` returns a plain `C:\...` while realpath of the binary returns `\\?\C:\...`, the prefix match fails and the binary is mis-classified as `null`, triggering the refuse path. Either strip the `\\?\` prefix before comparison or use `path.relative`-based containment.

const normalize = (p: string): string =>
  process.platform === "win32" ? p.toLowerCase().replace(/\//g, sep) : p;

</details>

packages/cli-core/src/commands/update/index.ts
```diff
@@ -57,47 +111,147 @@ async function runGlobalInstall(installer: Installer, packageSpec: string): Prom
...
+  log.info(`Run ${cyan("clerk update --all")} to update them too.`);
+}
+
+/** Hint for invalidating the current shell's command-hash cache after update. */
+function hashHint(): string | null {
+  // Windows shells (cmd.exe, PowerShell) don't cache command paths, and `$SHELL`
+  // is typically unset — don't emit a POSIX hint in that case.
+  if (process.platform === "win32") return null;
+  const shell = (process.env.SHELL ?? "").toLowerCase();
+  if (shell.endsWith("/fish") || shell.endsWith("fish.exe")) return null; // auto-rehashes
```
wyattjoh: **[LOW] hashHint PowerShell branch is unreachable**

The `process.platform === "win32"` early-return on line above makes the `shell.endsWith("powershell.exe")` check unreachable. Not a bug (the return is correct — PowerShell doesn't cache), but the dead branch should either be deleted or the win32 early-return should be scoped to only non-PowerShell contexts (for PowerShell-on-Linux via `pwsh`, which does hit the later branch already).

if (process.platform === "win32") return null;
const shell = (process.env.SHELL ?? "").toLowerCase();
...
if (shell.endsWith("/pwsh") || shell.endsWith("powershell.exe")) return null;

</details>

wyattjoh (approved):

PR #182, opened 2026-04-17
fix(homebrew): desc must not start with formula name

wyattjoh:
## Summary
- Rephrase the generated Homebrew formula `desc` so it no longer starts with `Clerk`, which violates Homebrew's `FormulaAudit/Desc` rule and is causing `brew test-bot` failures in the `homebrew-stable` tap (e.g. the 0.8.4 release).
- Add a regression test asserting the rendered `desc` never starts with the formula name, for both unversioned and versioned formulas.

## Test plan
- [x] `bun run format`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test` (71 passed, incl. new regression)
- [ ] After merge and next release, verify `brew test-bot` is green in the homebrew-stable tap

### DIFF
--- a/scripts/lib/homebrew.test.ts
+++ b/scripts/lib/homebrew.test.ts
@@ -82,6 +82,34 @@ describe("renderFormula", () => {
 
     expect(result).toContain("keg_only :versioned_formula");
   });
+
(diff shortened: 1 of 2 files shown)


rafa-thayto (approved):
lgtm

PR #183, opened 2026-04-20
feat(log): expand --verbose diagnostics and surface silent env fallback

wyattjoh:
## Summary

Expand `--verbose` diagnostics across the CLI so network, credential, and config failures come with enough context to diagnose in one run. Also warns users when a saved environment is not available in the current binary (previously a silent fallback to production) and cleans up a few snapshot release rough edges surfaced along the way.

## What's new

- **`--verbose` coverage**: every outbound HTTP call logs URL, method, status, and response body on error; credential store, config I/O, environment resolution, OAuth callback, git detection, framework detection, autolink, and package-manager runner probing all gain debug lines. All HTTP now goes through a shared `loggedFetch` helper, codified in a new `.claude/rules/` entry so future call sites can't forget the pattern.
- **Env fallback warning**: when the CLI falls back from a saved environment to production because the profile isn't compiled in, it now warns at the info level and lists the available environments. Previously, the only symptom was a surprise 401/403 from the next API call.
- **Snapshot releases**: the version string now embeds the commit SHA (e.g. `0.8.6-snapshot.a1b2c3d`) instead of a datetime; the job refuses to re-publish an already-published commit (use `!snapshot <tag>` to force a re-run); snapshot builds now inherit the env-profile secrets the same way canary and stable do, so `switch-env staging` works on a snapshot install; platform packages publish in parallel, wrapper still follows sequentially.

## Test plan

- [x] Unit + integration suite passes
- [x] Snapshot build verified end-to-end on this PR
- [ ] Spot-check `switch-env staging` + an API call on the final snapshot install


wyattjoh:
**Stack: wyattjoh/roomy-arrhinceratops**

- **#183 👈 this PR**

wyattjoh:
!snapshot

wyattjoh:
!snapshot

wyattjoh:
!snapshot

wyattjoh:
!snapshot

.changeset/tired-pillows-divide.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
rafa-thayto: This changeset is empty -- it has frontmatter markers but no package entry and no description. `changeset version` will silently ignore it, but it'll sit in the repo forever since nothing consumes it.

Was this left over from testing? If it's intentional (to satisfy the enforce-changeset CI check for the scripts/ changes), `bun changeset --empty` would be the canonical way. But `roomy-arrhinceratops.md` already covers the `clerk` package bump, so this file can probably just be deleted.

```suggestion
```

scripts/lib/npm.ts
```diff
@@ -1,15 +1,16 @@
...
+export async function run(cmd: string[], opts?: { cwd?: string }): Promise<void> {
+  const proc = Bun.spawn(cmd, {
     cwd: opts?.cwd,
-    stdio: ["ignore", "pipe", "pipe"],
+    stdio: ["ignore", "ignore", "pipe"],
   });
-  if (result.exitCode !== 0) {
-    const stderr = result.stderr.toString().trim();
+  const exitCode = await proc.exited;
+  if (exitCode !== 0) {
```
rafa-thayto, line 10: Small thing: `run()` now pipes stderr but ignores stdout. That's fine for the current callers (`npm publish`, `git tag`, etc.), but worth noting that if a future caller needs stdout it'll silently get nothing.

More importantly: when reading stderr after `proc.exited`, the stream may already be closed. In Bun, `proc.stderr` is a `ReadableStream` -- wrapping it with `new Response(...).text()` works, but only if the stream hasn't already been consumed or garbage collected. Since you `await proc.exited` first and only then read stderr on the error path, the data should be buffered. This looks correct for Bun's current behavior, just flagging it as something to keep an eye on if Bun changes stream lifecycle semantics.

scripts/releaser.ts
```diff
@@ -67,16 +67,18 @@ console.log(
...
-  const name = packageName(target.name);
-  if (isPublished(name, version)) {
-    console.log(`Skipping ${name}@${version} (already published)`);
-    continue;
-  }
-  console.log(`Publishing ${name}@${version}...`);
-  const dir = await generatePlatformPackage(target, version);
-  publish(dir, { dryRun, tag });
-}
+await Promise.all(
```
rafa-thayto, line 70: Nice speedup! A few things to think about with parallel publish:

1. **`generatePlatformPackage` writes to `DIST_DIR/<target.name>/`** -- each target gets its own subdirectory, so there's no filesystem race. Looks safe.

2. **npm publish with `--provenance`** -- npm's provenance relies on the OIDC token from the GitHub Actions runner. Multiple concurrent `npm publish` calls sharing the same OIDC token *should* be fine since each gets its own provenance statement, but if you ever see intermittent `401`s or provenance failures in CI, this parallelism would be the first thing I'd look at.

3. **Error handling** -- if one target fails, `Promise.all` rejects immediately and the remaining in-flight publishes continue in the background (their results are just ignored). That's probably fine since npm publish is idempotent for the same version, but a `Promise.allSettled` + check-for-failures pattern would give cleaner error reporting if you wanted it.

None of these are blockers, just things to be aware of.

scripts/snapshot.ts
```diff
@@ -50,7 +51,36 @@ try {
...
+    stdio: ["ignore", "pipe", "pipe"],
+  });
+  if (shortShaResult.exitCode !== 0) {
+    throw new Error(
+      `git rev-parse --short HEAD failed: ${shortShaResult.stderr.toString().trim()}`,
+    );
+  }
+  const shortSha = shortShaResult.stdout.toString().trim();
+  const originalVersion: string = pkg.version;
+  const finalVersion = originalVersion.replace(/[a-f0-9]{40}$/, shortSha);
```
rafa-thayto: The regex `/[a-f0-9]{40}$/` assumes the version string always ends with a full 40-char SHA. That holds today because `prereleaseTemplate` is `{tag}.{commit}`, but if someone adds a suffix to the template later (e.g. `{tag}.{commit}.1`), this silently stops matching and the version goes out with a 40-char SHA.

A slightly more defensive pattern:
```suggestion
  const finalVersion = originalVersion.replace(/\b[a-f0-9]{40}\b/, shortSha);
```
This matches the SHA anywhere in the string (word-bounded), so it's resilient to template changes. The `\b` anchors prevent matching partial hex strings.

packages/cli-core/src/lib/plapi.ts
```diff
@@ -32,12 +33,20 @@ export async function getAuthToken(): Promise<string> {
   const key = process.env.CLERK_PLATFORM_API_KEY;
   if (key) {
     validateKeyPrefix(key, "ak_");
+    log.debug(
+      `plapi: using CLERK_PLATFORM_API_KEY for auth (env=${getCurrentEnvName()}, target=${getPlapiBaseUrl()})`,
```
rafa-thayto, line 38: This logs the existence of `CLERK_PLATFORM_API_KEY` with the target URL, which is fine. Just double-checking: `log.debug` doesn't log the key value itself, only that it's being used. Looks good.

One thought -- these debug lines repeat `env=${getCurrentEnvName()}, target=${getPlapiBaseUrl()}` in two places (here and line 46). You could pull those into a helper or use `log.withTag("plapi")` + a one-time init log to reduce the repetition. Not a big deal though.

packages/cli-core/src/lib/environment.ts
```diff
@@ -49,12 +55,29 @@ function loadFileProfiles(): Record<string, EnvProfileConfig> | undefined {
...
+    profilesSourceLogged = true;
+    log.debug(
+      `env: profiles from defaults — no CLI_ENV_PROFILES and no .env-profiles.json (${Object.keys(DEFAULT_PROFILES).join(", ")})`,
+    );
+  }
+  return DEFAULT_PROFILES;
 }
 
 let currentEnvName: string | undefined;
+let profilesSourceLogged = false;
```
rafa-thayto: `profilesSourceLogged` is a module-level `let` declared after `currentEnvName` at line 81 in the PR diff, but it's read inside `loadFileProfiles()` and `getProfiles()` which are called *before* the variable declaration in the source order. This works fine in practice because `let` declarations are hoisted (just not initialized until the declaration is reached, and these functions aren't called at module-load time). Not a bug, but it reads a bit oddly having the flag declared below the functions that use it.

Consider moving `let profilesSourceLogged = false;` up next to `currentEnvName` (or near the top of the module with the other state) so readers don't have to reason about hoisting.

.github/workflows/release.yml
```diff
@@ -403,6 +403,7 @@ jobs:
       version: ${{ needs.snapshot.outputs.version }}
       ref: ${{ needs.snapshot.outputs.sha }}
       artifact-prefix: clerk-snapshot
+    secrets: inherit
```
rafa-thayto, line 406: Good catch -- `build-binaries.yml` references `${{ secrets.ENV_PROFILES }}` (line 39 of that workflow), so without `secrets: inherit` the snapshot builds were getting an empty `ENV_PROFILES` and producing binaries with only the default production profile. This fixes that.

packages/cli-core/src/lib/plapi.ts
wyattjoh: Both log lines are in one function (`getAuthToken`) and only fire when a specific auth-source branch is taken. Extracting a helper for two adjacent ~60-char lines adds indirection without removing complexity, and collapsing to a one-time init log would lose the per-call visibility — each log currently shows *which* auth source was chosen *and* where it's pointing in a single line. Happy to revisit if the pattern spreads to a third call site.

wyattjoh:
!snapshot

rafa-thayto (approved):

rafa-thayto (approved):

PR #184, opened 2026-04-20
fix(init): tighten bootstrap flow (skip redundant confirm, put next steps last)

rafa-thayto:
## Summary

Two small bootstrap-flow polish fixes for `clerk init`:

- **Skip redundant "Proceed?" during bootstrap.** When bootstrapping a new project (either `--starter` or implicit bootstrap on an empty directory), the user already opted into project creation, so the secondary `? Proceed?` scaffold prompt is dropped. The scaffold preview (`CREATE`/`MODIFY` list) is still printed.
- **Next steps print last.** Moved the bootstrap next-steps block (`→ cd <project>`, `→ <pm> dev`, `→ clerk auth login`) to run after the "Install agent skills?" prompt, so those steps stay on-screen as the final thing the user sees when the command finishes.

Existing-project flows are unchanged.

## Test plan

- [x] `bun run format` / `lint` / `typecheck` / `test` pass
- [x] Unit tests cover:
  - Implicit bootstrap skips scaffold "Proceed?"
  - `--starter` without `-y` skips scaffold "Proceed?"
  - Existing-project flow still prompts "Proceed?" (regression guard)
- [ ] Manual: `clerk init --starter` in a fresh tmpdir — confirm no "Proceed?" after preview; confirm `→ cd … / → <pm> dev` appears as the final block before `└ Done`
- [ ] Manual: `clerk init` in an empty dir → same verification as above
- [ ] Manual: `clerk init` in an existing linked project → "Proceed?" still appears; no bootstrap next steps printed (unchanged behavior)

### DIFF
--- a/.changeset/init-adjustments.md
+++ b/.changeset/init-adjustments.md
@@ -2,4 +2,7 @@
 "clerk": patch
 ---
 
-Skip the redundant "Proceed?" scaffold confirmation when `clerk init` bootstraps a new project (via `--starter` or on an empty directory). The scaffold plan is still previewed; only the now-superfluous prompt is removed since the user already opted in by starting bootstrap.
+Tighten the `clerk init` bootstrap flow:
+
+- Skip the redundant "Proceed?" scaffold confirmation when bootstrapping a new project (via `--starter` or on an empty directory). The scaffold plan is still previewed; only the now-superfluous prompt is removed since the user already opted in by starting bootstrap.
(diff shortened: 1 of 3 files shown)


Railly (approved):
**Nit**: Tests assert call counts but not order, if someone moves `printBootstrapNextSteps` above `installSkills` again, nothing catches it. Capture `captured.out` and assert next-steps appears after the skills call. ([`index.test.ts:172-214`](https://github.com/clerk/cli/blob/ff1a815/packages/cli-core/src/commands/init/index.test.ts#L172-L214))

Overall, LGTM

PR #185, opened 2026-04-20
feat(init): install all core and feature skills by default

rafa-thayto:
## Summary

- Expands `DEFAULT_SKILLS` from 2 (`clerk`, `clerk-setup`) to 7 by also installing `clerk-custom-ui`, `clerk-backend-api`, `clerk-orgs`, `clerk-testing`, and `clerk-webhooks` on every `clerk init` run. Framework-specific skills still layer on top, one per detected dep.
- Drops `express` and `fastify` from `FRAMEWORK_SKILL_MAP` — they no longer need a framework-specific skill because `clerk-backend-api` is now a default for all projects.
- Summarizes the confirm prompt so 8 slugs don't wrap a narrow terminal: `Install agent skills? (clerk core + features + nextjs-patterns)`. Pass `--no-skills` to opt out, same as before.

## Test plan

- [x] `bun run format` / `bun run lint` / `bun run typecheck` — pass
- [x] `bun test packages/cli-core/src/commands/init/skills.test.ts` — 10/10 pass (5 new `resolveSkills` cases, 2 new `formatSkillsPromptMessage` cases, 3 existing `buildSkillsArgs` cases)
- [x] `bun changeset status --since=origin/main` — exit 0 (patch bump for `clerk`)
- [x] Manual smoke: run `clerk init` in a throwaway Next.js app and confirm the prompt reads `Install agent skills? (clerk core + features + nextjs-patterns)` and the post-accept log line lists all 8 skills
- [x] Manual smoke: run `clerk init` in an Express project and confirm only the 7 defaults install (no duplicate `clerk-backend-api`)
- [x] Manual smoke: run `clerk init --no-skills` and confirm the install step is skipped entirely

### DIFF
--- a/packages/cli-core/src/commands/init/skills.ts
+++ b/packages/cli-core/src/commands/init/skills.ts
@@ -7,9 +7,10 @@
  *     imports). Delegated to the `clerk skill install` core helpers in
  *     `commands/skill/install.ts`.
  *
- *  2. The framework-pattern skills (`clerk-setup`,
- *     `clerk-<framework>-patterns`) ship from the upstream `clerk/skills`
- *     repo and version independently of the CLI.
+ *  2. The upstream skills (`clerk-setup`, `clerk-custom-ui`,
+ *     `clerk-backend-api`, `clerk-orgs`, `clerk-testing`, `clerk-webhooks`,
+ *     plus a framework-specific skill when one matches) ship from the
+ *     upstream `clerk/skills` repo and version independently of the CLI.
  *
  * The skills CLI itself handles agent auto-detection and scope selection:
  * in interactive mode we hand off entirely (no `--agent` / `-y`), so the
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/commands/init/skills.ts
```diff
@@ -76,11 +93,12 @@ export async function installSkills(
   skipPrompt: boolean,
 ): Promise<void> {
   const skills = resolveSkills(frameworkDep);
+  const frameworkSkill = getFrameworkSkill(frameworkDep);
   const skillList = skills.join(", ");
```
wyattjoh: **[MEDIUM] Post-confirmation log line undoes the prompt's shortening**

The prompt was shortened to `Install agent skills? (clerk core + features + nextjs-patterns)` specifically so it doesn't wrap on narrow terminals. But after the user accepts, the next log line is `Installing skills with \`bun dlx\`: \`clerk, clerk-setup, clerk-custom-ui, clerk-backend-api, clerk-orgs, clerk-testing, clerk-webhooks, clerk-nextjs-patterns\`` — all 8 slugs comma-joined. That defeats the motivation for the shorter prompt. Consider either using the same summary form in this log line or dropping the full list to `log.debug`.

log.info(`Installing skills with \`${runner.display}\`: \`${skillList}\``);

</details>

packages/cli-core/src/commands/init/skills.ts
```diff
@@ -25,10 +25,20 @@ import {
...
+  "clerk-setup",
+  "clerk-custom-ui",
+  "clerk-backend-api",
+  // features/
+  "clerk-orgs",
+  "clerk-testing",
+  "clerk-webhooks",
+];
+
+// Express/Fastify have no entry — their skill is clerk-backend-api, which is a default.
```
wyattjoh, line 39: **[MEDIUM] Set-based dedupe silently masks future FRAMEWORK_SKILL_MAP overlap**

The shift from `push` to `new Set(DEFAULT_SKILLS).add(...)` defends against accidental overlap (e.g. if someone re-adds `express: "clerk-backend-api"` later), but the silent dedupe means dead map entries would never fail a test. Worth either an assertion at module load that FRAMEWORK_SKILL_MAP and DEFAULT_SKILLS are disjoint, or a unit test that fails when a framework entry duplicates a default. Cheap, catches a real maintenance foot-gun.

export function resolveSkills(frameworkDep: string | undefined): string[] {
  const skills = new Set(DEFAULT_SKILLS);
  if (frameworkDep && FRAMEWORK_SKILL_MAP[frameworkDep]) {
    skills.add(FRAMEWORK_SKILL_MAP[frameworkDep]);
  }
  return [...skills];
}

</details>

wyattjoh (commented):
## Code review

Small, focused change. Expands the default skill set, drops the now-redundant Express/Fastify entries, and shortens the confirm prompt. Tests are tight. Two small consistency suggestions below.

wyattjoh (approved):

PR #187, opened 2026-04-20
fix(config): Normalize keys and improve help text

dmoerner:
Agents got confused and were trying to use comma-separated individual keys. Specify in the help texts that these are handler keys, and allow comma-separated keys.

### DIFF
--- a/.changeset/aie-866-consider-whether-config-schema-keys-should-take-a-comma.md
+++ b/.changeset/aie-866-consider-whether-config-schema-keys-should-take-a-comma.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Accept comma-separated values for `--keys` in `config pull` and `config schema`, and clarify that keys refer to top-level config sections.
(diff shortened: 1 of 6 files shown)


Railly:
A few notes:

- `plapi.ts` `appendKeys`: `--keys "auth_email, session"` -> Commander passes `["auth_email, session"]` → split yields `[" session"]` which is truthy, so the URL ends up with `keys=auth_email&keys=+session` (space URL-encoded). `.trim()` inside the loop fixes it.
- Help text says `(e.g. auth_email, session)` (comma) but examples use `--keys auth_email session` (space). `"separate with spaces or commas"` makes both paths explicit.
- Nit: new `schema.test.ts` case is missing the `expect(requestedUrl).toContain("/config/schema")` assertion its sibling has.

Only the `.trim()` feels blocking.

Railly (approved):

dmoerner:
Thank you, comments addressed in [9adb492](https://github.com/clerk/cli/pull/187/commits/9adb49278fa269bd72e7c97f3f5ba3ce30031aec)

PR #188, opened 2026-04-20
feat(push): dry-run diff

dmoerner:
We already support diffs on push (patch/put). Add diffs for dry-runs as well.

Sample output (includes color/dim/bolding):

```
[LOCAL]
◇  Fetching current config

[dry-run] Would PATCH config on My Application (development):

  connection_oauth_google:
    enabled:
      - true
      + false
```

### DIFF
--- a/packages/cli-core/src/commands/config/push.ts
+++ b/packages/cli-core/src/commands/config/push.ts
@@ -72,48 +72,26 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
   // Strip config_version — it's returned by pull but not accepted by the backend
   delete configPayload.config_version;
 
-  if (options.dryRun) {
-    const currentConfig = await withSpinner("Fetching current config...", () =>
-      withApiContext(
-        fetchInstanceConfig(ctx.appId, ctx.instanceId),
-        "Failed to fetch current config",
-      ),
-    );
-    delete currentConfig.config_version;
-
-    const isPatch = op.method === "PATCH";
-    const hasChanges = hasConfigChanges(currentConfig, configPayload, isPatch);
-
-    if (!hasChanges) {
-      log.info("[dry-run] No changes detected");
-      return;
-    }
-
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/commands/config/push.ts
```diff
@@ -73,8 +73,24 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
...
+
+    if (!hasChanges) {
+      log.info("[dry-run] No changes detected");
+      return;
+    }
+
+    log.info(`\n[dry-run] Would ${op.method} config on ${ctx.appLabel} (${ctx.instanceLabel}):\n`);
+    printDiff(currentConfig, configPayload, isPatch);
     return;
   }
```
rafa-thayto: The dry-run and push paths now do the same fetch, `config_version` strip, `isPatch`, and `hasConfigChanges` check back-to-back. Worth hoisting it above the `if (options.dryRun)` so the flow lives in one place:

```ts
const currentConfig = await withSpinner("Fetching current config...", () =>
  withApiContext(
    fetchInstanceConfig(ctx.appId, ctx.instanceId),
    "Failed to fetch current config",
  ),
);
delete currentConfig.config_version;
const isPatch = op.method === "PATCH";

if (!hasConfigChanges(currentConfig, configPayload, isPatch)) {
  log.info(options.dryRun ? "[dry-run] No changes detected" : "No changes detected");
  return;
}

const prefix = options.dryRun ? `[dry-run] Would ${op.method}` : op.verb;
log.info(`\n${prefix} config on ${ctx.appLabel} (${ctx.instanceLabel}):\n`);
printDiff(currentConfig, configPayload, isPatch);

if (options.dryRun) return;
```

Then the existing confirm + `apiFn` block below stays untouched.

dmoerner: Great suggestion of course. Cleaned up in [395c189](https://github.com/clerk/cli/pull/188/commits/395c18912ecf62bd83d8271816d3a2fbeca0287f)

rafa-thayto (approved):
LGTM

PR #189, opened 2026-04-20
fix(skill): fix clerk skill install failing on published releases

wyattjoh:
## Summary

- `bunx clerk@canary skill install` fails with `No valid skills found` because the bundled SKILL.md description contains `raw HTTP: it handles`. Strict YAML frontmatter parsers treat the unquoted `: ` as a nested mapping and reject the document, so the upstream `skills` CLI drops the skill as invalid.
- Rewrite the description as a YAML folded block scalar (`>-`) so the whole value is a plain scalar regardless of internal punctuation, including future prose containing an apostrophe.
- Add a regression test that parses the staged SKILL.md frontmatter via the already-present `yaml` dep, so a future stray `: ` fails locally instead of at install time.

## Test plan

- [x] `bun run test`, `bun run typecheck`, `bun run lint`, `bun run format:check`
- [x] Manual: `bunx skills@1.5.1 add <staged-dir>` against the fix now completes with `Done!` (reproduced in a temp dir, both with the CLI's own staging flow and a minimal stand-in)
- [ ] Re-verify once a canary publishes: `bunx clerk@<canary> skill install` in a fresh temp dir

### DIFF
--- a/skills/clerk/SKILL.md
+++ b/skills/clerk/SKILL.md
@@ -1,6 +1,12 @@
 ---
 name: clerk
-description: 'Operate the Clerk CLI (`clerk` binary) for authentication, user/org/session management, instance config, env keys, and any Clerk Backend or Platform API call. Use when the user mentions Clerk management tasks, "list clerk users", "create a clerk user", "update organization", "pull clerk config", "clerk env pull", "clerk doctor", "clerk api", or any ad-hoc Clerk API request. Prefer the CLI over raw HTTP: it handles auth, key resolution, app/instance targeting, and formatting automatically.'
+description: >-
+  Operate the Clerk CLI (`clerk` binary) for authentication, user/org/session management,
+  instance config, env keys, and any Clerk Backend or Platform API call. Use when the user
+  mentions Clerk management tasks, "list clerk users", "create a clerk user", "update
+  organization", "pull clerk config", "clerk env pull", "clerk doctor", "clerk api", or
+  any ad-hoc Clerk API request. Prefer the CLI over raw HTTP: it handles auth, key
+  resolution, app/instance targeting, and formatting automatically.
 ---
 
 # Clerk CLI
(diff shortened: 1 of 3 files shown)


wyattjoh:
**Stack: fix/skill-install-frontmatter**

- **#189 👈 this PR**

skills/clerk/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: clerk
-description: Operate the Clerk CLI (`clerk` binary) for authentication, user/org/session management, instance config, env keys, and any Clerk Backend or Platform API call. Use when the user mentions Clerk management tasks, "list clerk users", "create a clerk user", "update organization", "pull clerk config", "clerk env pull", "clerk doctor", "clerk api", or any ad-hoc Clerk API request. Prefer the CLI over raw HTTP: it handles auth, key resolution, app/instance targeting, and formatting automatically.
+description: 'Operate the Clerk CLI (`clerk` binary) for authentication, user/org/session management, instance config, env keys, and any Clerk Backend or Platform API call. Use when the user mentions Clerk management tasks, "list clerk users", "create a clerk user", "update organization", "pull clerk config", "clerk env pull", "clerk doctor", "clerk api", or any ad-hoc Clerk API request. Prefer the CLI over raw HTTP: it handles auth, key resolution, app/instance targeting, and formatting automatically.'
```
rafa-thayto: A folded block scalar sidesteps any future quoting gotcha (single-quoted scalars still need `''` to escape `'`, which is easy to slip into prose like `the user's app`):

```suggestion
description: >-
  Operate the Clerk CLI (`clerk` binary) for authentication, user/org/session management,
  instance config, env keys, and any Clerk Backend or Platform API call. Use when the user
  mentions Clerk management tasks, "list clerk users", "create a clerk user", "update
  organization", "pull clerk config", "clerk env pull", "clerk doctor", "clerk api", or
  any ad-hoc Clerk API request. Prefer the CLI over raw HTTP: it handles auth, key
  resolution, app/instance targeting, and formatting automatically.
```

rafa-thayto (approved):

PR #193, opened 2026-04-21
feat(auth): opt in to dashboard CLI OAuth flow via clerk_client=cli

djgould:
## Summary

Two additions to the CLI's OAuth flow:

1. **Pass `clerk_client=cli` on the OAuth authorize URL** so the Clerk dashboard can recognize CLI-originated sign-up flows and preserve the OAuth return URL through onboarding.
2. **Ensure the logged-in user has at least one application** after a successful OAuth login. If the user has zero applications, create one named "My Application" with defaults; otherwise, skip. Errors are swallowed (non-fatal) so a bootstrap hiccup never fails `clerk auth login`.

## Context

When a user runs a CLI command that triggers OAuth (`clerk auth login`, `clerk link`, `clerk init`), the browser opens the Clerk OAuth authorize URL. If the user isn't signed in, FAPI redirects to the dashboard sign-up page with the original authorize URL as `redirect_url`. Without the `clerk_client=cli` marker, the dashboard can't distinguish a CLI-driven sign-up from any normal sign-up and loses the CLI's return URL. This PR's first change closes that gap by tagging the authorize URL.

The second change moves responsibility for "user must have at least one application" from the dashboard (where it previously lived in `/prepare-account`) to the CLI (where the product decision about "what's a sensible default first app" belongs). This also removes a class of duplicate-app bugs caused by React 18 Strict Mode double-mount in the dashboard. The dashboard side of this move is in the companion PR.

## Companion PR

Dashboard side: https://github.com/clerk/dashboard/pull/9046

## Test plan

- [x] Unit: `login.test.ts` covers the `clerk_client` query-param on the authorize URL
- [x] Unit: `first-application.test.ts` covers idempotency (zero → create, any → skip, list-error → swallow, create-error → swallow) — 4/4 passing
- [x] Unit: `login.test.ts` covers `ensureFirstApplication` being called once after fresh OAuth AND not called on reused-session early-return — 15/15 passing
- [ ] Manual: run `bun run dev -- auth logout && bun run dev -- --verbose auth login` against staging; verify the verbose log shows either `ensure-first-app: user has N application(s), skipping` or `ensure-first-app: created app_...`


rafa-thayto (approved):
LGTM

packages/cli-core/src/lib/first-application.ts
```diff
@@ -0,0 +1,18 @@
...
+  try {
+    const apps = await listApplications();
+    if (apps.length > 0) {
+      log.debug(`ensure-first-app: user has ${apps.length} application(s), skipping`);
+      return;
+    }
+    const created = await createApplication(DEFAULT_FIRST_APPLICATION_NAME);
+    log.debug(`ensure-first-app: created ${created.application_id}`);
+  } catch (err) {
+    log.debug(`ensure-first-app: failed (non-fatal): ${String(err)}`);
```
rafa-thayto, line 17: If creation actually fails the user walks away without an app and the next `clerk link` / `clerk init` then fails with no clue why. Per `.claude/rules/debug-logging.md`, a non-obvious fallback like this deserves a `log.warn` alongside the debug.

```suggestion
    log.warn("Could not set up a default application. You can create one at https://dashboard.clerk.com.");
    log.debug(`ensure-first-app: failed (non-fatal): ${String(err)}`);
```

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -110,6 +112,10 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
 
   const userInfo = await performOAuthFlow();
 
+  // Best-effort: ensure the user has at least one application so downstream
+  // commands (clerk link, clerk init) have something to operate on.
+  await ensureFirstApplication();
```
rafa-thayto: Two network calls run here with no feedback, so the UI sits silent right after OAuth. Everywhere else in the flow uses `withSpinner` for slow ops, could match that:

```suggestion
  await withSpinner("Setting up your workspace...", () => ensureFirstApplication());
```

packages/cli-core/src/lib/first-application.test.ts
```diff
@@ -0,0 +1,43 @@
...
+    await ensureFirstApplication();
+    expect(mockCreateApplication).toHaveBeenCalledTimes(1);
+    expect(mockCreateApplication).toHaveBeenCalledWith(DEFAULT_FIRST_APPLICATION_NAME);
+  });
+
+  it("does not create when the user already has applications", async () => {
+    mockListApplications.mockResolvedValue([{ application_id: "app_existing" }]);
+    await ensureFirstApplication();
+    expect(mockCreateApplication).not.toHaveBeenCalled();
+  });
```
rafa-thayto, line 42: The PR description lists four scenarios but only two shipped. The error paths are the ones that actually guarantee the non-fatal behavior, worth covering:

```ts
it("swallows listApplications errors", async () => {
  mockListApplications.mockRejectedValue(new Error("boom"));
  await expect(ensureFirstApplication()).resolves.toBeUndefined();
  expect(mockCreateApplication).not.toHaveBeenCalled();
});

it("swallows createApplication errors", async () => {
  mockCreateApplication.mockRejectedValue(new Error("boom"));
  await expect(ensureFirstApplication()).resolves.toBeUndefined();
});
```

packages/cli-core/src/lib/first-application.ts
djgould: if the app creation fails they should still be able to create one via the cli?

rafa-thayto: Yeah, you're right, I think in this case we should say to they create the app using `clerk apps create` command, what do you think?

rafa-thayto (approved):
LGTM

PR #194, opened 2026-04-21
feat(help): suppress skill-install tip when skill is already installed

rafa-thayto:
## Summary

- Detect installed Clerk skill across common local agents (Claude Code, Codex, Cursor, Windsurf, Zed, Cline, VS Code, GitHub Copilot) — global + project-local paths.
- Gate the `Give AI agents better Clerk context…` tip in `clerk --help` and bare `clerk` output on that detection, so returning users don't see a nag on every help invocation.
- New synchronous helper `isClerkSkillInstalled()` (no shell-out, pure `fs.existsSync` — fits Commander's sync help-render path).

Follow-up to #191 which surfaced the tip unconditionally.

## Test plan

- [x] `bun run format && bun run lint && bun run typecheck && bun run test` all green
- [x] Manual: `HOME=\$(mktemp -d) bun run src/cli.ts --help` — tip visible
- [x] Manual: `bun run src/cli.ts --help` (real \$HOME with skill installed) — tip hidden
- [x] Unit tests cover all Tier-A agent dirs (global + project-local) and Tier-B VS Code/Copilot paths

### DIFF
--- a/packages/cli-core/src/cli-program.test.ts
+++ b/packages/cli-core/src/cli-program.test.ts
@@ -3,6 +3,7 @@ import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { createProgram, formatApiBody } from "./cli-program.ts";
+import { STANDARD_AGENT_DIRS, EXTRA_REL_PATHS } from "./lib/skill-detection.ts";
 
 describe("formatApiBody", () => {
   // --- Single error with meta ---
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/lib/skill-detection.ts
```diff
@@ -0,0 +1,52 @@
...
+export function isClerkSkillInstalled(): boolean {
+  const home = process.env.HOME ?? homedir();
+  const cwd = process.cwd();
+
+  for (const dir of STANDARD_AGENT_DIRS) {
+    if (existsSync(join(home, dir, STANDARD_SKILL_REL))) return true;
+    if (existsSync(join(cwd, dir, STANDARD_SKILL_REL))) return true;
+  }
+  for (const rel of EXTRA_REL_PATHS) {
+    if (existsSync(join(home, rel))) return true;
```
wyattjoh: `EXTRA_REL_PATHS` is documented here as project-local-only markers, but the loop also checks `join(home, rel)`. That means a stray `$HOME/.github/prompts/clerk.md` or `$HOME/.vscode/skills/clerk/SKILL.md` will suppress the help tip globally even though `clerk skill install` does not install those layouts as global agent directories. Restrict these extra paths to `cwd` or split the detector into explicit global and project-local path sets.

packages/cli-core/src/cli-program.test.ts
```diff
@@ -200,3 +203,90 @@ describe("formatApiBody", () => {
     expect(result).toBe("Plan limitation");
   });
 });
+
+describe("help: clerk skill install tip", () => {
+  const TIP_SUBSTR = "Give AI agents better Clerk context";
+  const STANDARD_AGENT_DIRS = [
```
wyattjoh: Maybe better to export this from the other file?

wyattjoh (approved):

rafa-thayto:
!snapshot

PR #196, opened 2026-04-21
docs(skill): document --all flag on clerk update

wyattjoh:
## Summary

- Add `--all` to the `clerk update` row in the bundled Clerk agent skill's command reference table.
- The flag already exists in the CLI but was missing from the skill, so agents couldn't surface it to users with multiple clerk installs on PATH.

## Test plan

- [ ] Confirm the skill's command reference table now lists `--all` under `clerk update`.


wyattjoh:
**Stack: wyattjoh/skill-document-update-all-flag**

- **#196 👈 this PR**

rafa-thayto (approved):

PR #197, opened 2026-04-21
fix(init,link): handle expired OAuth sessions gracefully

rafa-thayto:
## Summary

Stacked on top of #195. Closes the expired-session gap that's adjacent to that PR's subdirectory fix.

Before: a stale OAuth token on disk (user "logged in" but session expired) sneaks past the presence-only `isAuthenticated` check, then blows up downstream — `clerk link` surfaces a raw Platform API 401, `clerk init` opens a browser with no context, and keyless-capable frameworks never get the fallback they should.

After:
- `clerk init --starter` on a keyless-capable framework (Next.js, Astro, Nuxt, TanStack Start, React Router) with an expired session drops into keyless mode and prints a dim note explaining the fallback.
- `clerk init` on a non-keyless framework (React, Vue, Expo…) with an expired session prints "Your previous session expired — signing you back in…" *before* opening the browser.
- `clerk link` validates the session against `/userinfo` before hitting the Platform API. Standalone invocations get a distinct "Session expired" vs. "Not logged in" message depending on what's actually wrong. The no-token path skips the userinfo round-trip entirely.

## Implementation

- New `hasValidSession()` helper in `init/heuristics.ts` — network-validating sibling of `isAuthenticated` (presence-only).
- `init/index.ts` branches on validity (not presence) when deciding keyless vs. authenticated; emits the dim expiry note when the fallback fires.
- `link/index.ts` `ensureAuth` gates userinfo behind a presence check so the no-token path stays cheap.
- `getAuthenticatedEmail` memoized for the lifetime of a CLI invocation — three call sites (`init`'s `hasValidSession`, `resolveAuthLabel`, `link`'s `ensureAuth`) share one `/userinfo` round-trip. `login`/`logout` clear the cache on token mutation.

## Test plan

- [x] `bun run test` — 73/73 test files pass, 116 new assertions across `heuristics.test.ts` (memoization, cache invalidation, concurrent callers), `init/index.test.ts` (keyless fallback with + without expiry note, non-keyless pre-message), `link/index.test.ts` (distinct messages for no-token vs. expired).
- [x] `bun run typecheck` + `bun run format:check` clean.
- [x] Manually verified on branch with a real bogus token in the macOS Keychain:
  - `clerk whoami` → 1 userinfo call, "Session expired" JSON.
  - `clerk init --starter --framework next` (blank dir) → bootstraps Next.js, prints "starting in keyless mode" note, scaffolds with dev keys, 1 userinfo call.
  - `clerk link` in linked project → "Session expired. Authenticating first..." message, OAuth browser opens.
  - (Not re-verified since the final memoization tweak landed; unit tests cover it.)

### DIFF
--- a/.changeset/init-expired-session.md
+++ b/.changeset/init-expired-session.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Handle expired OAuth sessions gracefully in `clerk init` and `clerk link`. When a stored token is no longer valid, keyless-capable frameworks (e.g. Next.js) automatically drop into keyless mode during `clerk init --starter` with a note explaining the fallback; non-keyless frameworks print "Your previous session expired — signing you back in…" before opening the OAuth browser. `clerk link` now catches 401s from the Platform API and re-authenticates inline so an expired token can no longer surface as a raw error.
--- a/packages/cli-core/src/commands/init/README.md
+++ b/packages/cli-core/src/commands/init/README.md
@@ -45,11 +45,11 @@ Use `--prompt` to output a setup prompt for an AI agent without running init.
 
 1. **`--prompt`**: outputs a framework-specific prompt, then exits
 2. Gathers project context (framework, router variant, TypeScript, `src/` directory, package manager)
-3. Determines auth mode from credential presence (no user prompt):
-   - **Authenticated** (OAuth token or `CLERK_PLATFORM_API_KEY` set): uses the authenticated flow — runs `clerk link` if not already linked and pulls real API keys into `.env` at the end
-   - **Bootstrap + keyless-capable framework + not authenticated**: automatically uses keyless mode — the app runs on auto-generated dev keys and the user can connect a Clerk account later with `clerk auth login`
-   - **Bootstrap + non-keyless framework + not authenticated** (with `--yes` or agent mode): skips authentication and prints manual setup instructions (run `clerk auth login` / `clerk link` / `clerk env pull` when ready)
-   - **Existing project + not authenticated**: runs the authenticated flow, which triggers an interactive login so real keys can be pulled
+3. Determines auth mode from session validity (no user prompt). "Validity" means the stored OAuth token passes a `/userinfo` check or `CLERK_PLATFORM_API_KEY` is set — an expired or revoked token counts as "not authenticated" at the branch:
+   - **Valid session** (live OAuth token or `CLERK_PLATFORM_API_KEY`): uses the authenticated flow — runs `clerk link` if not already linked and pulls real API keys into `.env` at the end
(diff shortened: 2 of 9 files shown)


wyattjoh (commented):
Nice work on this. The `fetchAppWithReauth` helper and the inline retry in `resolveApp` land the UX exactly right: one round trip, one friendly "Session expired" message, one retry, no infinite loop.

One suggestion before this ships: consider centralizing the pattern in `lib/plapi.ts` rather than wrapping call sites one at a time.

The re-auth flow is identical across every PLAPI endpoint (401/403, then `login({ showNextSteps: false })`, then retry once, then bubble the second failure). Keeping it at the call site means the ~10 other PLAPI consumers that sit outside `link` will still surface a raw 401:

- `fetchApplication`: `apps/create.ts`, `env/pull.ts`, `api/index.ts` (secret-key resolution via `--app`), plus two more spots in `link/index.ts` (`createAndFetchApp`, `handleExistingProfile`)
- `listApplications`: `apps/list.ts`
- `fetchInstanceConfig` / `fetchInstanceConfigSchema` / `putInstanceConfig` / `patchInstanceConfig`: all `config/*` commands
- `createApplication`: `apps/create.ts`, `link/index.ts`

A single helper in `plapi.ts`, roughly:

```ts
export async function withReauthOnUnauthorized<T>(fn: () => Promise<T>): Promise<T> {
  try {
    return await fn();
  } catch (error) {
    if (!(error instanceof PlapiError) || (error.status !== 401 && error.status !== 403)) throw error;
    log.info("Session expired. Re-authenticating...");
    const { login } = await import("../commands/auth/login.ts");
    await login({ showNextSteps: false });
    return fn();
  }
}
```

…lets every PLAPI call site opt in with one wrapper, and `fetchAppWithReauth` / `resolveApp`'s inline retry collapse into the same call. Deliberate non-candidates (`bapiRequest` in `api/index.ts` uses a user-provided secret key, `autolink.ts` is best-effort silent, `doctor/context.ts` should report rather than re-auth, token-exchange endpoints would loop) stay uncovered on purpose, which is easy to call out at the call site since they don't wrap.

Benefits:

- Single test ("401 triggers login + retry", "second 401 bubbles") covers the whole surface.
- New PLAPI endpoints automatically get the behavior instead of having to remember the dance.
- The "no infinite retry" invariant lives in one place.

Happy to take this on as a follow-up PR on top of yours if the scope expansion is out of bounds for this one. The init+link-specific flow here stands on its own either way.

rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

rafa-thayto:
closed because we are now refreshing the token: #205

PR #198, opened 2026-04-21
fix(build): install all platform keyring bindings for cross-compile

wyattjoh:
## Summary

Cross-compiled macOS and Windows CLI binaries were silently falling back to plaintext-file credential storage instead of using the OS keychain.

Root cause: the build runs on a Linux CI host. Default `bun install` skips `@napi-rs/keyring` optional deps whose `os` or `cpu` field doesn't match the host, so the darwin and win32 bindings are not in `node_modules` when the cross-compile runs. Bun then rewrites the unresolved `require()` calls to throw stubs and the bundle ships without the native binding. At runtime, the `import()` fails, the catch swallows the error, and the CLI falls back to `~/Library/Application Support/clerk-cli/credentials` (chmod 600).

## Fix

- `bun install --frozen-lockfile --cpu='*' --os='*'` in `.github/workflows/build-binaries.yml` so every target's binding is installed before the Linux host bundles each target.
- Preflight check in `scripts/build.ts` so local `build:compile:all` runs on a developer Mac also fail fast with the install command if a required binding is missing.

Lockfile unchanged; `--frozen-lockfile` still holds.

## Test plan

- [x] Reproduced the failure in a Docker `ubuntu:24.04` + Bun 1.3.11 container matching CI, then applied the fix and confirmed every target bundle embeds its platform's native `.node` asset.
- [x] Pulled the Docker-built `darwin-arm64` binary onto a Mac, signed with a real Team ID + `scripts/entitlements.plist`, and confirmed `clerk whoami --verbose` reports `found token in keyring`.
- [x] Removed two bindings and reran `scripts/build.ts`; preflight check fired with the expected error and install hint.
- [x] `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test` all pass.
- [ ] Validate a released canary on a clean Mac and confirm new logins land in the Keychain.

### DIFF
--- a/.changeset/fix-keyring-cross-compile.md
+++ b/.changeset/fix-keyring-cross-compile.md
@@ -1,0 +1,8 @@
+---
+"clerk": patch
+---
+
+Store macOS credentials in the system Keychain instead of a plaintext file.
+
+- Previously, macOS builds silently stored the OAuth token in `~/Library/Application Support/clerk-cli/credentials` because cross-compiled binaries were missing the native Keychain binding.
+- Run `clerk login` after upgrading so the CLI writes a fresh token into the Keychain and removes the old plaintext file.
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -31,7 +31,14 @@ jobs:
       - uses: oven-sh/setup-bun@v2
         with:
           bun-version: "1.3.11"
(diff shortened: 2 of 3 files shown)


wyattjoh:
**Stack: wyattjoh/fix-keyring-cross-compile**

- **#198 👈 this PR**

rafa-thayto (approved):

PR #200, opened 2026-04-21
fix(config): Run server-side dry-run

dmoerner:
When we implemented config diffing we started doing the diffing client-side. I think this is still generally the correct behavior. However, we should also run and display the server-side dry-run on a dry-run, since this actually runs the validation before a change, which is very valuable.

### DIFF
--- a/packages/cli-core/src/commands/config/push.test.ts
+++ b/packages/cli-core/src/commands/config/push.test.ts
@@ -544,7 +544,7 @@ describe("config push", () => {
       dryRun: true,
     });
     expect(patchUrl).toContain("dry_run=true");
-    expect(captured.err).toContain("[dry-run] Would PATCH");
+    expect(captured.err).toContain("[dry-run] Proposing PATCH");
     expect(captured.err).toContain("- 604800");
     expect(captured.err).toContain("+ 3600");
     expect(captured.out).toContain(JSON.stringify(mockResponse, null, 2));
@@ -588,7 +588,7 @@ describe("config push", () => {
 
     await runConfigPut({ json: '{"session":{"lifetime":3600}}', dryRun: true });
     expect(putUrl).toContain("dry_run=true");
-    expect(captured.err).toContain("[dry-run] Would PUT");
+    expect(captured.err).toContain("[dry-run] Proposing PUT");
   });
 
   test("dry-run surfaces server validation errors", async () => {
@@ -610,7 +610,7 @@ describe("config push", () => {
 
     await expect(
       runConfigPatch({ json: '{"session":{"lifetime":3600}}', dryRun: true }),
-    ).rejects.toThrow("API error");
+    ).rejects.toThrow("invalid organization_settings");
   });
 
   // --- API error handling ---
--- a/packages/cli-core/src/commands/config/push.test.ts
+++ b/packages/cli-core/src/commands/config/push.test.ts
@@ -525,10 +525,10 @@ describe("config push", () => {
 
   // --- Dry run ---
 
-  test("dry-run fetches current config and shows diff without mutating", async () => {
-    let mutatingCallMade = false;
-    stubFetch(async (_input, init) => {
-      if (init?.method && init.method !== "GET") mutatingCallMade = true;
+  test("dry-run sends PATCH with ?dry_run=true and prints diff and projected result", async () => {
+    let patchUrl = "";
+    stubFetch(async (input, init) => {
+      if (init?.method === "PATCH") patchUrl = input.toString();
       const body = init?.method && init.method !== "GET" ? mockResponse : currentConfig;
       return new Response(JSON.stringify(body), { status: 200 });
     });
@@ -543,13 +543,21 @@ describe("config push", () => {
       json: '{"session":{"lifetime":3600}}',
       dryRun: true,
     });
-    expect(mutatingCallMade).toBe(false);
+    expect(patchUrl).toContain("dry_run=true");
     expect(captured.err).toContain("[dry-run] Would PATCH");
     expect(captured.err).toContain("- 604800");
     expect(captured.err).toContain("+ 3600");
+    expect(captured.out).toContain(JSON.stringify(mockResponse, null, 2));
+    expect(captured.err).toContain("[dry-run] Validation passed");
   });
 
-  test("dry-run reports no changes when payload matches current", async () => {
+  test("dry-run reports no changes when payload matches current without hitting server", async () => {
+    let mutatingCallMade = false;
+    stubFetch(async (_input, init) => {
+      if (init?.method && init.method !== "GET") mutatingCallMade = true;
+      return new Response(JSON.stringify(currentConfig), { status: 200 });
+    });
+
     await setProfile(process.cwd(), {
       workspaceId: "org_1",
       appId: "app_1",
@@ -560,10 +568,18 @@ describe("config push", () => {
       json: '{"session":{"lifetime":604800}}',
       dryRun: true,
     });
+    expect(mutatingCallMade).toBe(false);
     expect(captured.err).toContain("[dry-run] No changes detected");
   });
 
-  test("dry-run for put shows PUT method", async () => {
+  test("dry-run for put sends PUT with ?dry_run=true", async () => {
+    let putUrl = "";
+    stubFetch(async (input, init) => {
+      if (init?.method === "PUT") putUrl = input.toString();
+      const body = init?.method && init.method !== "GET" ? mockResponse : currentConfig;
+      return new Response(JSON.stringify(body), { status: 200 });
+    });
+
     await setProfile(process.cwd(), {
       workspaceId: "org_1",
       appId: "app_1",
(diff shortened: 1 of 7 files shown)


packages/cli-core/src/lib/plapi.ts
```diff
@@ -147,6 +147,9 @@ async function sendInstanceConfig(
   if (options?.destructive) {
     url.searchParams.set("destructive", "true");
   }
+  if (options?.dryRun) {
```
wyattjoh, line 150: The CLI now always appends `?dry_run=true` when `--dry-run` is passed. Most REST frameworks silently ignore unknown query parameters, so if any Platform API deployment this CLI can reach has not yet shipped `dry_run=true` support, the `--dry-run` becomes a real `PATCH`/`PUT` — a silent downgrade from "no side effects" to "full mutation."

Please confirm `dry_run` is live everywhere the CLI targets (dev, staging, prod, and any BAPI proxy variants), and ideally that the server rejects unknown query params so the CLI fails closed. A note in the PR description confirming server-side landed first would also help.

```ts
if (options?.dryRun) {
  url.searchParams.set("dry_run", "true");
}
```

packages/cli-core/src/commands/config/push.ts
```diff
@@ -103,16 +101,24 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
...
-        "Failed to push config",
-      ),
+  const spinnerMsg = options.dryRun
+    ? `[dry-run] Validating config on ${ctx.appLabel} (${ctx.instanceLabel})...`
+    : `${op.verb} config on ${ctx.appLabel} (${ctx.instanceLabel})...`;
+  const result = await withSpinner(spinnerMsg, () =>
+    withApiContext(
+      op.apiFn(ctx.appId, ctx.instanceId, configPayload, {
+        destructive: options.destructive,
+        dryRun: options.dryRun,
```
wyattjoh, line 111: Before this PR, `--dry-run` returned at `if (options.dryRun) return;` before reaching `log.data(JSON.stringify(result, null, 2))`. Now the projected config is written to **stdout** on every dry-run.

This is useful for agent-mode consumers, but it's still an observable surface change: scripts that previously ran `clerk config patch --dry-run` expecting no stdout output will now see a JSON blob. Worth calling out explicitly in `commands/config/README.md` ("stdout now contains the server's projected configuration JSON") and possibly in the changeset, so users piping the command aren't surprised.

packages/cli-core/src/commands/config/push.ts
```diff
@@ -103,16 +101,24 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
...
-  const result = await withSpinner(
-    `${op.verb} config on ${ctx.appLabel} (${ctx.instanceLabel})...`,
-    () =>
-      withApiContext(
-        op.apiFn(ctx.appId, ctx.instanceId, configPayload, { destructive: options.destructive }),
-        "Failed to push config",
-      ),
+  const spinnerMsg = options.dryRun
+    ? `[dry-run] Validating config on ${ctx.appLabel} (${ctx.instanceLabel})...`
+    : `${op.verb} config on ${ctx.appLabel} (${ctx.instanceLabel})...`;
```
wyattjoh, line 106: Previously `--dry-run` only needed a read-scoped flow (GET current config). After this change a `PATCH`/`PUT` request (with `dry_run=true`) is issued, so any middleware that authorizes on HTTP method rather than the `dry_run` query param may reject the call, or return a different error shape than a pure read.

Worth verifying bapi-proxy / BAPI middleware treats `?dry_run=true` consistently for authz (same scopes as a read, or a clear error if not), and documenting that `--dry-run` is no longer offline/read-only.

packages/cli-core/src/commands/config/push.ts
```diff
@@ -103,16 +101,24 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
...
   }
 
-  const result = await withSpinner(
-    `${op.verb} config on ${ctx.appLabel} (${ctx.instanceLabel})...`,
-    () =>
-      withApiContext(
-        op.apiFn(ctx.appId, ctx.instanceId, configPayload, { destructive: options.destructive }),
-        "Failed to push config",
-      ),
+  const spinnerMsg = options.dryRun
```
wyattjoh, line 104: Order is: print `[dry-run] Would PATCH` + client diff, then spin up the server validation. If validation fails, the user has just been shown a clean diff for a payload the server won't actually accept — mild cognitive dissonance.

Either validate first and then print the diff on success, or reword the diff header so it's clearly "proposed payload" rather than a commitment the server will accept it. Polish-level, not blocking.

packages/cli-core/src/commands/config/push.ts
```diff
@@ -103,16 +101,24 @@ async function configPush(options: ConfigPushOptions, op: Operation): Promise<vo
...
+        destructive: options.destructive,
+        dryRun: options.dryRun,
+      }),
+      options.dryRun ? "Dry-run failed" : "Failed to push config",
+    ),
   );
   log.data(JSON.stringify(result, null, 2));
-  log.success("Config pushed successfully");
+  log.success(
+    options.dryRun
```
wyattjoh, line 118: When `hasConfigChanges` returns false the command prints `[dry-run] No changes detected` and returns without calling the server. Practically fine (nothing to validate), but the changeset promises "validation errors are caught and the projected configuration is returned before changes are committed" — an invariant that isn't actually true on the no-change path.

Tiny wording fix to the changeset, or leave as-is and consider it understood.

packages/cli-core/src/commands/config/push.test.ts
```diff
@@ -560,20 +568,51 @@ describe("config push", () => {
...
+
+    await setProfile(process.cwd(), {
+      workspaceId: "org_1",
+      appId: "app_1",
+      instances: { development: "ins_dev" },
+    });
+
+    await expect(
+      runConfigPatch({ json: '{"session":{"lifetime":3600}}', dryRun: true }),
+    ).rejects.toThrow("API error");
```
rafa-thayto: The test name says it surfaces server validation errors, but we only assert the error class. The server's message (`invalid organization_settings`) never gets verified, so a regression in error formatting would slip through. Tightening the assertion catches that.

```suggestion
    await expect(
      runConfigPatch({ json: '{"session":{"lifetime":3600}}', dryRun: true }),
    ).rejects.toThrow("invalid organization_settings");
```

packages/cli-core/src/commands/config/push.test.ts
```diff
@@ -560,20 +568,51 @@ describe("config push", () => {
...
     await setProfile(process.cwd(), {
       workspaceId: "org_1",
       appId: "app_1",
       instances: { development: "ins_dev" },
     });
 
     await runConfigPut({ json: '{"session":{"lifetime":3600}}', dryRun: true });
+    expect(putUrl).toContain("dry_run=true");
     expect(captured.err).toContain("[dry-run] Would PUT");
   });
```
rafa-thayto, line 596: Small one: the PATCH dry-run test above asserts the projected result on stdout and the `Validation passed` line. The PUT test stops earlier, so a regression in either would slip through for PUT only. Worth mirroring.

```suggestion
    await runConfigPut({ json: '{"session":{"lifetime":3600}}', dryRun: true });
    expect(putUrl).toContain("dry_run=true");
    expect(captured.err).toContain("[dry-run] Would PUT");
    expect(captured.out).toContain(JSON.stringify(mockResponse, null, 2));
    expect(captured.err).toContain("[dry-run] Validation passed");
```

packages/cli-core/src/lib/plapi.ts
dmoerner: Yes, of course it's live everywhere haha

packages/cli-core/src/commands/config/push.ts
dmoerner: I think calling this out before we've done any releases is more likely to cause confusion than help anyone.

packages/cli-core/src/commands/config/push.ts
dmoerner: We do treat it consistently right now, although this is something that may change in the future. I'll make a note of this on the backend side.

packages/cli-core/src/commands/config/push.ts
dmoerner: I opted for the second approach, "proposed payload", since the first increases scope since we would need to handle corner cases where the server dry run has a different output (e.g., some settings forced on) than the client side.

packages/cli-core/src/commands/config/push.test.ts
dmoerner: Thank you, fixed!

packages/cli-core/src/commands/config/push.ts
dmoerner: I updated the changeset wording to fix this, thank you.

packages/cli-core/src/commands/config/push.test.ts
dmoerner: Thanks, adopted (and fixed the "Would" in the error assertion from Wyatt's comment)

wyattjoh (approved):

PR #201, opened 2026-04-21
fix(auth): isolate unsigned macOS keychain entries

wyattjoh:
## Summary

On macOS, local and unsigned CLI builds were writing OAuth tokens under the same Keychain service name (`clerk-cli`) as the signed release binary. macOS binds Keychain item ACLs to the code signature of the writer, so a later read from a differently signed (or unsigned) binary triggers a login-password prompt, and an accepted prompt grants that build permanent access to the release credential entry.

Net effect: running a locally compiled `clerk` (via `bun run build:compile`, `bun run src/cli.ts`, or a Bun-launched script) after a signed release login would either pop a password prompt or silently contaminate the release Keychain entry with a dev-build ACL.

## Fix

- Reserve the `clerk-cli` Keychain service for binaries that codesign-verify as an official release: `TeamIdentifier=L8SD6SB282` **and** `Identifier=clerk`, with `CLI_VERSION` populated (i.e. built by the release workflow, not `0.0.0-dev`).
- Route every other darwin invocation (unversioned dev builds, unsigned local compiles, `bun run` from source, non-Clerk signatures) to a new `clerk-cli-dev` service. Linux and Windows paths are unchanged.
- The classifier (`isReleaseSignedMacosBinary`) is a pure function over the `codesign -dvvv` output + `CLI_VERSION`, so it is covered by unit tests without shelling out. The per-process lookup is memoized in a promise so we run `codesign` at most once per CLI run.

Existing release credentials are not migrated. A developer who had previously approved a prompt will find their dev logins now live in `clerk-cli-dev`; the release `clerk` keeps using `clerk-cli` as before.

## Test plan

- [x] `bun test packages/cli-core/src/lib/credential-store-signing.test.ts` (new classifier: dev/unversioned, release, non-Clerk signer, mismatched identifier)
- [x] `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test`
- [ ] Manual on a Mac with a prior signed-release login: run a locally compiled `clerk whoami`, confirm no password prompt and `--verbose` reports the `clerk-cli-dev` service; then run the signed release `clerk whoami` and confirm it still reads from `clerk-cli`.

### DIFF
--- a/.changeset/clerk-cli-macos-keychain-password-prompt-issue.md
+++ b/.changeset/clerk-cli-macos-keychain-password-prompt-issue.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Prevent local unsigned macOS builds from sharing the release keychain entry.
--- a/packages/cli-core/src/lib/credential-store-signing.test.ts
+++ b/packages/cli-core/src/lib/credential-store-signing.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, test } from "bun:test";
+import { isReleaseSignedMacosBinary } from "./credential-store.ts";
+
+describe("isReleaseSignedMacosBinary", () => {
+  test("returns false for unversioned dev builds", () => {
+    const output = `
+Executable=/tmp/clerk
+Identifier=clerk
+TeamIdentifier=L8SD6SB282
+`;
+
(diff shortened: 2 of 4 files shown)


rafa-thayto (approved):
LGTM

PR #199, opened 2026-04-21
feat(release): prepare clerk 1.0.0 release

wyattjoh:
## Summary

- add a major changeset so the next version-packages PR resolves to `clerk@1.0.0`
- remove the stray release-process test changeset and add version-specific `v1.0.0` GitHub release intro notes
- update the stable releaser to prepend checked-in release intro text ahead of GitHub autogenerated notes
- document the release-note source in `docs/releasing.md` and remove the stale open-questions footer from `README.md`
- the currently open version-packages PR still targets `0.9.0`; after this branch lands, Changesets should update/regenerate that PR to `1.0.0`
- stable GitHub release notes still use `--generate-notes`; this change only prepends `.github/release-notes/vX.Y.Z.md` when present

## Test plan

- [x] `bun test scripts/lib/release-notes.test.ts`
- [x] `bunx tsc --noEmit -p scripts/tsconfig.json`
- [x] `bun x changeset status --verbose`
- [ ] Confirm the next version-packages PR regenerates to `clerk@1.0.0`

PR #202, opened 2026-04-21
fix(release): install.sh --install-dir and release docs

wyattjoh:
## Summary
- `install.sh` now creates the `--install-dir` target when it does not already exist, so `./install.sh --install-dir /new/path` succeeds on a fresh directory. Previously `mkdir -p` only ran inside the `~/.local/bin` fallback branch, and the final `mv` failed on any user-supplied path that did not exist.
- `docs/releasing.md` stops pointing at the removed `scripts/releaser/` directory (moved to `scripts/releaser.ts` and `scripts/lib/targets.ts`), documents that the releaser expects the CI-style `$ARTIFACTS_DIR/clerk-<target>/` layout (distinct from the `dist/artifacts/<target>/` layout written by `scripts/build.ts`), and notes the Bun 1.3.11 pin used by CI so local darwin-arm64 rehearsal matches production.

## Context
Follow-up to #199. Covers the remaining release-readiness gaps surfaced during review of the 1.0.0 prep branch. Stacked on top of #199 so review stays scoped. Changelog entry added for the `install.sh` fix; docs change is non-user-facing and needs no changeset.

## Test plan
- [x] Reproduced the original `install.sh --install-dir /tmp/new-path` failure, confirmed the fix resolves it.
- [x] Dry-ran the new rehearsal recipe in `docs/releasing.md` against a staged `clerk-darwin-arm64/` artifact.
- [x] `bun run format:check`, `lint`, `typecheck`, `test` all pass locally.

### DIFF
--- a/.changeset/install-sh-create-install-dir.md
+++ b/.changeset/install-sh-create-install-dir.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Fix `install.sh --install-dir <path>` so it creates the directory when it does not already exist, matching the behavior of the `~/.local/bin` fallback.
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -29,7 +29,7 @@ PR comment "!snapshot [name]"
 
 ## Architecture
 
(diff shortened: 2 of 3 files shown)


wyattjoh:
**Stack: wyattjoh/release-clerk-1-0-0**

- ~~#199~~
  - **#202 👈 this PR**

rafa-thayto (approved):
LGTM

PR #203, opened 2026-04-22
Refactor credential file storage to use writeFile with mode

dominic-clerk:
## Summary
Refactored the credential file storage mechanism to use Node.js `writeFile` API with atomic mode setting instead of separate `chmod` call, and improved directory creation with explicit mode.

## Key Changes
- Replaced `Bun.write()` with `writeFile()` from `node:fs/promises` for better Node.js compatibility
- Combined file permission setting into the `writeFile` call using the `mode` option (0o600) instead of separate `chmod` call
- Updated `mkdir` call to explicitly set directory mode to 0o700 for better security
- Updated imports to include `writeFile` and remove `chmod`

## Implementation Details
- The credentials file is now created with restrictive permissions (0o600) atomically during write
- The credentials directory is created with 0o700 permissions (rwx for owner only)
- This approach is more efficient as it eliminates the separate `chmod` system call and ensures permissions are set correctly from file creation

https://claude.ai/code/session_014o99K9iccehEG6uV8fuH96

### DIFF
--- a/packages/cli-core/src/lib/credential-store.ts
+++ b/packages/cli-core/src/lib/credential-store.ts
@@ -8,7 +8,7 @@
  */
 
 import { dirname } from "node:path";
-import { mkdir, chmod, unlink } from "node:fs/promises";
+import { mkdir, writeFile, unlink } from "node:fs/promises";
 import { CREDENTIALS_FILE } from "./constants.ts";
 import { getCurrentEnvName } from "./environment.ts";
 import { log } from "./log.ts";
@@ -142,9 +142,8 @@ async function keyringDelete(): Promise<boolean> {
 async function fileStore(token: string): Promise<void> {
   const path = credentialsFile();
   log.debug(`credentials: storing token in file ${path}`);
-  await mkdir(dirname(path), { recursive: true });
-  await Bun.write(path, token);
-  await chmod(path, 0o600);
+  await mkdir(dirname(path), { recursive: true, mode: 0o700 });
+  await writeFile(path, token, { mode: 0o600 });
 }
 
 async function fileGet(): Promise<string | null> {
--- a/packages/cli-core/src/lib/credential-store.ts
+++ b/packages/cli-core/src/lib/credential-store.ts
@@ -144,6 +144,7 @@ async function fileStore(token: string): Promise<void> {
   log.debug(`credentials: storing token in file ${path}`);
   await mkdir(dirname(path), { recursive: true, mode: 0o700 });
   await writeFile(path, token, { mode: 0o600 });
+  await chmod(path, 0o600);
 }
 
 async function fileGet(): Promise<string | null> {
--- a/packages/cli-core/src/lib/credential-store.ts
+++ b/packages/cli-core/src/lib/credential-store.ts
@@ -8,7 +8,7 @@
  */
 
 import { dirname } from "node:path";
-import { mkdir, writeFile, unlink } from "node:fs/promises";
+import { mkdir, chmod, writeFile, unlink } from "node:fs/promises";
 import { CREDENTIALS_FILE } from "./constants.ts";
 import { getCurrentEnvName } from "./environment.ts";
 import { log } from "./log.ts";
--- a/packages/cli-core/src/lib/credential-store.ts
+++ b/packages/cli-core/src/lib/credential-store.ts
@@ -144,6 +144,8 @@ async function fileStore(token: string): Promise<void> {
   log.debug(`credentials: storing token in file ${path}`);
   await mkdir(dirname(path), { recursive: true, mode: 0o700 });
   await writeFile(path, token, { mode: 0o600 });
+  // We keep the chmod because if the file permission had changed
+  // `writeFile` wouldn't set it back to 0o600
   await chmod(path, 0o600);
 }
 


dominic-clerk:
See https://clerkinc.slack.com/archives/C0AR86TQS9G/p1776783702547329

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -142,9 +142,8 @@ async function keyringDelete(): Promise<boolean> {
 async function fileStore(token: string): Promise<void> {
   const path = credentialsFile();
   log.debug(`credentials: storing token in file ${path}`);
-  await mkdir(dirname(path), { recursive: true });
-  await Bun.write(path, token);
-  await chmod(path, 0o600);
+  await mkdir(dirname(path), { recursive: true, mode: 0o700 });
+  await writeFile(path, token, { mode: 0o600 });
```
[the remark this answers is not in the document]
dominic-clerk: @wyattjoh can we do this without breaking anything?

The CLI is new and we did have the `chmod` call in place so I'm not too worried about existing installations, but it's a decent hardening suggestion by coderabbit.

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -142,9 +142,8 @@ async function keyringDelete(): Promise<boolean> {
```
rafa-thayto, line 141: +1 to the unlink-before-write approach CodeRabbit suggested. Since `writeFile`'s `mode` only applies on creation, an existing file keeps its old permissions. `unlink` is already imported so this is a one-liner:

```suggestion
  await mkdir(dirname(path), { recursive: true, mode: 0o700 });
  await unlink(path).catch(() => {});
  await writeFile(path, token, { mode: 0o600 });
```

packages/cli-core/src/lib/credential-store.ts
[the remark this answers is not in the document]
wyattjoh: I think we could be a little defensive here!

```suggestion
  await writeFile(path, token, { mode: 0o600 });
  await chmod(path, 0o600);
```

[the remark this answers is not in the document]
dominic-clerk: I like it, we don't delete the file if it already existed and had modifications but we reset the permission in case it changed

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -142,8 +142,8 @@ async function keyringDelete(): Promise<boolean> {
 async function fileStore(token: string): Promise<void> {
   const path = credentialsFile();
   log.debug(`credentials: storing token in file ${path}`);
-  await mkdir(dirname(path), { recursive: true });
-  await Bun.write(path, token);
+  await mkdir(dirname(path), { recursive: true, mode: 0o700 });
+  await writeFile(path, token, { mode: 0o600 });
   await chmod(path, 0o600);
```
dominic-clerk, line 149: Adding a comment because this looks redundant at first.

```suggestion
  // We keep the chmod because if the file permission had changed
  // `writeFile` wouldn't set it back to 0o600
  await chmod(path, 0o600);
```

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -8,7 +8,7 @@
  */
 
 import { dirname } from "node:path";
-import { mkdir, chmod, unlink } from "node:fs/promises";
+import { mkdir, writeFile, unlink } from "node:fs/promises";
```
dominic-clerk: Missing `chmod` here
```suggestion
import { mkdir, chmod, writeFile, unlink } from "node:fs/promises";
```

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -142,8 +142,10 @@ async function keyringDelete(): Promise<boolean> {
 async function fileStore(token: string): Promise<void> {
   const path = credentialsFile();
   log.debug(`credentials: storing token in file ${path}`);
-  await mkdir(dirname(path), { recursive: true });
-  await Bun.write(path, token);
+  await mkdir(dirname(path), { recursive: true, mode: 0o700 });
+  await writeFile(path, token, { mode: 0o600 });
+  // We keep the chmod because if the file permission had changed
+  // `writeFile` wouldn't set it back to 0o600
   await chmod(path, 0o600);
```
wyattjoh, line 149: `fileStore()` still calls `chmod(path, 0o600)`, but this PR removes `chmod` from the `node:fs/promises` import. I checked the PR head in a detached worktree and `tsc -p packages/cli-core/tsconfig.json --noEmit` fails with `TS2304: Cannot find name 'chmod'` at this line. Re-add `chmod` to the import, or remove this call only if you also preserve the existing-file permission fix another way.

wyattjoh (commented):
This change is headed in the right direction, but the current PR head has a blocking correctness issue.

packages/cli-core/src/lib/credential-store.ts
dominic-clerk: Fixed in 32c73e88617976462f39bba4c03e1a13b693f4be

wyattjoh (approved):

PR #206, opened 2026-04-22
fix(completion): make install tips work on fresh systems

rafa-thayto:
## Summary

- `clerk doctor` showed a shell-completion remedy that blew up on a fresh system. The zsh tip was `clerk completion zsh > ~/.zfunc/_clerk`, but `~/.zfunc` doesn't exist by default, and even with `mkdir -p` the user still needed `fpath=(~/.zfunc $fpath)` + `autoload -Uz compinit` in their `~/.zshrc` — which the remedy didn't mention. Lead with the zero-config `eval "$(clerk completion zsh)"` form and defer the file-install method to `clerk completion --help`.
- Fix the fish remedy to `mkdir -p ~/.config/fish/completions` before redirecting, since that directory only exists if fish has been run at least once.
- Update the banner inside the generated zsh completion script to include `mkdir -p ~/.zfunc` before the redirect (so users debugging `~/.zfunc/_clerk` see the complete recipe).
- Bash was already fine (eval form, no filesystem touching).

### Before

```
! Shell completion not installed for zsh
    Add `eval "$(clerk completion zsh)"` to your ~/.zshrc, or run `clerk completion zsh > ~/.zfunc/_clerk`

$ clerk completion zsh > ~/.zfunc/_clerk
zsh: no such file or directory: /Users/thayto/.zfunc/_clerk
```

### After

```
! Shell completion not installed for zsh
    Add `eval "$(clerk completion zsh)"` to your ~/.zshrc (run `clerk completion --help` for other install methods)
```

## Test plan

- [ ] `bun run test` — passes (added regression tests for zsh/fish remedy wording and for the zsh banner `mkdir -p` ordering)
- [ ] `SHELL=/bin/zsh HOME=/tmp/empty bun run src/cli.ts doctor` — zsh remedy now suggests the eval form
- [ ] `SHELL=/usr/bin/fish HOME=/tmp/empty bun run src/cli.ts doctor` — fish remedy includes `mkdir -p`
- [ ] `bun run src/cli.ts completion zsh | head -20` — banner comment tells users to `mkdir -p ~/.zfunc` before writing

### DIFF
--- a/packages/cli-core/src/commands/doctor/checks.ts
+++ b/packages/cli-core/src/commands/doctor/checks.ts
@@ -369,7 +369,8 @@ const SHELL_COMPLETION: Record<
 > = {
   fish: {
     isInstalled: (home) => Bun.file(join(home, ".config/fish/completions/clerk.fish")).exists(),
-    remedy: "Run `clerk completion fish > ~/.config/fish/completions/clerk.fish`",
+    remedy:
+      "Run `mkdir -p ~/.config/fish/completions && clerk completion fish > ~/.config/fish/completions/clerk.fish`",
   },
   bash: {
     isInstalled: (home) =>
@@ -381,7 +382,7 @@ const SHELL_COMPLETION: Record<
       (await fileContains([join(home, ".zshrc")], "clerk completion")) ||
       (await Bun.file(join(home, ".zfunc/_clerk")).exists()),
     remedy:
-      'Add `eval "$(clerk completion zsh)"` to your ~/.zshrc, or run `clerk completion zsh > ~/.zfunc/_clerk`',
+      'Add `eval "$(clerk completion zsh)"` to your ~/.zshrc (run `clerk completion --help` for other install methods)',
   },
 };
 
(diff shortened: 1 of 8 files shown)


packages/cli-core/src/commands/doctor/checks.ts
```diff
@@ -381,7 +382,7 @@ const SHELL_COMPLETION: Record<
       (await fileContains([join(home, ".zshrc")], "clerk completion")) ||
       (await Bun.file(join(home, ".zfunc/_clerk")).exists()),
     remedy:
```
wyattjoh, line 384: This updates the `doctor` copy, but the alternative-installation path it points people to is still incomplete: `clerk completion --help` and `packages/cli-core/src/commands/completion/README.md` still show `clerk completion fish > ~/.config/fish/completions/clerk.fish` without first creating `~/.config/fish/completions`. On a fresh system that command still fails with `No such file or directory`, so users who follow the help output instead of `doctor` will hit the same problem this PR is trying to fix. Please update the shared help/docs examples to use the mkdir-prefixed fish command too.

wyattjoh (commented):
The patch is directionally correct, but one user-facing install path is still left broken on fresh fish setups.

wyattjoh (approved):

PR #212, opened 2026-04-22
fix(cli-core): support agent-mode link and unlink

wyattjoh:
## Summary

`clerk link` and `clerk unlink` previously hard-returned in agent mode, even when the caller had already provided the flags needed to make the operation deterministic. That kept non-interactive flows like `init`, onboarding, and app switching from completing end-to-end under an agent harness, and the bundled `skills/clerk` docs had to document the limitation instead of the intended workflow. This change removes the blanket agent-mode block, preserves the safety checks that actually matter, and updates the skill/docs to match the new behavior.

## What changed

- **`link` now supports deterministic agent-mode flows.** In agent mode, `clerk link --app <id>` links directly, and the no-flag path first tries silent autolink from detected keys. If the CLI still cannot determine the target app without a picker, it exits with a usage error telling the caller to pass `--app` or inspect `clerk apps list --json`.
- **`unlink` now works in agent mode with explicit confirmation.** The unconditional agent-mode early return is gone. `clerk unlink --yes` removes the link non-interactively, while omitting `--yes` now fails with a usage error instead of printing guidance.
- **The integration coverage now exercises the agent flows that were previously skipped.** Agent-mode success/error cases were added for `link` and `unlink`, and the previously disabled agent variants of the onboard and switch-app flows are now enabled so the non-interactive path is covered end-to-end.
- **Bundled docs and the tracked Clerk skill were updated to match runtime behavior.** The command READMEs and `skills/clerk/*` now describe deterministic agent linking, `unlink --yes`, and the current agent-mode error semantics instead of documenting the old limitation.

## Test plan

- [x] `bun test packages/cli-core/src/commands/link/index.test.ts packages/cli-core/src/commands/unlink/index.test.ts packages/cli-core/src/test/integration/agent-mode.test.ts packages/cli-core/src/test/integration/onboard.test.ts packages/cli-core/src/test/integration/switch-apps.test.ts`
- [x] `bun run --filter @clerk/cli-core build`
- [x] Built-JS smoke checks for `--version`, `doctor --json`, `completion zsh`, `init --prompt`, `api ls users`, and the new agent-mode `link` / `unlink` usage errors
- [ ] Authenticated manual smoke test against a live Clerk app from this agent environment


packages/cli-core/src/commands/link/index.ts
```diff
@@ -72,18 +55,32 @@ export async function link(options: LinkOptions = {}): Promise<void> {
...
+        return;
+      }
+    } else {
+      const shouldRelink = await handleExistingProfile(existing, normalizedRemote, options);
+      if (!shouldRelink) {
+        outro();
+        return;
+      }
     }
   }
```
rafa-thayto: A couple of small readability tweaks here:

1. `targetsDifferentApp` already requires `options.app` to be truthy, so `!options.app || !targetsDifferentApp` reduces to just `!targetsDifferentApp`.
2. The nested `if (agent) / else` can flatten with `else if (existing)` so both existing-profile paths sit at the same level.

```suggestion
  if (existing && agent) {
    printExistingStatus(existing, normalizedRemote);
    if (!targetsDifferentApp) {
      outro();
      return;
    }
  } else if (existing) {
    const shouldRelink = await handleExistingProfile(existing, normalizedRemote, options);
    if (!shouldRelink) {
      outro();
      return;
    }
  }
```

rafa-thayto (approved):

PR #213, opened 2026-04-22
fix(auth): recover from concurrent session refreshes

wyattjoh:
## Summary

Follow-up to #205. When two separate `clerk` processes start from the same expired OAuth session, they can both race into `refreshStoredSession()`. If one process refreshes successfully and rotates the refresh token, the loser sees `invalid_grant` and, in the current implementation, deletes the credentials that the winner just rewrote.

This PR keeps the existing refresh flow but hardens the `invalid_grant` path so a losing process re-reads the credential store before treating the session as dead. If another process already stored a newer session, the loser reuses that session instead of deleting credentials and forcing a re-auth.

## Fix

- Add a small recovery path for `invalid_grant` in `credential-store.ts`.
- Capture the failing session's refresh token, then re-read the store with a short retry window (`25ms`, `50ms`, `100ms`) before deleting anything. A different refresh token implies a sibling process already rotated the session.
- If a newer session appears during that window, treat it as the winner of the race and continue with its access token instead of clearing auth state.
- Keep the existing cleanup behavior for the real expired-session case where the store never changes.
- Add regression coverage for both paths and a patch changeset for the follow-up fix.

## Test plan

- [x] `bunx tsc -p packages/cli-core/tsconfig.json --noEmit`
- [x] `bun test packages/cli-core/src/lib/plapi.test.ts packages/cli-core/src/lib/token-exchange.test.ts packages/cli-core/src/commands/auth/login.test.ts packages/cli-core/src/commands/doctor/doctor.test.ts packages/cli-core/src/commands/whoami/index.test.ts`
- [x] Added a regression test that simulates another process winning the refresh race and verifies the loser reuses the newer stored session.

### DIFF
--- a/packages/cli-core/src/lib/credential-store.test.ts
+++ b/packages/cli-core/src/lib/credential-store.test.ts
@@ -112,6 +112,32 @@ describe("credential-store", () => {
     });
   });
 
+  test("getValidToken reuses a newer stored session after invalid_grant from another refresh", async () => {
+    const session = {
+      accessToken: "expired-access-token",
+      refreshToken: "refresh-token",
+      expiresAt: Date.now() - 60_000,
+      tokenType: "Bearer",
+    };
+    const refreshedSession = {
+      accessToken: "other-process-access-token",
+      refreshToken: "other-process-refresh-token",
+      expiresAt: Date.now() + 60_000,
+      tokenType: "Bearer",
+    };
+    await storeToken(session);
+
+    mockRefreshAccessToken.mockImplementation(async () => {
+      setTimeout(() => {
+        void storeToken(refreshedSession);
+      }, 5);
+      throw new ApiError(400, "invalid_grant");
+    });
+
+    expect(await getValidToken()).toBe("other-process-access-token");
+    expect(await getStoredSession()).toEqual(refreshedSession);
+  });
+
   test("getValidToken deletes stored credentials when refresh returns invalid_grant", async () => {
     const session = {
       accessToken: "expired-access-token",
--- a/packages/cli-core/src/lib/credential-store.test.ts
+++ b/packages/cli-core/src/lib/credential-store.test.ts
@@ -112,7 +112,7 @@ describe("credential-store", () => {
     });
   });
 
-  test("getValidToken reuses a newer stored session after invalid_grant from another refresh", async () => {
+  test("getValidToken recovers from a concurrent refresh race when another process completes the refresh first (invalid_grant)", async () => {
     const session = {
       accessToken: "expired-access-token",
       refreshToken: "refresh-token",
--- a/packages/cli-core/src/lib/credential-store.ts
+++ b/packages/cli-core/src/lib/credential-store.ts
@@ -294,15 +294,6 @@ function sessionExpiredError(): AuthError {
   return new AuthError({ reason: "session_expired" });
 }
 
-function sessionFingerprint(session: OAuthSession): string {
-  return JSON.stringify([
-    session.accessToken,
-    session.refreshToken,
-    session.expiresAt,
-    session.tokenType,
-  ]);
-}
-
 function isInvalidGrant(error: unknown): boolean {
   return (
     error instanceof ApiError &&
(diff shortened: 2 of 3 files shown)


rafa-thayto (commented):
LGTM with two suggestions below. The approach is solid and well-scoped. The fingerprint + retry strategy is a clean way to handle this race.

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -309,13 +267,45 @@ async function readStoredValue(): Promise<string | null> {
...
+  if (!isExpiredSession(session)) {
+    return session.accessToken;
+  }
+
+  return refreshStoredSession(session);
+}
+
+async function recoverFromInvalidGrant(session: OAuthSession): Promise<string | null> {
+  const originalFingerprint = sessionFingerprint(session);
+
```
rafa-thayto: `recoverFromInvalidGrant` calls `getValidAccessToken`, which can call `refreshStoredSession`, which on `invalid_grant` calls `recoverFromInvalidGrant` again. If the recovered session happens to be expired too (clock skew, very short-lived token), this recurses without a depth guard.

In practice unlikely, but easy to prevent by not re-entering the refresh pipeline from recovery:

```suggestion
    log.debug("credentials: detected a newer stored session after invalid_grant");
    if (isExpiredSession(storedSession)) {
      continue;
    }
    return storedSession.accessToken;
```

This keeps recovery leaf-level. If the winner's session is already expired, the loop continues and eventually falls through to the existing delete + session_expired path.

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -309,13 +267,45 @@ async function readStoredValue(): Promise<string | null> {
...
+}
+
 async function refreshStoredSession(session: OAuthSession): Promise<string> {
   let tokenResponse: TokenResponse;
   try {
     log.debug("credentials: refreshing OAuth session");
     tokenResponse = await refreshAccessToken(session.refreshToken);
   } catch (error) {
     if (isInvalidGrant(error)) {
+      const recoveredToken = await recoverFromInvalidGrant(session);
```
rafa-thayto: If `recoverFromInvalidGrant` throws (e.g. the nested `getValidAccessToken` hits a network error), the exception escapes the catch block and skips both `deleteToken()` and the `sessionExpiredError()` below. The stale credentials stay on disk and the user gets a raw error instead of the friendly re-auth prompt.

```suggestion
      try {
        const recoveredToken = await recoverFromInvalidGrant(session);
        if (recoveredToken) {
          return recoveredToken;
        }
      } catch {
        log.debug("credentials: recovery from invalid_grant failed, cleaning up");
      }
```

This way any recovery failure falls through to the existing delete + session_expired path.

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -27,6 +28,7 @@ export const KEYCHAIN_ACCOUNT = "oauth-access-token";
 const RELEASE_MACOS_TEAM_ID = "L8SD6SB282";
 const RELEASE_MACOS_IDENTIFIER = "clerk";
 const JWT_EXPIRY_LEEWAY_MS = 30_000;
+const INVALID_GRANT_RETRY_DELAYS_MS = [25, 50, 100];
```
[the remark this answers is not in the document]
wyattjoh: The window exists but the suggested diff isn't the fix. Dropping `deleteToken()` entirely (as shown in the collapsed diff) leaves genuinely-stale credentials on disk forever: on real session expiry (not a race) every subsequent call re-enters `recoverFromInvalidGrant`, finds no newer session, throws `session_expired`, and the bad creds persist. The user never gets cleared state.

The correct shape is the prompt-variant: gate `deleteToken()` on a fingerprint re-check so we only wipe when the stored session still matches what we failed to refresh. That narrows (though doesn't fully close) the TOCTOU. Holding that for a follow-up rather than expanding this PR's scope. The current 175 ms budget covers the common race, and worst-case tail is a forced re-auth, not loss of persisted state other than the rotated-out session.

Am I missing a tighter failure mode?

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -292,6 +294,15 @@ function sessionExpiredError(): AuthError {
   return new AuthError({ reason: "session_expired" });
 }
 
+function sessionFingerprint(session: OAuthSession): string {
```
jfoshee: I think it's sufficient to just use the refresh token as the fingerprint. It will be unique.

packages/cli-core/src/lib/credential-store.test.ts
```diff
@@ -112,6 +112,32 @@ describe("credential-store", () => {
     });
   });
 
+  test("getValidToken reuses a newer stored session after invalid_grant from another refresh", async () => {
```
jfoshee: The test description could be a little clearer. At least we should explicitly mention that this is a concurrent refresh / race-condition scenario.

packages/cli-core/src/lib/credential-store.ts
```diff
@@ -309,13 +320,52 @@ async function readStoredValue(): Promise<string | null> {
...
 
+async function getValidAccessToken(session: OAuthSession): Promise<string> {
+  if (!isExpiredSession(session)) {
+    return session.accessToken;
+  }
+
+  return refreshStoredSession(session);
+}
+
+async function recoverFromInvalidGrant(session: OAuthSession): Promise<string | null> {
```
jfoshee: This function and/or the caller needs some documentation about why/what we are doing here. Remember the reader won't have the benefit of the PR description. 

I don't love the name `recoverFromInvalidGrant` because normally that's not recoverable. It's really something more like, `checkForConcurrentSuccessfulRefresh`... But maybe you can think of something less verbose.  Naming this stuff is not easy. ;-)

packages/cli-core/src/lib/credential-store.ts
wyattjoh: Done. Simplified to compare `refreshToken` directly and dropped the `sessionFingerprint` helper. Added a one-liner in the JSDoc noting that detection relies on the OAuth server rotating refresh tokens on every successful exchange (Clerk does, and rotation is the safe assumption here).

packages/cli-core/src/lib/credential-store.test.ts
wyattjoh: Renamed to make the race-condition framing explicit: `getValidToken recovers from a concurrent refresh race when another process completes the refresh first (invalid_grant)`.

packages/cli-core/src/lib/credential-store.ts
wyattjoh: Renamed to `awaitConcurrentRefresh` and added a JSDoc covering the race window, the polling budget, and why we compare refresh tokens. Skipped `checkForConcurrentSuccessfulRefresh` only because the verb is the load-bearing signal here. The function blocks/polls, so `await` reads more accurately than `check`. Happy to switch if you prefer the longer form.

rafa-thayto (approved):

jfoshee (approved):

PR #214, opened 2026-04-22
fix(release): await homebrew publish commands

wyattjoh:
## Summary

The stable release workflow's `homebrew` job did run in GitHub Actions run `#327`, but `scripts/homebrew.ts` was firing its async shell commands without awaiting them. That let the script race past `gh release upload`, `git clone`, and the later git operations, which made the job fail while publishing the Homebrew tap.

## Fix

- await each `run(...)` call in `scripts/homebrew.ts`
- serialize the GitHub Release uploads before computing checksums and updating the tap repo
- ensure the tap clone, git config, add, commit, and push steps finish before the script advances

## Test plan

- [x] `bun test scripts/lib/homebrew.test.ts`
- [x] `bunx tsc --noEmit -p scripts/tsconfig.json`
- [x] Verified the failing Actions log for run `24797651889` / job `72572381605` and confirmed the crash path was inside `scripts/homebrew.ts`

### DIFF
--- a/.changeset/homebrew-await-commands.md
+++ b/.changeset/homebrew-await-commands.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Fix the stable release Homebrew publish step so it awaits each release upload and tap repository command before moving to the next step. This prevents the Homebrew workflow from racing past `gh release upload`, `git clone`, and the follow-up git operations while publishing a release.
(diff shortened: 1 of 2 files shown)


rafa-thayto (approved):

jfoshee (approved):
🕐

PR #219, opened 2026-04-23
feat(cli-core): add clerk enable and clerk disable commands for orgs and billing

nicolas-angelo:
## Summary

This PR adds two top-level commands — `clerk enable` and `clerk disable` — for toggling Clerk features on the linked instance. The toggle layer is intentionally narrow: simple on/off plus a few discoverable convenience flags. Anything beyond that continues to flow through `clerk config patch`.

## What changed

- **`clerk enable orgs` / `clerk disable orgs`**: toggle organizations. `enable` accepts `--force-selection`, `--auto-create`, `--max-members <n>`, and `--domains` for the most common settings.
- **`clerk enable billing` / `clerk disable billing`**: toggle billing for organizations and/or users. `--for` defaults to both targets when omitted; enabling for `org` cascades to enabling organizations.
- **Agent skill nudge**: after a successful `enable billing`, offers to install the `clerk-billing` skill from `clerk/skills` (suppress with `--no-skills`). `clerk init` doesn't bundle this one as a default — billing is opt-in — so `enable billing` is the natural moment to surface it.
- **Next-steps hints**: after each enable, prints pointers to `clerk config schema --keys <section>` (what's tunable) and `clerk config pull --keys <section>` (what's currently set) for deeper exploration. Skipped in agent mode.

### Organizations

```sh
clerk enable orgs                              # bare toggle
clerk enable orgs --force-selection            # toggle + force selection
clerk enable orgs --max-members 10 --domains   # toggle + multiple flags
clerk enable orgs --dry-run                    # preview the patch
clerk disable orgs                             # turn off
```

### Billing

```sh
clerk enable billing                           # both targets (default)
clerk enable billing --for org                 # org only (cascades to enabling orgs)
clerk enable billing --for user                # user only
clerk enable billing --for org user            # variadic form
clerk enable billing --for org,user            # CSV form
clerk disable billing --for org                # org billing off; orgs themselves untouched
clerk enable billing --no-skills               # skip the clerk-billing skill install
```

## Before / After

**Before** (manual JSON):

```sh
clerk config patch --json '{"organization_settings":{"enabled":true,"force_organization_selection":true}}' --yes
clerk config patch --json '{"billing":{"organization_enabled":true},"organization_settings":{"enabled":true}}' --yes
```

**After** (convenience toggles):

```sh
clerk enable orgs --force-selection
clerk enable billing --for org
```


packages/cli-core/src/cli-program.ts
```diff
@@ -414,6 +423,175 @@ Give AI agents better Clerk context: install the Clerk skills
...
+  orgs
+    .command("enable")
+    .description("Enable organizations on the linked instance")
+    .option("--app <id>", "Application ID to target")
+    .option("--instance <id>", "Instance to target (dev, prod, or instance ID)")
+    .option("--force-selection", "Force organization selection on login")
+    .option("--auto-create", "Auto-create an organization for new users")
+    .option("--max-members <n>", "Maximum members per organization")
+    .option("--domains", "Enable verified domains")
+    .option("--yes", "Skip confirmation prompts")
```
wyattjoh, line 643: `--yes` is advertised on `orgs enable/disable`, `billing enable/disable`, and `plans create/update/remove`, but no implementation consults `options.yes` or calls `confirm()`. The commands always mutate silently. Users who see `--yes` will reasonably assume there is a prompt to skip.

Compare with `config patch`, which these shortcuts wrap: `packages/cli-core/src/commands/config/push.ts:94-102` does `if (!options.dryRun && isHuman() && !options.yes) { await confirm(...) }`. Without an equivalent branch here, the shortcut is strictly more dangerous than the command it's supposed to replace.

Fix: either remove `--yes` and the `yes?: boolean` fields until a prompt exists, or add the real confirmation branch for every mutating command. Option two is the right call, these are production config edits.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,270 @@
...
+
+  delete plans[slug];
+
+  const config = { billing: { plans } };
+
+  const result = await withSpinner(
+    `Removing plan ${cyan(slug)} on ${ctx.appLabel} (${ctx.instanceLabel})...`,
+    () =>
+      withApiContext(
+        patchInstanceConfig(ctx.appId, ctx.instanceId, config, { destructive: true }),
```
wyattjoh: `plansRemove` fetches the current config, deletes the slug in memory, and PATCHes with `{ destructive: true }`. No prompt, no diff, no dry-run. A typo in `<slug>` or the active instance silently wipes a live subscription plan.

The equivalent `clerk config patch` path would print a diff, warn, and ask "Proceed?". The shortcut regresses this safety.

Fix: gate the mutation behind `isHuman() && !options.yes` confirmation, support `--dry-run`, and show the plan (name, price, payer) being removed before confirming.

packages/cli-core/src/cli-program.ts
```diff
@@ -414,6 +423,175 @@ Give AI agents better Clerk context: install the Clerk skills
...
+      },
+    ]);
+
+  plans
+    .command("create")
+    .description("Create a subscription plan")
+    .argument("<slug>", "Plan slug (display name auto-derived via title case)")
+    .option("--name <name>", "Override display name")
+    .requiredOption("--amount <cents>", "Monthly price in cents")
+    .addOption(createOption("--payer <type>", "Who pays").choices(["org", "user"]))
```
wyattjoh: Line 527 uses `.requiredOption("--amount <cents>", …)`. Line 528 uses `.addOption(createOption("--payer <type>", "Who pays").choices(["org", "user"]))` with no `.makeOptionMandatory()`. `packages/cli-core/src/commands/billing/README.md:18` claims `--payer` is **(required)**.

When omitted, `options.payer === undefined`, so `billing/index.ts:110` writes `payer_type: undefined`. `JSON.stringify` drops the field, sending a malformed plan to the API.

Fix:
```ts
.addOption(
  createOption("--payer <type>", "Who pays")
    .choices(["org", "user"])
    .makeOptionMandatory(),
)
```
Add a test covering the missing-`--payer` rejection.

packages/cli-core/src/cli-program.ts
```diff
@@ -414,6 +423,175 @@ Give AI agents better Clerk context: install the Clerk skills
...
+        description: "Create a Pro plan at $19.99/mo",
+      },
+    ]);
+
+  plans
+    .command("create")
+    .description("Create a subscription plan")
+    .argument("<slug>", "Plan slug (display name auto-derived via title case)")
+    .option("--name <name>", "Override display name")
+    .requiredOption("--amount <cents>", "Monthly price in cents")
```
wyattjoh: `patchInstanceConfig()` accepts `dryRun` (see `lib/plapi.ts:156-184`). `clerk config patch` exposes `--dry-run` at `cli-program.ts:367`. None of the new shortcuts plumb it through.

A maintainer who wants to preview `billing enable --for org` or `plans remove pro` on production has to fall back to the raw `config patch` command, defeating the point of the wrappers. These are the exact operations you most want to dry-run.

Fix: add `--dry-run` to every mutating subcommand and pass `{ dryRun: options.dryRun }` into `patchInstanceConfig`, matching `config/push.ts:107-115`.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,270 @@
...
+    .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
+    .join(" ");
+}
+
+export async function plansCreate(slug: string, options: PlansCreateOptions): Promise<void> {
+  const ctx = await resolveAppContext(options);
+
+  const plan: Record<string, unknown> = {
+    name: options.name || titleCase(slug),
+    amount: parseInt(options.amount, 10),
```
wyattjoh: Every numeric option (`--amount`, `--max-members`, `--trial-days`, `--annual-amount`) goes through raw `parseInt(value, 10)` with no validation, here and at billing/index.ts:117, 120, 201, 204, 209, plus orgs/index.ts:23.

`parseInt("abc", 10)` returns `NaN`, which `JSON.stringify` emits as `null`. So `clerk billing plans create pro --amount abc --payer org` sends `amount: null` to the API, and `--max-members abc` sends `max_allowed_memberships: null`. Neither failure mode is user-friendly.

Fix: use Commander's option argParser at the registration site:
```ts
.option("--amount <cents>", "Monthly price in cents", (v) => {
  const n = Number.parseInt(v, 10);
  if (!Number.isFinite(n) || n < 0) {
    throwUsageError(`Invalid --amount: "${v}". Must be a non-negative integer.`);
  }
  return n;
})
```
Or extract a shared `parseNonNegativeInt` helper.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,270 @@
...
+  const ctx = await resolveAppContext(options);
+
+  const plan: Record<string, unknown> = {};
+  if (options.name) plan.name = options.name;
+  if (options.amount) plan.amount = parseInt(options.amount, 10);
+  if (options.currency) plan.currency = options.currency;
+  if (options.description) plan.description = options.description;
+  if (options.annualAmount) plan.annual_monthly_amount = parseInt(options.annualAmount, 10);
+  if (options.hidden) plan.publicly_visible = false;
+  if (options.visible) plan.publicly_visible = true;
```
wyattjoh:
```ts
if (options.hidden) plan.publicly_visible = false;
if (options.visible) plan.publicly_visible = true;
```
If both flags are passed (scripting mistake, alias confusion), the second write wins with no warning. There's no "last flag wins" convention on this CLI and the user's intent is genuinely ambiguous.

Fix: reject up front.
```ts
if (options.hidden && options.visible) {
  throwUsageError("--hidden and --visible are mutually exclusive.");
}
```

packages/cli-core/src/commands/orgs/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  const current = await withSpinner("Checking current config...", () =>
+    withApiContext(
+      fetchInstanceConfig(ctx.appId, ctx.instanceId, ["billing"]),
+      "Failed to fetch config",
+    ),
+  );
+
+  const billing = current.billing as Record<string, unknown> | undefined;
+  if (billing?.organization_enabled) {
+    log.warn(
```
wyattjoh: `orgsDisable` fetches billing config and calls `log.warn("…Disabling organizations will also disable org billing.")`, then unconditionally issues the PATCH.

Emitting a warning and then doing the thing the warning describes is worse than not warning at all, in a CI log it looks like the warning was heeded. If the dependency is worth detecting (it is), it's worth gating behind confirmation.

Fix: when `billing.organization_enabled` is true, require explicit confirmation in human mode or a `--force` / `--yes` flag to proceed. In non-human mode, consider failing with a usage error directing the user to pass `--force`.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,270 @@
...
+  } else {
+    patch.user_enabled = true;
+  }
+  if (options.requirePaymentMethod !== undefined) {
+    patch.free_trial_requires_payment_method = options.requirePaymentMethod;
+  }
+
+  const config = { billing: patch };
+
+  const result = await withSpinner(
```
wyattjoh: `config push.ts:75-92` fetches the current config, computes a diff with `hasConfigChanges` / `printDiff`, prints it, then prompts. The shortcut commands send blind.

If confirmation is added per the critical-priority findings, pair it with the diff, otherwise the prompt just says "Proceed?" with no context about what's about to change.

Fix: reuse `hasConfigChanges` / `printDiff` from `config/push.ts`. They already handle partial-payload patch mode, which is what these wrappers generate.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,270 @@
...
+    () =>
+      withApiContext(
+        patchInstanceConfig(ctx.appId, ctx.instanceId, config),
+        "Failed to enable billing",
+      ),
+  );
+
+  log.data(JSON.stringify(result, null, 2));
+  log.success(`Billing enabled for ${target === "org" ? "organizations" : "users"}`);
+}
```
wyattjoh: `log.data(JSON.stringify(result, null, 2))` emits the full server response on stdout after every mutation, here and at billing/index.ts:80, 128, 229, 267, plus orgs/index.ts:41, 73.

Per `.claude/rules/logging.md`, `log.data` is stdout for pipeable output. `config patch` does this because the user opted into a low-level config operation. For `clerk orgs enable` the user asked a single-intent question, piping a ~200-line config blob out every time is noise and will trip up shell pipelines.

Fix: demote to `log.debug(...)` for routine success, or emit only when `--json` / `--verbose` is set. Keep the `log.success(...)` line as the human-facing signal.

packages/cli-core/src/commands/orgs/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  yes?: boolean;
+}
+
+export async function orgsEnable(options: OrgsOptions): Promise<void> {
+  const ctx = await resolveAppContext(options);
+
+  const patch: Record<string, unknown> = { enabled: true };
+  if (options.forceSelection) patch.force_organization_selection = true;
+  if (options.domains) patch.domains_enabled = true;
+  if (options.maxMembers) patch.max_allowed_memberships = parseInt(options.maxMembers, 10);
```
wyattjoh:
```ts
if (options.forceSelection) patch.force_organization_selection = true;
if (options.domains) patch.domains_enabled = true;
if (options.autoCreate) { … enabled: true }
```
There's no `--no-force-selection` or `--force-selection=false`. To turn a flag off while keeping orgs enabled, the user has to fall back to `clerk config patch`, partially defeating the wrapper.

Fix: either (a) document in the README that these flags are one-way and point users to `config patch` for the inverse, or (b) accept Commander's boolean negation (`--no-force-selection`) and write `false` when it's set.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,270 @@
...
+    log.info("No plans configured. Use `clerk billing plans create` to add one.");
+    return;
+  }
+
+  for (const [slug, plan] of entries) {
+    const amount = plan.amount as number;
+    const currency = (plan.currency as string) || "usd";
+    const price = amount === 0 ? "Free" : `${(amount / 100).toFixed(2)} ${currency.toUpperCase()}`;
+    const payer = plan.payer_type as string;
+    const visible = plan.publicly_visible !== false;
```
wyattjoh: `cyan(plan.name as string)` casts without checking. If the config contains a plan with no `name` field (legacy or malformed), output reads `undefined (slug) — $X.XX/mo …`.

Fix: `cyan((plan.name as string) ?? slug)`.

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  // `clerk init` doesn't bundle clerk-billing — it's opt-in. Surface it here.
+  if (!options.dryRun && options.skills !== false) {
+    await offerBillingSkillInstall(options);
+  }
+
+  if (!options.dryRun) printNextSteps(NEXT_STEPS.ENABLE_BILLING);
+}
+
+async function offerBillingSkillInstall(options: BillingOptions): Promise<void> {
+  const skipPrompt = options.yes === true || isAgent();
```
rafa-thayto, line 81: `offerBillingSkillInstall` runs even when `applyConfigPatch` returns early ("No changes detected"). On an already-configured instance, every `clerk enable billing` invocation in agent mode will silently attempt to install `clerk-billing` again since `isAgent()` skips the confirmation prompt.

Consider having `applyConfigPatch` return a boolean so you can gate the skill offer:

```suggestion
  const applied = await applyConfigPatch({
    ctx,
    payload,
    verb: `Enabling billing for ${describeTargets(targets)}`,
    successMessage: `Billing enabled for ${describeTargets(targets)}`,
    failureContext: "Failed to enable billing",
    yes: options.yes,
    dryRun: options.dryRun,
  });

  // `clerk init` doesn't bundle clerk-billing -- it's opt-in. Surface it here.
  if (applied && !options.dryRun && options.skills !== false) {
    await offerBillingSkillInstall(options);
  }

  if (applied && !options.dryRun) printNextSteps(NEXT_STEPS.ENABLE_BILLING);
```

(with `applyConfigPatch` returning `false` on the early "No changes" exit)

packages/cli-core/src/commands/config/apply-patch.ts
```diff
@@ -0,0 +1,61 @@
...
+  const headline = dryRun
+    ? `[dry-run] Proposing PATCH on ${ctx.appLabel} (${ctx.instanceLabel}):`
+    : `${verb} on ${ctx.appLabel} (${ctx.instanceLabel}):`;
+  log.info(`\n${headline}\n`);
+  printDiff(current, payload, true);
+
+  if (!dryRun && isHuman() && !yes) {
+    if (warning) log.warn(warning);
+    const ok = await confirm({ message: "Proceed?" });
+    if (!ok) throwUserAbort();
```
rafa-thayto, line 49: The `warning` is only printed inside the `!dryRun && isHuman() && !yes` block. When `orgsDisable` calls this with `--yes` (or in agent mode with `--yes`), the warning about stranded billing is silently dropped and the PATCH fires with no indication.

Consider moving the warning emission before the confirmation gate so it's always visible:

```suggestion
  if (warning) log.warn(warning);

  if (!dryRun && isHuman() && !yes) {
    const ok = await confirm({ message: "Proceed?" });
    if (!ok) throwUserAbort();
  }
```

rafa-thayto (approved):

PR #226, opened 2026-04-23
fix(agent): warn on sandboxed Clerk runs

wyattjoh:
## Summary

This PR teaches the Clerk CLI to detect likely sandboxed agent runs at the actual host-integration boundaries and warn once per invocation before agents trust auth, link, env, config, or API failures.

It also adds a dedicated `Host execution` check to `clerk doctor` in agent mode, and updates the bundled Clerk skill so agents know how to recognize the warning, when to rerun on the host, and how sandboxed runs can misreport auth and targeting state.

## What changed

- Added `packages/cli-core/src/lib/host-execution.ts` as the shared host-capability warning layer.
- Moved sandbox detection to the real host-sensitive library boundaries instead of command wrappers:
  - config file access
  - credential store and keychain access
  - outbound network requests
  - browser launch
  - localhost OAuth callback binding
  - API catalog cache/network access
- Added a `Host execution` diagnostic to `clerk doctor` in agent mode, with structured detail and remedy output.
- Updated doctor tests to cover the new host-execution check.
- Updated the bundled `skills/clerk` docs and references to explain the sandbox warning semantics, host rerun guidance, and explicit `agent` mode behavior.
- Consolidated the changeset into a single patch release note covering the CLI warning, doctor check, and skill guidance.

## Test plan

- [x] `bun test packages/cli-core/src/lib/config.test.ts`
- [x] `bun test packages/cli-core/src/commands/whoami/index.test.ts`
- [x] `bun test packages/cli-core/src/commands/doctor/doctor.test.ts`
- [x] `bun test packages/cli-core/src/cli-program.test.ts`
- [x] `bun test packages/cli-core/src/lib/auth-server.test.ts` on the host
- [x] `bun run --filter @clerk/cli-core typecheck`
- [x] `bun run --filter @clerk/cli-core lint`
- [x] `bun run --filter @clerk/cli-core format:check`
- [x] Verified `CLERK_MODE=agent bun run packages/cli-core/src/cli.ts whoami --verbose`, `apps list --json --verbose`, and `doctor --json --spotlight` warn in the sandbox
- [x] Verified the same commands do not emit the warning on the host

Refs AIE-896.

### DIFF
--- a/packages/cli-core/src/lib/auth-server.ts
+++ b/packages/cli-core/src/lib/auth-server.ts
@@ -155,6 +155,7 @@ export function startAuthServer(expectedState: string): AuthServerResult {
       },
     });
   } catch (error) {
+    clearTimeout(timeout);
     observeHostCapabilityFailure("localhost-bind", error, {
       operation: "listen",
       target: "127.0.0.1:0",
--- a/packages/cli-core/src/lib/auth-server.ts
+++ b/packages/cli-core/src/lib/auth-server.ts
@@ -4,6 +4,7 @@
  */
 
 import { AUTH_TIMEOUT_MS, CALLBACK_PATH } from "./constants.ts";
+import { observeHostCapabilityFailure } from "./host-execution.ts";
 import { log } from "./log.ts";
 
 function escapeHtml(str: string): string {
@@ -79,6 +80,7 @@ interface AuthServerResult {
 export function startAuthServer(expectedState: string): AuthServerResult {
   let resolveCallback: (value: { code: string }) => void;
   let rejectCallback: (reason: Error) => void;
+  let server: ReturnType<typeof Bun.serve> | undefined;
 
   const callbackPromise = new Promise<{ code: string }>((resolve, reject) => {
     resolveCallback = resolve;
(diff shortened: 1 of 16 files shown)


rafa-thayto (approved):

.changeset/doctor-fixes.md
```diff
@@ -2,4 +2,14 @@
...
+
+The CLI now warns once per invocation when host-only Clerk state or system
+capabilities are unavailable in agent mode, which helps distinguish real auth
+and linking failures from sandbox-induced ones. `clerk doctor` also includes a
+`Host execution` check in agent mode so the sandbox condition is visible in
+structured diagnostics.
+
+This release also updates the bundled Clerk skill docs to explain the warning,
+when to rerun commands on the host, and how sandboxed agent runs can misreport
+auth, linking, env, and API failures.
```
rafa-thayto: I think you should create another .changeset

packages/cli-core/src/lib/auth-server.ts
```diff
@@ -88,78 +90,88 @@ export function startAuthServer(expectedState: string): AuthServerResult {
...
+        });
+      },
+    });
+  } catch (error) {
+    observeHostCapabilityFailure("localhost-bind", error, {
+      operation: "listen",
+      target: "127.0.0.1:0",
+      label: CALLBACK_PATH,
+    });
+    throw error;
```
rafa-thayto, line 164: If `Bun.serve` throws here, `server` stays `undefined` but the `setTimeout` from line 92 is still live. When it eventually fires, it calls `rejectCallback` on a `callbackPromise` that nobody is listening to (since `startAuthServer` already threw), which could surface as an unhandled promise rejection.

Adding `clearTimeout(timeout)` before re-throwing should fix it:

```suggestion
    clearTimeout(timeout);
    throw error;
```

PR #227, opened 2026-04-23
docs(skill): teach +clerk_test email + test-phone pattern for agents

wyattjoh:
## Summary

When an AI agent creates a test user via `clerk api /users` and tries to sign in as that user, Clerk's default-on client trust blocks the flow: the OTP email or SMS goes to an address the agent cannot read. The bundled `clerk` skill teaches user creation with plain `alice@example.com` style addresses and says nothing about either of Clerk's test-account patterns, so agents hit a dead end during demos and automated sign-in flows.

This PR teaches both agent-facing surfaces (the bundled Clerk skill and the `clerk init --prompt` framework handoffs) the full test-account contract that our own E2E helper already relies on:

- **Email:** `+clerk_test` subaddress, fixed OTP `424242`.
- **Phone:** US fictional numbers `+1 (XXX) 555-0100` through `+1 (XXX) 555-0199`, same `424242` OTP.

Both patterns are dev-instance-only and test addresses and numbers do not count against Clerk's monthly caps.

## What changed

- **New "Test users (development only)" subsection in `skills/clerk/references/recipes.md`.** Placed directly after the existing "Create a user" block under `## Users`, so agents browsing user-management recipes find the test variants in the same scroll. The section opens with a shared intro, then presents email and phone patterns side by side, each with its own `clerk api /users` example using the exact body shape the Backend API expects. Uses `skip_password_checks: true` to mirror the flag we set in `test/e2e/lib/test-user.ts`.
- **One-sentence pointer appended to every `clerk init --prompt` handoff.** All 13 framework prompts under `packages/cli-core/src/commands/init/prompts/*.md` now carry a uniform sentence in their `## After Setup` section so agents learn the pattern at scaffold time, when they are most likely to exercise auth-gated code. The existing human-signup instruction is left intact because it targets a different flow that does not hit client trust.
- **Docs link points at the canonical Clerk page** (`https://clerk.com/docs/guides/development/testing/test-emails-and-phones`) for the full magic-values contract and production warnings.
- **Patch changeset for `clerk`.** Skill content and init prompts ship inside the compiled `clerk` binary via `{ type: "text" }` imports in `packages/cli-core/src/commands/init/prompts/index.ts`, so a content change is user-facing and belongs in the changelog.

## Test plan

- [x] `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test` (79 passed)
- [x] `bun changeset status` reports `clerk` at `patch`
- [ ] CI E2E suite on this PR (the fixture helper already uses `+clerk_test`, so any regression in the underlying Clerk behavior shows up there)
- [ ] Reviewer: skim the new "Test users" subsection and confirm the email range, phone range, OTP, and Clerk docs link are all correct

## History

Originally shipped as two stacked PRs (#227 email, #229 phone) and folded into this single PR after review feedback noted the phone addition did not conceptually depend on the email one. #229 is marked MERGED in the GitHub UI only because the force-push that carried the fold made the phone commit reachable from this branch's history; the phone content reaches main through this PR.

Refs AIE-897.

### DIFF
--- a/.changeset/aie-897-test-phone-users.md
+++ b/.changeset/aie-897-test-phone-users.md
@@ -1,5 +1,0 @@
----
-"clerk": patch
----
-
-Extend the bundled skill's "Test users" recipe with the US fictional-phone pattern (`+1 (XXX) 555-0100` through `+1 (XXX) 555-0199`) so agents can create test users that verify with OTP `424242` via SMS flows without real phone access.
--- a/.changeset/aie-897-test-phone-users.md
+++ b/.changeset/aie-897-test-phone-users.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Extend the bundled skill's "Test users" recipe with the US fictional-phone pattern (`+1 (XXX) 555-0100` through `+1 (XXX) 555-0199`) so agents can create test users that verify with OTP `424242` via SMS flows without real phone access.
--- a/.changeset/aie-897-test-user-guidance.md
+++ b/.changeset/aie-897-test-user-guidance.md
@@ -2,4 +2,4 @@
 "clerk": patch
 ---
 
-Teach agents the `+clerk_test` email suffix and `424242` OTP pattern for creating test users that bypass client trust in development, both in the bundled skill's recipes and in every `clerk init --prompt` handoff.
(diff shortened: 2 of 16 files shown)


wyattjoh:
**Stack: wyattjoh/aie-897-linear-issue-brainstorming-session**

- **#227 👈 this PR**

dmoerner (approved):
Non-blocking, but I wonder if it would be worth explicitly mentioning the test phone numbers in recipes.md

PR #229, opened 2026-04-23
docs(skill): add test-phone pattern for agent test users

wyattjoh:
## Summary

The parent PR taught agents how to create test users with `+clerk_test` email addresses that bypass client trust via the fixed `424242` OTP. This PR completes the pattern by adding the companion phone guidance: Clerk also treats US fictional phone numbers in the `+1 (XXX) 555-0100` through `+1 (XXX) 555-0199` range as test phones that verify with the same OTP. Agents building SMS-first or multi-factor flows now have a documented, deterministic way to exercise them in development without real phone access.

## What changed

- **Restructured the "Test users (development only)" recipe in `skills/clerk/references/recipes.md`** so the email and phone patterns sit side by side under a shared intro, rather than treating the email pattern as the only option and mentioning phones only in a trailing footnote. Each pattern gets its own `clerk api /users` example using the exact body shape the Backend API expects.
- **Updated the canonical docs link** to `https://clerk.com/docs/guides/development/testing/test-emails-and-phones` to match Clerk's current site layout.
- **Called out the dev-instance monthly caps** (20 SMS, 100 emails) and the fact that test addresses and numbers do not count against them, because that is the first question an agent asks when seeing the recipe for the first time.
- **Patch changeset for `clerk`.** Bundled skill content change, same rationale as the parent PR.

## Test plan

- [x] `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test` (79 passed)
- [x] `bun changeset status` reports `clerk` at `patch` (both PRs in the stack contribute)
- [ ] CI on this PR (stacked on top of #227, so CI here includes both changes)
- [ ] Reviewer: confirm the phone-range bullet is accurate and the Clerk docs link still resolves

Refs AIE-897.

### DIFF
--- a/.changeset/aie-897-test-phone-users.md
+++ b/.changeset/aie-897-test-phone-users.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Extend the bundled skill's "Test users" recipe with the US fictional-phone pattern (`+1 (XXX) 555-0100` through `+1 (XXX) 555-0199`) so agents can create test users that verify with OTP `424242` via SMS flows without real phone access.
--- a/skills/clerk/references/recipes.md
+++ b/skills/clerk/references/recipes.md
@@ -54,10 +54,12 @@ clerk api /users/user_abc123 -X DELETE --yes
 
 ### Test users (development only)
 
-For test accounts you need to sign into without real email access, use the `+clerk_test` email suffix. Clerk recognizes these as test emails: no message is sent, and the verification code is a fixed `424242`. The domain portion is arbitrary.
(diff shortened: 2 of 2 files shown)


wyattjoh:
**Stack: wyattjoh/aie-897-linear-issue-brainstorming-session**

- #227
  - **#229 👈 this PR**

wyattjoh:
Note for history: this PR was folded into #227, not landed to main. After the fold, the parent branch's force-push to origin brought the phone commit into the parent's ancestry, which GitHub auto-detected as a merge and flipped this PR's state to MERGED. The phone content ships as part of #227; nothing here reached main on its own.

PR #230, opened 2026-04-23
fix(update): target the running binary instead of the first PATH hit

rafa-thayto:
## Summary

- Fix `clerk update` picking the wrong install when multiple `clerk` binaries coexist on the same machine (e.g. bun + asdf-npm). The primary update target is now the install that owns `process.execPath` — the binary the user just invoked — instead of the first `PATH` hit. PATH-first remains the fallback when the running binary isn't on PATH or has no known owner.
- Extract the match logic to a pure helper `findRunningInstallIndex` in `lib/installer.ts`, export the previously-private `safeRealpath` (it's now needed from the command), and fix a latent bug where the PATH-empty fallback stored the un-realpath'd `process.execPath` as `resolvedPath` (breaks owner matching on macOS `/var` → `/private/var` paths).
- Rewrite the update command's `README.md` "Primary target selection" section; the old "Why PATH-walking matters" text documented the buggy design as intentional.

### Repro (from the bug report)

```
$ clerk update
◇  Checking for updates
│    Current: 1.0.0
│    Latest:  1.0.2
│    Target:  /Users/.../.asdf/installs/nodejs/22.16.0/lib/node_modules/clerk/bin/clerk (npm)
✔ Update clerk 1.0.0 → 1.0.2? Yes
◇  Updated npm
│  Also found 1 other clerk install:
│    /Users/.../.bun/install/global/node_modules/clerk/bin/clerk (bun)
│
└  Successfully updated to 1.0.2

$ clerk -v
1.0.0                      # ← bun's install (what the shell actually runs) is unchanged
$ which clerk
/Users/.../.bun/bin/clerk  # ← bun, not npm
```

After this fix the bun install is picked as primary and gets upgraded, so `clerk -v` reflects the new version.

### Why owner-matching, not path-equality

`process.execPath` inside the compiled CLI lands on the platform binary at `@clerk/cli-<arch>/bin/clerk`, while PATH candidates resolve to the shim at `clerk/bin/clerk`. Both sit under the same installer's `node_modules`, so `ownerOfBinary()` returns the same installer for both. `installDirs[installer]` tracks exactly one active dir per installer, so "same owner" uniquely identifies "same install" — no cross-install false matches. Binaries under *inactive* asdf-nodejs versions return `owner=null` and safely fall back to PATH order.

## Test plan

- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test` (79 passed; 8 new unit tests for `findRunningInstallIndex` covering: bug-case reproduction, reverse-case, already-first, empty-candidates, unowned execPath, no-matching-candidate, Homebrew, and inactive-asdf-nodejs fallback)
- [x] Local `clerk update` smoke test on a machine with both bun and asdf-npm installs (pending CI)

### DIFF
--- a/packages/cli-core/src/commands/update/index.ts
+++ b/packages/cli-core/src/commands/update/index.ts
@@ -5,10 +5,12 @@ import {
   asdfPluginFromPath,
   asdfReshim,
   findClerkOnPath,
+  findRunningInstallIndex,
   getInstallerPackageDirs,
   globalInstallCommand,
   ownerOfBinary,
   resolveAsdfShim,
+  safeRealpath,
   type Installer,
 } from "../../lib/installer.ts";
 import { log } from "../../lib/log.ts";
@@ -61,9 +63,29 @@ async function resolveTargets(
     unique.push(c);
   }
 
+  // Promote the install that owns the currently-running binary to primary.
+  // The binary the user just invoked is the authoritative update target:
+  // shell hash caches (zsh/bash) and PATH ordering quirks (asdf shims before
+  // `~/.bun/bin`) can make a fresh PATH walk disagree with what actually ran,
+  // leaving `clerk -v` unchanged after an "Updated" message because a
+  // different install got the upgrade. When the running binary isn't on PATH
+  // (invoked by absolute path, PATH mutated mid-session), fall through to
+  // PATH order.
+  const runningResolved = await safeRealpath(runningPath);
+  const runningIdx = findRunningInstallIndex(unique, runningResolved, installDirs);
+  const reordered =
+    runningIdx > 0
+      ? [unique[runningIdx]!, ...unique.slice(0, runningIdx), ...unique.slice(runningIdx + 1)]
+      : unique;
+
   // Fallback when PATH discovery yields nothing: use the running binary.
+  // `resolvedPath` gets the realpath'd variant so ownerOfBinary sees the
+  // same form (e.g. `/private/var/...` on macOS) that PM install dirs use —
+  // otherwise an unresolved `/var/...` execPath would fail owner matching.
   const effective =
-    unique.length > 0 ? unique : [{ displayPath: runningPath, resolvedPath: runningPath }];
+    reordered.length > 0
+      ? reordered
+      : [{ displayPath: runningPath, resolvedPath: runningResolved }];
 
   const toTarget = (c: { displayPath: string; resolvedPath: string }): Target => ({
     displayPath: c.displayPath,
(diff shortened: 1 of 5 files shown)


rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

packages/cli-core/src/commands/update/index.ts
```diff
@@ -61,9 +63,29 @@ async function resolveTargets(
...
+  // The binary the user just invoked is the authoritative update target:
+  // shell hash caches (zsh/bash) and PATH ordering quirks (asdf shims before
+  // `~/.bun/bin`) can make a fresh PATH walk disagree with what actually ran,
+  // leaving `clerk -v` unchanged after an "Updated" message because a
+  // different install got the upgrade. When the running binary isn't on PATH
+  // (invoked by absolute path, PATH mutated mid-session), fall through to
+  // PATH order.
+  const runningResolved = await safeRealpath(runningPath);
+  const runningIdx = findRunningInstallIndex(unique, runningResolved, installDirs);
+  const reordered =
```
wyattjoh, line 78: Primary-target selection here is a decision drawn from multiple sources (running-install owner vs. PATH order), which is exactly the pattern `.claude/rules/debug-logging.md` calls out for `log.debug()` instrumentation. If the collision ever reoccurs under a new shell/installer combination, there's nothing in `--verbose` output that explains *why* a particular target won — the user would have to reason from `process.execPath`, PATH, and `installDirs` themselves.

A single line at the promotion site would cover it, e.g.:

```ts
log.debug(
  `update: primary=${effective[0]!.resolvedPath} (runningIdx=${runningIdx}, execPath=${runningResolved})`,
);
```

Fires once per invocation, no noise, keeps the decision auditable.

packages/cli-core/src/commands/update/index.ts
```diff
@@ -61,9 +63,29 @@ async function resolveTargets(
     unique.push(c);
   }
 
+  // Promote the install that owns the currently-running binary to primary.
+  // The binary the user just invoked is the authoritative update target:
+  // shell hash caches (zsh/bash) and PATH ordering quirks (asdf shims before
+  // `~/.bun/bin`) can make a fresh PATH walk disagree with what actually ran,
+  // leaving `clerk -v` unchanged after an "Updated" message because a
+  // different install got the upgrade. When the running binary isn't on PATH
+  // (invoked by absolute path, PATH mutated mid-session), fall through to
```
wyattjoh, line 75: `findClerkOnPath()` (line 47) and `safeRealpath(runningPath)` are independent fs traversals but run sequentially. Trivial to fold together:

```ts
const [onPath, runningResolved] = await Promise.all([
  findClerkOnPath(),
  safeRealpath(runningPath),
]);
```

Then drop the separate `await safeRealpath(...)` further down. Cosmetic; no behavior change.

wyattjoh (approved):

wyattjoh (approved):

PR #232, opened 2026-04-23
feat(cli): add --input-json for passing options as JSON

rafa-thayto:
## Summary

- Adds a global `--input-json <json|@file>` flag. Keys convert from camelCase/snake_case to kebab-case and expand into argv flags before Commander parses, so any existing flag can come from JSON instead of a shell-escaped string.
- Built for AI-agent and scripted invocations where passing a JSON object is cleaner than composing a command line.

## Behavior

| JSON             | Expansion                               |
| ---------------- | --------------------------------------- |
| `"str"` / number | `--flag <value>`                        |
| `true`           | `--flag`                                |
| `false` / `null` | omitted                                 |
| `["a","b"]`      | `--flag a --flag b` (empty arrays omit) |
| `{…}` (nested)   | rejected — `invalid_json`, exit `2`     |

**Placement.** Put `--input-json` after the leaf subcommand. Before a subcommand, expanded flags land on the root program, so only root options (`--mode`, `--verbose`) work there — subcommand flags like `--json` / `--app` error as unknown. Explicit flags after `--input-json` override its values (last-flag-wins).

**Flags only.** Positional arguments (e.g. the `<name>` in `clerk apps create <name>`, the path in `clerk api <path>`) cannot come from JSON — pass them on the command line as usual.

```sh
clerk init --input-json '{"framework":"next","yes":true}'
clerk config pull --input-json '{"keys":["auth_email","session"]}'
clerk init --input-json @init-opts.json
```

## Test plan

- [x] `bun run test` — 81 files, all passing
- [x] 41 unit tests (`lib/input-json.test.ts`) cover key conversion, expansion rules, `@file` reads, error paths
- [x] 25 integration tests (`test/integration/input-json.test.ts`) cover `init`, `doctor`, `api` (positional args), `apps list`, `config pull`/`patch`/`schema` (nested subcommands + array flags), precedence, and placement (both working and erroring cases)
- [x] Changeset: `clerk` minor bump

## Docs

- New `--input-json` section in `skills/clerk/references/agent-mode.md` (expansion table, placement rule, error codes).
- Pointer in `skills/clerk/SKILL.md`.
- Root `README.md` help block regenerated.


wyattjoh (commented):
Think it's possible for us to support stdin as well? That way the agent can just pipe to it.

wyattjoh (approved):

PR #240, opened 2026-04-28
feat(users): add users command scaffolding and `clerk users create`

wyattjoh:
## Summary

Adds the `clerk users` command family with the first subcommand, `clerk users create`. The top-level `users` command exposes shared targeting (`--app`, `--secret-key`, `--instance`, `--dry-run`, `--yes`, `--json`) that upcoming subcommands will reuse.

`clerk users create` accepts curated flags (`--email`, `--phone`, `--username`, `--password`, `--first-name`, `--last-name`, `--external-id`) for the common path, and `-d, --data <json>` / `--file <path>` for raw BAPI bodies when a field the curated flags don't expose (`primary_email_address_id`, `skip_password_checks`, `web3_wallets`) is needed. Program-level `--input-json` composes cleanly so agents can drive curated flags from a JSON object. BAPI enforces identifier and required-field rules server-side, so the command only needs a BAPI secret key, not the `applications:manage` Platform API scope.

In human mode, `clerk users` invoked with no subcommand opens an interactive menu that dispatches to registered actions, and `clerk users create` invoked without curated flags or `-d`/`--file` enters a guided wizard. The wizard reads the instance's Frontend API user_settings and prompts only for fields the instance accepts, marking required ones. When no project is linked the wizard falls back to a shared application picker (also used by `clerk link`), so users don't have to run `clerk link` first. Agent mode disables every interactive flow and exits with a structured usage error instead.

The application picker now lists "Create a new application" at the bottom and de-emphasizes it until highlighted, so it reads as a fallback rather than a primary choice. `lib/listage.ts` gained a per-choice `style` hook to support that without baking dim styling into the prompt theme.

Alongside `create`, this PR lands shared scaffolding upcoming subcommands will build on: `lib/bapi-command.ts` (secret-key resolution and error formatting), `lib/users.ts` (payload builders, JSON input parsing), `commands/users/output.ts` (mutation print helpers), and `commands/users/lifecycle-runner.ts` (shared runner for direct state transitions).

## Test plan

- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test`
- [x] `bun run test:e2e:op` (every framework fixture's test-user creation now runs through `clerk users create -d`)
- [ ] Reviewer: spot-check `clerk users create --help` and the curated-flags + `-d` interaction
- [ ] Reviewer: drive the interactive `clerk users create` wizard with no project linked, confirm picker → wizard → BAPI handoff

### DIFF
--- a/packages/cli-core/src/commands/users/create-wizard.ts
+++ b/packages/cli-core/src/commands/users/create-wizard.ts
@@ -1,0 +1,99 @@
+import { input, password } from "@inquirer/prompts";
+import {
+  bootstrapDevBrowser,
+  decodePublishableKey,
+  fetchUserSettings,
+  type InstanceType,
+  type UserSettingsJSON,
+} from "../../lib/fapi.ts";
+import { withSpinner } from "../../lib/spinner.ts";
+import { isEnabled, isRequired, type AttributeName } from "./interactive/attributes.ts";
+import { resolveUsersInstanceContext } from "./interactive/instance-context.ts";
+
+export type CreateWizardFields = {
+  email?: string;
+  phone?: string;
+  username?: string;
+  password?: string;
+  firstName?: string;
+  lastName?: string;
+};
+
+export type CreateWizardTargeting = {
+  app?: string;
+  instance?: string;
+  secretKey?: string;
+};
+
+export type CreateWizardResult = {
+  fields: CreateWizardFields;
+  targeting: CreateWizardTargeting;
+};
+
+type WizardOptions = {
+  app?: string;
+  instance?: string;
+  secretKey?: string;
+};
+
+type FieldDef = {
+  attr: AttributeName;
+  key: keyof CreateWizardFields;
+  message: string;
+  isPassword?: boolean;
+};
+
+const ALL_FIELDS: FieldDef[] = [
+  { attr: "email_address", key: "email", message: "Email address" },
+  { attr: "phone_number", key: "phone", message: "Phone number" },
+  { attr: "username", key: "username", message: "Username" },
+  { attr: "password", key: "password", message: "Password", isPassword: true },
+  { attr: "first_name", key: "firstName", message: "First name" },
+  { attr: "last_name", key: "lastName", message: "Last name" },
+];
+
+export async function runCreateWizard(options: WizardOptions): Promise<CreateWizardResult> {
+  const ctx = await resolveUsersInstanceContext(options);
+  const settings =
+    ctx.fapiHost && ctx.publishableKey
+      ? await loadSettings(ctx.fapiHost, decodePublishableKey(ctx.publishableKey).instanceType)
+      : undefined;
+
+  const fields: CreateWizardFields = {};
+  for (const field of ALL_FIELDS) {
+    const enabled = settings ? isEnabled(settings, field.attr) : true;
+    if (!enabled) continue;
+    const required = settings ? isRequired(settings, field.attr) : false;
+    const value = await promptField(field, required);
+    if (value) fields[field.key] = value;
+  }
+
+  const targeting: CreateWizardTargeting = {};
+  if (ctx.appId) targeting.app = ctx.appId;
+  if (ctx.instanceId) targeting.instance = ctx.instanceId;
+  if (ctx.secretKey) targeting.secretKey = ctx.secretKey;
+
+  return { fields, targeting };
+}
+
+async function loadSettings(
+  fapiHost: string,
+  instanceType: InstanceType,
+): Promise<UserSettingsJSON> {
+  return withSpinner("Loading instance settings...", async () => {
+    const jwt = instanceType === "development" ? await bootstrapDevBrowser(fapiHost) : undefined;
+    return fetchUserSettings(fapiHost, jwt ? { jwt } : {});
+  });
+}
+
+async function promptField(field: FieldDef, required: boolean): Promise<string> {
+  const message = required ? `${field.message} *` : `${field.message} (optional)`;
+  const validate = required
+    ? (value: string) => value.trim().length > 0 || `${field.message} is required`
+    : undefined;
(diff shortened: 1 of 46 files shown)


wyattjoh:
!snapshot

packages/cli-core/src/lib/users.ts
```diff
@@ -0,0 +1,124 @@
...
+  phone?: string;
+  username?: string;
+  password?: string;
+  firstName?: string;
+  lastName?: string;
+  externalId?: string;
+}) {
+  const payload: Record<string, unknown> = {};
+
+  if (options.email) payload.email_address = [options.email];
```
rafa-thayto, line 19: These truthiness checks (`if (options.email)`) will silently drop empty strings, which matters for `buildUpdateUserPayload` where `""` could mean "clear this field". Using `!= null` preserves the "omit if not provided" semantics while allowing explicit empty strings:

```suggestion
  if (options.email != null) payload.email_address = [options.email];
  if (options.phone != null) payload.phone_number = [options.phone];
  if (options.username != null) payload.username = options.username;
  if (options.password != null) payload.password = options.password;
  if (options.firstName != null) payload.first_name = options.firstName;
  if (options.lastName != null) payload.last_name = options.lastName;
  if (options.externalId != null) payload.external_id = options.externalId;
```

packages/cli-core/src/commands/users/lifecycle-runner.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+    if (handleBapiError(error)) {
+      return;
+    }
+    throw error;
+  }
+}
+
+function getLifecycleSuccessMessage(path: string): string {
+  const userId = getUserIdFromPath(path);
```
rafa-thayto: These two parallel if-chains (`getLifecycleSuccessMessage` and `getLifecycleErrorMessage`) duplicate the same `path.endsWith` logic across ~60 lines. A lookup map would cut this down significantly:

```suggestion
const LIFECYCLE_LABELS: Record<string, { success: string; error: string }> = {
  "/ban": { success: "Banned user", error: "Failed to ban user" },
  "/unban": { success: "Unbanned user", error: "Failed to unban user" },
  "/lock": { success: "Locked user", error: "Failed to lock user" },
  "/unlock": { success: "Unlocked user", error: "Failed to unlock user" },
  "/profile_image": { success: "Removed profile image for user", error: "Failed to remove profile image for user" },
  "/mfa": { success: "Disabled MFA for user", error: "Failed to disable MFA for user" },
  "/totp": { success: "Removed TOTP for user", error: "Failed to remove TOTP for user" },
  "/backup_code": { success: "Removed backup codes for user", error: "Failed to remove backup codes for user" },
};

function getLifecycleLabel(path: string, type: "success" | "error"): string {
  const userId = getUserIdFromPath(path);
  const suffix = "/" + (path.split("/").pop() ?? "");
  const labels = LIFECYCLE_LABELS[suffix];
  const fallback = type === "success" ? "Updated user" : "Failed to update user";
  return `${labels?.[type] ?? fallback} ${userId}`;
}
```

packages/cli-core/src/commands/users/create-wizard.ts
```diff
@@ -0,0 +1,99 @@
...
+  { attr: "username", key: "username", message: "Username" },
+  { attr: "password", key: "password", message: "Password", isPassword: true },
+  { attr: "first_name", key: "firstName", message: "First name" },
+  { attr: "last_name", key: "lastName", message: "Last name" },
+];
+
+export async function runCreateWizard(options: WizardOptions): Promise<CreateWizardResult> {
+  const ctx = await resolveUsersInstanceContext(options);
+  const settings =
+    ctx.fapiHost && ctx.publishableKey
```
rafa-thayto: If FAPI is down or the request fails, `loadSettings` throws `FapiError` and the entire wizard crashes. This contrasts with `fetchAppsTolerantly` in `app-picker.ts` which gracefully degrades on 5xx. Wrapping this in a try-catch would let the wizard fall back to showing all fields as optional (which it already does when `fapiHost` is undefined):

```suggestion
  let settings: UserSettingsJSON | undefined;
  if (ctx.fapiHost && ctx.publishableKey) {
    try {
      settings = await loadSettings(ctx.fapiHost, decodePublishableKey(ctx.publishableKey).instanceType);
    } catch {
      // Fall back to prompting the full curated set when FAPI is unreachable.
    }
  }
```

packages/cli-core/src/lib/users.ts
wyattjoh: The build* helpers are fed exclusively from individual Commander flag values, and the JSON-clear path goes through `mergeUsersPayload` (untouched by these helpers). For create, `""` semantically means absent rather than clear. For update, even if BAPI accepted `""` as clear for some fields, applying `!= null` uniformly here would also forward `password: ""`, `email_address: [""]`, `phone_number: [""]`, which BAPI rejects.

If a user wants to clear `first_name`/`last_name`/`external_id` on update, the supported route is `--data '{"first_name":""}'`, which already bypasses this branch via `mergeUsersPayload`.

Is there a caller you have in mind where an empty string surfaces through the flag parser? Happy to scope `!= null` to just `firstName`/`lastName`/`externalId` on the update helper if so.

packages/cli-core/src/commands/users/create-wizard.ts
wyattjoh: Going with the try/catch + a `log.debug("create-wizard: failed to load FAPI user settings, prompting full curated set: ...")` line so the fallback shows up under `--verbose` (per `.claude/rules/debug-logging.md`).

Holding off on a `log.warn` because the all-optional fallback is non-destructive and the user can still abandon the wizard. Let me know if youd rather make it a warn.

rafa-thayto (approved):

PR #241, opened 2026-04-28
feat(init): remove --prompt and bundled agent prompt templates

rafa-thayto:
## Summary

- Remove the `clerk init --prompt` flag and the per-framework prompt templates that lived in `packages/cli-core/src/commands/init/prompts/`. Agents should now run `clerk init -y` for the full non-interactive flow, or install the published skills directly via `skills add clerk/skills`.
- Move the small `pmInstallCommand` helper to `lib/package-manager.ts` (it lived next to the prompt templates only because they referenced it; its only other consumer was `init/heuristics.ts`).
- Lift the `declare module "*.md"` text-import declaration from the deleted `prompts/md.d.ts` into `globals.d.ts` so `skills.ts` and `skill/install.ts` keep type-checking.
- Tests that previously used `init --prompt` as a non-side-effecting smoke test for `--input-json` and `--app` registration are rewritten to assert via `clerk.raw` (no `unknown option` in stderr / structured agent error format) without depending on the removed flag.
- Init `README.md`, the `cli-program.ts` examples, and the `scaffold.ts` doc comment are updated to drop `--prompt` references.

## Versioning

Released as a `minor` bump rather than `major` — `--prompt` had no known users, so the removal is treated as additive cleanup rather than a breaking change.

## Test plan
- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test` (93 passed)
- [ ] `bun run test:e2e:op`


packages/cli-core/src/lib/package-manager.ts
```diff
@@ -8,6 +8,18 @@ export const PACKAGE_MANAGERS = ["bun", "pnpm", "yarn", "npm"] as const;
...
+  yarn: "yarn add",
+  pnpm: "pnpm add",
+  npm: "npm install",
+} satisfies Record<PackageManager, string>;
+
+/** Returns the `<pm> add`-style command for installing dependencies. */
+export function pmInstallCommand(pm: PackageManager): string {
+  return PM_INSTALL_COMMANDS[pm];
+}
+
```
wyattjoh, line 22: Maybe valuable to combine with:

https://github.com/clerk/cli/blob/88c3754868cfc415aa8ad771f49ed7e044e6dd95/packages/cli-core/src/lib/installer.ts#L314-L326

packages/cli-core/src/globals.d.ts
```diff
@@ -1,3 +1,8 @@
+declare module "*.md" {
+  const content: string;
+  export default content;
+}
+
```
wyattjoh, line 5: Why was this added?

wyattjoh (approved):

PR #243, opened 2026-04-28
fix(update): Do not hang if package manager is not present

dmoerner:
I don't have `yarn` installed, and this produced a hang on `clerk update`. The cause of the hang was that it was sitting waiting for user confirmation to download `yarn`, but never surfaced this to the user. Instead, ignore `stdin`, explicitly do not use corepack download prompts, and add a timeout.


dmoerner:
!snapshot

rafa-thayto (approved):

PR #244, opened 2026-04-29
feat(init): Allow agent init without implicit app selection

djgould:
## Summary
- classify agent init runs into real-app, keyless, or manual setup modes
- use keyless for keyless-capable frameworks when no app target exists
- avoid implicit app selection/creation and fail real-app agent runs without auth guidance

## Validation
- bun test packages/cli-core/src/commands/init/index.test.ts packages/cli-core/src/commands/link/index.test.ts packages/cli-core/src/test/integration/agent-mode.test.ts
- bun run --filter @clerk/cli-core typecheck
- bun run --filter @clerk/cli-core format:check
- bunx oxfmt --check skills/clerk/SKILL.md skills/clerk/references/agent-mode.md
- git diff --check


packages/cli-core/src/commands/init/index.ts
```diff
@@ -214,28 +227,36 @@ function printBootstrapNextSteps(
...
+  ctx: ProjectContext;
+  authed: boolean;
+  hasRealAppTarget: boolean;
+}): boolean {
   if (ctx.framework.supportsKeyless) {
+    if (agent) return !hasRealAppTarget;
+
+    // Auto-keyless is scoped to bootstrap (new-project) flows only in human
+    // mode. Existing projects keep the authenticated flow so real keys can be
+    // pulled unless an agent explicitly chooses keyless by omitting an app.
```
djgould, line 247: to be honest i'm not sure this is what we want, but its what we were doing before. i find the fact that `clerk init` doesn't require auth for bootstrapping and does require auth in existing projects a bit odd

packages/cli-core/src/commands/init/index.ts
```diff
@@ -253,19 +273,33 @@ function resolveKeylessMode(
...
   if (hasApiKey) return "Using API key";
 
   const email = await getAuthenticatedEmail();
   if (email) return `Logged in as ${email}`;
 
+  if (!allowInteractiveLogin) {
+    throwUsageError(
+      "clerk init in agent mode requires CLERK_PLATFORM_API_KEY or a valid stored session when using a real Clerk application. Pass --app only with auth available, or omit --app for keyless-capable frameworks.",
+    );
+  }
```
rafa-thayto: This throws `USAGE_ERROR` but it is actually an auth problem, not a syntax issue. Agents that branch on error codes (as the `ERROR_CODE` JSDoc suggests) would misinterpret this as a bad command instead of missing credentials.

```suggestion
    throwUsageError(
      "clerk init in agent mode requires CLERK_PLATFORM_API_KEY or a valid stored session when using a real Clerk application. Pass --app only with auth available, or omit --app for keyless-capable frameworks.",
      undefined,
      ERROR_CODE.AUTH_REQUIRED,
    );
```

djgould: looking at this with fresh eyes this morning and i realized that we shouldn't do this check at all. we can just trigger the login flow in agent mode when --app is passed, i just tested it and it works well.

removed the check entirely

rafa-thayto (approved):

rafa-thayto (approved):

packages/cli-core/src/commands/init/index.ts
```diff
@@ -205,28 +218,35 @@ function printBootstrapNextSteps(
...
+  authed: boolean;
+  hasRealAppTarget: boolean;
+}): boolean {
   if (ctx.framework.supportsKeyless) {
+    if (agent) return !hasRealAppTarget;
+
+    // Auto-keyless is scoped to bootstrap (new-project) flows only in human
+    // mode. Existing projects keep the authenticated flow so real keys can be
+    // pulled unless an agent explicitly chooses keyless by omitting an app.
+    if (!bootstrap) return false;
```
wyattjoh, line 248: Refactoring `resolveKeylessMode` moved the `if (!bootstrap) return false;` early-return inside the `if (ctx.framework.supportsKeyless)` block. The non-keyless branch (which still ends in `log.info("X requires API keys — keyless mode is not yet supported for this framework.")`) is no longer guarded by `bootstrap`, so the log now fires on paths where keyless was never being considered:

- Human mode re-running `clerk init` in an existing Vue / Express / Fastify project (previously silent).
- Agent mode with `--app` on a non-keyless framework (the real-app flow is what's running; the user gets a misleading 'keyless not supported' note).
- The new agent manual-setup path, where `printBootstrapManualSetupInfo` already prints a clearer 'requires API keys — set them up manually' block. Users see two near-duplicate messages back-to-back.

The new integration test `init prints manual setup for non-keyless framework without an app target in agent mode` only asserts `expect(stderr).toContain("clerk init --app <app_id>")`, which passes despite the duplicate output, so the regression isn't caught.

Suggested fix: short-circuit on the non-keyless framework before the function does anything else, and let the caller own user-facing copy:

```ts
function resolveKeylessMode({...}): boolean {
  if (!ctx.framework.supportsKeyless) return false;
  if (agent) return !hasRealAppTarget;
  if (!bootstrap) return false;
  return !authed;
}
```

If the bootstrap-of-non-keyless framework hint is still wanted, emit it at the bootstrap call site (e.g., next to `printBootstrapManualSetupInfo`) rather than from inside the resolver. Either way, please add a regression test that asserts stderr does **not** contain `keyless mode is not yet supported` for an existing-project non-keyless human-mode run.

PR #250, opened 2026-04-30
feat: extend next-step suggestions to more commands

rafa-thayto:
## Summary

- Adds contextual next-step output to eight commands: `auth logout`, `switch-env`, `unlink`, `whoami`, `completion`, `skill install`, `config patch`, and `config put`.
- Centralizes the new strings in the `NEXT_STEPS` registry (`lib/next-steps.ts`); per-shell completion install hints stay local to `commands/completion/index.ts` since they aren't reusable vocabulary.
- All suggestions are gated on `isHuman()` via the existing `printNextSteps` helper, so agent mode and piped output stay clean.

### What users see now

After commands like `clerk unlink` or `clerk skill install`, the success line is followed by 1–3 cyan arrows pointing at the natural next workflow step (e.g. `clerk apps list`, `clerk config pull`, `clerk doctor`). `whoami` branches between linked / unlinked profiles for relevance. `completion` emits stderr-only install hints tailored per shell (bash/zsh/fish/powershell).

### Side-effect: `whoami` resilience fix

The previous `try/catch` wrapped both `fetchUserInfo` and `log.data(userInfo.email)` and attributed any throw to `session_expired`. Tightened the try to the network call only — incidental but worth noting.

## Test plan

- [x] `bun run typecheck` — pass
- [x] `bun run test` — 82 files pass, 0 fail (179 total tests including the touched commands)
- [x] Manually verified the suggestion text for each command path against the existing `NEXT_STEPS` style ("Run `clerk X` to Y")
- [ ] CI green

### DIFF
--- a/packages/cli-core/src/commands/completion/index.ts
+++ b/packages/cli-core/src/commands/completion/index.ts
@@ -29,7 +29,7 @@ const INSTALL_HINTS: Record<SupportedShell, readonly string[]> = {
   ],
   fish: [
     "Run `clerk completion fish | source` to enable completions in this session",
-    "Run `clerk completion fish > ~/.config/fish/completions/clerk.fish` to install permanently",
+    "Run `clerk completion fish > $__fish_config_dir/completions/clerk.fish` to install permanently",
   ],
   powershell: [
     "Run `clerk completion powershell | Out-String | Invoke-Expression` to enable completions in this session",
--- a/packages/cli-core/src/commands/completion/index.ts
+++ b/packages/cli-core/src/commands/completion/index.ts
@@ -3,6 +3,7 @@ import { generate as generateZsh } from "./shells/zsh.ts";
 import { generate as generateFish } from "./shells/fish.ts";
 import { generate as generatePowershell } from "./shells/powershell.ts";
 import { throwUsageError } from "../../lib/errors.ts";
+import { printNextSteps } from "../../lib/next-steps.ts";
 
 type CompletionGenerator = (binaryName: string) => string;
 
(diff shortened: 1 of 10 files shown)


packages/cli-core/src/commands/completion/index.ts
```diff
@@ -17,6 +18,25 @@ const GENERATORS: Record<SupportedShell, CompletionGenerator> = {
...
+    'Run `eval "$(clerk completion bash)"` to enable completions in this session',
+    "Append the same line to ~/.bashrc to make it permanent",
+  ],
+  zsh: [
+    'Run `eval "$(clerk completion zsh)"` to enable completions in this session',
+    "Append the same line to ~/.zshrc to make it permanent",
+  ],
+  fish: [
+    "Run `clerk completion fish | source` to enable completions in this session",
+    "Run `clerk completion fish > ~/.config/fish/completions/clerk.fish` to install permanently",
```
dmoerner: I think ` > $__fish_config_dir/completions/clerk.fish` would be a little less fragile here, although perhaps uglier.

dmoerner (approved):
Approving, I dropped a few more comments on specific wordings on Slack, I'll copy them here:

- CONFIG_PUSH, clerk doctor: I think "integration" is too vague here, like if someone does a `push` to set up an oauth provider, that might be what they connect to "integration" in context.
- CONFIG_UNLINK: Let's put clerk link first in the array
- SKILL_INSTALL, clerk config pull: I think the message here is kind of misleading, since it's not like doing that will pull the live config into an agent's context. I also wonder if we should make clear that manual clerk init is an alternative if you don't have an agent do it for you. I wonder if we need anything other than telling them to start Claude or whatever. (Or Codex?)
- dry run: We could have the push as the next step.

PR #252, opened 2026-04-30
fix(init): Avoid errors on directory name collisions

dmoerner:
Auto-increment the default project name in `clerk init` when a directory with that name already exists, and re-prompt instead of erroring out when the chosen name collides.


dmoerner:
!snapshot

rafa-thayto (approved):
LGTM

PR #254, opened 2026-05-01
feat(init): create real app when authed in agent mode

wyattjoh:
## Summary

When `clerk init` runs in agent mode without `--app` and the user is already authenticated, the CLI now creates a real Clerk application named after the project and links it, instead of falling back to keyless setup. Previously an authed agent run created an unowned keyless app and wrote a deferred-claim breadcrumb, so the project ended up unlinked even though the user had logged in before the run; auto-claim only fired on a subsequent `clerk auth login`, which authed users had no reason to re-run. Keyless still runs in agent mode when the user is not authenticated.

The change adds a `createIfMissing` library option to `link` so the agent-mode autolink path can fall through to a deterministic create instead of the existing usage error, and threads a project-name derivation (`package.json#name`, `--name`, or directory basename) through `init`'s authenticated branch.

## Test plan

- [ ] `bun run typecheck`
- [ ] `bun run lint`
- [ ] `bun run test`
- [ ] Manual: `clerk init --framework next --pm npm` in agent mode while logged in → real app appears in the dashboard, `.env.local` has dev keys, `clerk doctor` reports linked
- [ ] Manual: same command while not logged in → keyless flow still runs, breadcrumb still written, original deferred-claim path still works on next `clerk auth login`

### DIFF
--- a/.changeset/agent-init-authed-link.md
+++ b/.changeset/agent-init-authed-link.md
@@ -1,0 +1,5 @@
+---
+"clerk": minor
+---
+
+`clerk init` in agent mode now creates and links a real Clerk application when the user is authenticated, instead of falling back to keyless setup. Keyless still runs in agent mode when the user is not authenticated, but authenticated agent runs leave the project properly linked with real development keys in `.env`.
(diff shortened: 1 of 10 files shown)


Railly (commented):
Reviewed locally. Logic checks out, autolink runs first to avoid duplicates, the guard still trips when `createIfMissing` is missing.

A couple things worth a look:

PLAPI accepts duplicate app names. Created two apps with the same name to confirm, different IDs each time. Agents running `clerk init` across multiple Next default projects will fill the dashboard with "my-app" entries. Worth either disambiguating the derived name or mentioning it in the changeset.

Also poked at `createAndFetchApp` and the second call is redundant. The raw `POST /v1/platform/applications` response already includes `application_id`, `name`, and a full `instances[]` array with `instance_id`, `environment_type`, `secret_key`, `publishable_key`. The follow-up `GET /v1/platform/applications/{id}` returns the same shape minus `secret_key`, so it actually drops information from the create response. Doesn't break anything because `link` only persists the instance ID and `env pull` fetches keys later, but it's an extra PLAPI round-trip per authed init that buys nothing. Easy to drop the second call, or leave a comment if it's defensive.

Otherwise looks good.

Railly (approved):

PR #255, opened 2026-05-01
fix(test/e2e): make users-list importable by refresh-fixtures script

wyattjoh:
## Summary

The biweekly Refresh E2E Fixtures workflow has been failing on `main` since #237 landed `test/e2e/users-list.test.ts`. The refresh script imports every `test/e2e/*.test.ts` file to read each one's `FixtureConfig`, but `users-list.test.ts` threw at module top-level when `CLERK_CLI_TEST_APP_ID` and `CLERK_PLATFORM_API_KEY` were unset, which they are in that workflow because the script does not actually need them.

This change moves the env-var requirement into `beforeAll`, wraps the `bun:test` hook registrations in a `CLERK_REFRESH_FIXTURES` guard (matching the convention in `test/e2e/lib/fixture-test.ts`), and teaches the refresh script to skip imported files that do not export a `FixtureConfig`. Either fix on its own is insufficient: `bun:test`'s `beforeAll` itself throws when called outside the runner, and even with the file made import-safe, the script still crashes on `config.pinned` for non-fixture files. The combination also future-proofs the refresh against the next non-fixture e2e test that lands.

## Test plan

- [x] `bun run format` / `lint` / `typecheck` / `test` clean locally
- [x] `CLERK_REFRESH_FIXTURES=1 bun -e "import('./test/e2e/users-list.test.ts')"` succeeds
- [x] `bun run e2e:refresh-fixtures -- --only nextjs-pages-router` runs end-to-end
- [ ] Manually trigger the Refresh E2E Fixtures workflow on `main` after merge to verify

### DIFF
--- a/scripts/refresh-e2e-fixtures.ts
+++ b/scripts/refresh-e2e-fixtures.ts
@@ -54,8 +54,18 @@ for (const testFile of testFiles) {
   matchedOnly = true;
 
   const { config } = (await import(testFile)) as {
-    config: import("../test/e2e/lib/types.ts").FixtureConfig;
+    config?: import("../test/e2e/lib/types.ts").FixtureConfig;
   };
+
+  // Skip non-fixture e2e tests (e.g. live-API roundtrip tests that don't
+  // scaffold a project and therefore don't export a FixtureConfig).
+  if (!config) {
+    if (onlyName) {
+      console.error(`❌ ${name} is not a fixture (no config export).`);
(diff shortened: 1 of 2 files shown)


wyattjoh:
**Stack: wyattjoh/fix-refresh-fixtures-import**

- **#255 👈 this PR**

Railly (approved):

PR #260, opened 2026-05-06
feat(deploy): added deploy wizard

wyattjoh:
## Summary

Adds `clerk deploy`, an interactive wizard that promotes a linked Clerk application from development to production. The wizard validates clone compatibility, creates the production instance and primary domain, walks the user through the returned CNAME records (with optional BIND zone export to `./clerk-<domain>.zone`), verifies DNS/SSL/mail with a per-component spinner chain, and collects production OAuth credentials for Google, GitHub, Microsoft, Apple, and Linear.

State is derived from the Platform API on each run, not persisted locally. The wizard reads the live application, domains, deploy status, and instance config on entry and resumes from the next pending step. The only profile write is `instances.production` once the production instance exists. Pressing Ctrl-C or picking "Skip" at an OAuth provider exits with `Deploy paused` and exit code 1; rerunning resumes from whatever the API now reports.

Platform API errors specific to the deploy lifecycle are mapped to typed `CliError`s in `commands/deploy/errors.ts` (unsupported plan features, provider/taken domains, `production_instance_exists` recovery, SSL/mail retry throttles). The wizard refuses to start in agent mode because production configuration is interactive.

## Test plan

- [ ] `bun run format:check`
- [ ] `bun run lint`
- [ ] `bun run typecheck`
- [ ] `bun run test`
- [ ] `bun run test:e2e:op`
- [ ] Manual smoke: unlinked directory exits with `error: No Clerk project linked`; linked directory walks through create → DNS records → BIND export → verification → OAuth → production summary; Ctrl-C at OAuth exits with `Deploy paused` and exit code 1; rerun resumes at the skipped provider.


wyattjoh:
!snapshot

wyattjoh:
**Stack: wyattjoh/deploy**

- **#260 👈 this PR**

packages/cli-core/src/commands/deploy/index.ts
```diff
@@ -1,257 +1,865 @@
...
+  domainIdOrName: string,
+  domain: string,
+): Promise<DeployStatusOutcome> {
+  await triggerDeployStatusCheck(appId, domainIdOrName);
+  let response = await mapDeployError(getApplicationDomainStatus(appId, domainIdOrName));
+  let status = deployComponentStatusFromDomainStatus(response);
+  let retriesRemaining = DEPLOY_STATUS_MAX_RETRIES;
+  let nextRetryDelay = DEPLOY_STATUS_INITIAL_RETRY_DELAY_MS;
+
+  for (const component of DEPLOY_COMPONENT_ORDER) {
```
rafa-thayto: `retriesRemaining` and `nextRetryDelay` live outside the component loop, so all three spinners share one budget. With the constants here (5 retries, 3s base, x2 backoff) that totals ~93s, and whichever component is checked first (mail) can spend all of it. Once it does, the dns and ssl spinners reach `while (retriesRemaining > 0)` with zero left and return `false` right away, reporting the domain as unverified even when one more poll would have flipped them. Give each component its own budget:

```suggestion
  for (const component of DEPLOY_COMPONENT_ORDER) {
    let retriesRemaining = DEPLOY_STATUS_MAX_RETRIES;
    let nextRetryDelay = DEPLOY_STATUS_INITIAL_RETRY_DELAY_MS;
```

packages/cli-core/src/commands/deploy/index.ts
```diff
@@ -1,257 +1,865 @@
...
-  // Mock state — dev instance has Google OAuth enabled
-  const devSettings = {
-    socialProviders: ["google"],
+function deployComponentStatusFromDomainStatus(
+  response: DomainStatusResponse,
+): DeployComponentStatus {
+  return {
+    dns: response.dns?.status === "complete",
+    ssl: response.ssl ? checkStatusComplete(response.ssl) : false,
+    mail: checkStatusComplete(response.mail),
```
rafa-thayto: These three components handle a missing field three different ways. `dns` reads as pending when absent and ignores `required` entirely, `ssl` reads as pending when absent, but `mail` goes through `checkStatusComplete` where `!undefined?.required` is `true`, so an absent `mail` field reports as verified without ever being checked. Worth making all three uniform so the absent case means the same thing everywhere:

```ts
    dns: checkStatusComplete(response.dns),
    ssl: checkStatusComplete(response.ssl),
    mail: checkStatusComplete(response.mail),
```

That needs `dns?: DomainCheckStatus` in `plapi.ts` and `checkStatusComplete` taking `{ status: string; required?: boolean } | undefined`. Decide whether an absent component should count as complete or pending and apply it in the one helper rather than per field.

packages/cli-core/src/commands/deploy/index.ts
```diff
@@ -1,257 +1,865 @@
...
 
-  log.debug(`Found authenticated user: ${user.email} (${user.id})`);
-  log.debug(`Found linked application: ${application.name} (${application.id})`);
+async function resolveLiveDeploySnapshot(
+  ctx: DeployContext,
+): Promise<LiveDeploySnapshot | undefined> {
+  const productionInstanceId = ctx.productionInstanceId;
+  if (!productionInstanceId) return undefined;
+
+  const domain = await loadProductionDomain(ctx);
```
rafa-thayto: `loadProductionDomain` and `loadDevelopmentOAuthProviders` both take only `ctx` and don't depend on each other, yet they run one after the other, adding a round trip to every resume run. Only `loadProductionState` needs `domain.id`, so the first two can go together:

```ts
  const [domain, oauth] = await Promise.all([
    loadProductionDomain(ctx),
    loadDevelopmentOAuthProviders(ctx),
  ]);
  if (!domain) return undefined;
  const { descriptors: oauthProviderDescriptors, unsupported } = oauth;
```

packages/cli-core/src/commands/deploy/providers.ts
```diff
@@ -0,0 +1,456 @@
...
+ * Human-readable provider label, with shared metadata preferred.
+ */
+export function providerLabel(provider: string): string {
+  return SHARED_OAUTH_METADATA.get(provider)?.name ?? fieldLabel(provider);
+}
+
+/**
+ * Prompt fields for existing deploy callers, falling back to standard OAuth credentials.
+ */
+export function providerFields(provider: OAuthProvider): OAuthField[] {
```
rafa-thayto, line 372: These five lookups (`providerFields`, `providerCredentialLabel`, `providerRedirectLabel`, `providerSetupCopy`, `providerGotcha`) repeat the same `if (hasProviderOverride(provider)) return MAP[provider]; return default` shape. A small generic collapses the repetition:

```ts
function fromOverride<T>(
  provider: OAuthProvider,
  map: Record<ProviderWithOverride, T>,
  fallback: T,
): T {
  return hasProviderOverride(provider) ? map[provider] : fallback;
}

export function providerFields(provider: OAuthProvider): OAuthField[] {
  return fromOverride(provider, PROVIDER_FIELDS, [
    { key: "client_id", label: "Client ID" },
    { key: "client_secret", label: "Client Secret", secret: true },
  ]);
}
// providerCredentialLabel / providerRedirectLabel / providerSetupCopy / providerGotcha follow the same shape.
```

rafa-thayto (changes requested):

packages/cli-core/src/commands/deploy/copy.ts
```diff
@@ -0,0 +1,278 @@
...
+    default:
+      return "CNAME";
+  }
+}
+
+export function dnsDashboardHandoff(domain: string): string[] {
+  return [
+    `Check the Domains section in the Clerk Dashboard for ${domain} to monitor DNS propagation and SSL issuance.`,
+    "After OAuth setup, you can verify DNS or skip and finish. DNS propagation can take time.",
+  ];
```
rafa-thayto, line 119: I think we can add this tip here, or if we have the instance and app here, it would generate a better endpoint:

```suggestion
    "https://dashboard.clerk.com/last-active?path=domains"
  ];
```

or a better one:

```suggestion
    `https://dashboard.clerk.com/apps/${app}/instances/${instance}/domains`
  ];
```

wyattjoh:
!snapshot

rafa-thayto (approved):

wyattjoh:
!snapshot

PR #261, opened 2026-05-06
feat: honor HTTP_PROXY/HTTPS_PROXY env vars in loggedFetch

rafa-thayto:
## Summary
- Bun's `fetch()` does not read `HTTP_PROXY`/`HTTPS_PROXY` env vars, so traffic inspectors (mitmproxy, Charles, Proxyman) couldn't intercept the CLI's outbound requests.
- `loggedFetch` now resolves a proxy from curl-style env vars (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` — uppercase or lowercase) and passes it through Bun's per-request `proxy` option.
- Localhost (`localhost`, `127.0.0.1`, `::1`) is always skipped so the OAuth callback listener is never routed through an external proxy.
- With `--verbose`, the chosen proxy is logged alongside the request.

## Test plan
- [ ] `HTTPS_PROXY=http://localhost:8080 clerk apps list --verbose` shows `routing via proxy http://localhost:8080` and the request appears in mitmproxy.
- [ ] `NO_PROXY=api.clerk.com HTTPS_PROXY=http://localhost:8080 clerk apps list --verbose` does NOT route through the proxy.
- [ ] OAuth flow (`clerk auth login`) is unaffected when `HTTPS_PROXY` is set — the localhost callback skips the proxy.
- [ ] Lowercase variants (`https_proxy`, `no_proxy`) are also honored.

### DIFF
--- a/.changeset/fetch-proxy-env.md
+++ b/.changeset/fetch-proxy-env.md
@@ -1,0 +1,7 @@
+---
+"clerk": patch
+---
+
+Honor `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` environment variables for outbound HTTP requests.
+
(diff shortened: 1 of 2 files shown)


rafa-thayto:
!snapshot

PR #264, opened 2026-05-06
fix: detect fish shell via FISH_VERSION when login shell differs

rafa-thayto:
## Summary

- Fix `clerk doctor` shell detection for fish users whose login shell (`$SHELL`) is zsh — now checks `FISH_VERSION` env var as a fallback
- Fix `clerk update` showing an irrelevant `hash -r` hint to the same users
- Add `delete process.env.FISH_VERSION` guard in test `beforeEach` to prevent contamination when running tests from fish

## Details

On macOS, `$SHELL` reflects the login shell from `/etc/passwd`, not the active shell. Fish doesn't overwrite `$SHELL` — it sets `FISH_VERSION` instead. This caused two bugs:

1. **`clerk doctor`** — `detectShell()` read only `$SHELL`, so fish users with a zsh login shell saw zsh completion diagnostics
2. **`clerk update`** — `hashHint()` checked `$SHELL` for fish, missed it, and showed a `hash -r` hint (fish auto-rehashes, so this is noise)

The fix checks `$SHELL` first for bash/zsh (which don't export their version vars to children), then falls back to `FISH_VERSION`. This handles both the primary case (fish with zsh login shell) and the edge case (bash/zsh launched from within fish, where `FISH_VERSION` is still inherited).

## Test plan

- [x] New test: detects fish via `FISH_VERSION` when `SHELL` is unset
- [x] New test: trusts `SHELL` over `FISH_VERSION` when `SHELL` is bash or zsh (bash-inside-fish edge case)
- [x] New test: skips when `SHELL` points to an unrecognized shell (`/bin/csh`)
- [x] Existing fish/zsh/bash tests still pass (48/48)
- [x] `beforeEach` clears `FISH_VERSION` so tests are stable regardless of runner shell

### DIFF
--- a/packages/cli-core/src/commands/doctor/checks.ts
+++ b/packages/cli-core/src/commands/doctor/checks.ts
@@ -363,7 +363,9 @@ type DetectedShell = "bash" | "zsh" | "fish";
 
 function detectShell(): DetectedShell | null {
   const name = process.env.SHELL?.split("/").pop();
-  if (name === "zsh" || name === "bash" || name === "fish") return name;
+  if (name === "bash" || name === "zsh") return name;
+  // FISH_VERSION is always set inside fish, even when $SHELL still points to another shell.
+  if (name === "fish" || process.env.FISH_VERSION) return "fish";
   return null;
 }
 
--- a/packages/cli-core/src/commands/doctor/checks.ts
+++ b/packages/cli-core/src/commands/doctor/checks.ts
@@ -362,10 +362,11 @@ export async function checkCliVersion(): Promise<CheckResult> {
 type DetectedShell = "bash" | "zsh" | "fish";
 
 function detectShell(): DetectedShell | null {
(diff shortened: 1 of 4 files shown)


dmoerner:
!snapshot

packages/cli-core/src/commands/doctor/checks.ts
```diff
@@ -363,7 +363,9 @@ type DetectedShell = "bash" | "zsh" | "fish";
 
 function detectShell(): DetectedShell | null {
   const name = process.env.SHELL?.split("/").pop();
-  if (name === "zsh" || name === "bash" || name === "fish") return name;
+  if (name === "bash" || name === "zsh") return name;
+  // FISH_VERSION is always set inside fish, even when $SHELL still points to another shell.
+  if (name === "fish" || process.env.FISH_VERSION) return "fish";
```
dmoerner: I think this needs to be first, right now we branch early on the previous line if something else is in $SHELL

dmoerner (approved):

PR #268, opened 2026-05-08
feat: make --keyless an opt-in flag on clerk init

rafa-thayto:
## Summary

- `clerk init` now always authenticates and links a real Clerk application by default. The previous silent fallback to auto-generated dev keys (when run unauthenticated on a keyless-capable framework) is removed.
- `--keyless` is the opt-in flag for the keyless flow. On a non-keyless framework it exits with a usage error.
- `-y` no longer bypasses authentication — it skips y/n confirmations only. Without `--keyless`, an unauthenticated user is still prompted to log in via the browser.
- Agent mode without `--keyless` or `--app` prints manual setup guidance instead of silently generating dev keys (agents can't run interactive OAuth).

## Internal

- The `keyless` + `manualSetup` boolean pair is replaced by an `InitStrategy` union (`"keyless" | "manual" | "authenticate"`) so the post-scaffold dispatch is exhaustive.
- Auth-related I/O (`isAuthenticated`, `resolveProfile`) is gated behind `!optsKeyless` so `--keyless` skips disk reads it doesn't need.
- Test fixtures DRYed via shared `KEYLESS_CTX` constant + `mockBootstrapTo` / `mockExistingProject` / `mockMiddlewareScaffold` helpers — collapses ~140 lines of repeated setup.

## Test plan

- [x] `bun run format` clean
- [x] `bun run lint` clean
- [x] `bun run typecheck` clean
- [x] `bun run test` — 97/97 pass (47 in `init/index.test.ts`)
- [x] `bun run build` succeeds
- [x] `bun run build:compile` succeeds; compiled `dist/clerk init --help` shows `--keyless` flag and example
- [x] Manual: run `clerk init` in a fresh Next.js project with no credentials → should prompt browser login
- [x] Manual: run `clerk init --keyless` in a fresh Next.js project → should generate dev keys without prompting login
- [x] Manual: run `clerk init --keyless` in a Vue project → should exit with usage error
- [x] Manual: run `clerk init -y` in a fresh Next.js project with no credentials → should still prompt browser login (not skip auth)


djgould (approved):
LGTM

PR #270, opened 2026-05-11
fix: resolve --file absolute paths in `clerk env pull`

### LINKED ISSUE #269: `clerk env pull --file <path>` prints success but writes no file (exits 1) when framework cannot be detected
## Summary

`clerk env pull --app <id> --instance dev --file <absolute path>` prints `Environment variables written to <path>` and **exits with code 1**, but no file is created at the path. Reproducible in both directories without a `package.json` and in directories with a `package.json` that has no recognized framework dependency.

## Environment

- `clerk` CLI: **1.1.1**
- Install: Homebrew (`/opt/homebrew/bin/clerk`)
- OS: macOS, Darwin 25.4.0 arm64

## Reproduction

```bash
# From any directory (here: a non-framework dir)
$ clerk env pull --app app_xxx --instance dev --verbose --file ~/clerk-dev.env
config: reading /Users/u/Library/Preferences/clerk-cli/config.json
env: profiles from compile-time CLI_ENV_PROFILES (production, staging)
env: active environment is "production" (platformApiUrl=https://api.clerk.com)
framework: no package.json at /Users/u/Developer/some-dir or unable to parse
credentials: checking keyring (service=clerk-cli, account=oauth-access-token)
credentials: found session in keyring
plapi: using OAuth token from credential store for auth (env=production, target=https://api.clerk.com)
plapi: GET https://api.clerk.com/v1/platform/applications/app_xxx?include_secret_keys=true
...
framework: no package.json at /Users/u/Developer/some-dir or unable to parse
Environment variables written to /Users/u/clerk-dev.env
$ echo $?
1
$ ls -la ~/clerk-dev.env
ls: /Users/u/clerk-dev.env: No such file or directory
```

Also reproduces from inside a directory with a `package.json` that lacks a Clerk-recognized framework — the verbose log changes to `framework: no match in <path>/package.json dependencies` but the same exit-1-with-no-file behavior occurs.

## Expected

Either:
1. The file is written to `--file <path>` regardless of framework detection (since `--file` is an explicit absolute path the user provided), or
2. The CLI exits non-zero with an explanatory error, not the misleading "Environment variables written to <path>" success message.

## Actual

CLI prints the success message and exits 1 with no file on disk. Misleading for users and breaks scripted workflows.

## Workaround

Use the Platform API directly:

```bash
clerk api "/platform/applications/<app_id>?include_secret_keys=true" --platform \
  | jq -r '.instances[] | select(.environment_type=="development") | .secret_key'
```

## Suspected cause

The "framework detection failed" path appears to short-circuit the actual write step while still printing the post-write success message. `--file <path>` should bypass framework-specific logic since the user has explicitly named the target.

rafa-thayto:
## Summary

- Fixes #269. `clerk env pull --file <absolute path>` no longer silently writes under the project cwd; it writes exactly where the user asked.
- Root cause: `resolveTargetFile` called `path.join(cwd, flag)`, which doesn't treat an absolute second arg as a reset. `--file /Users/u/clerk-dev.env` from `/Users/u/Developer/some-dir` produced `/Users/u/Developer/some-dir/Users/u/clerk-dev.env`. `Bun.write` auto-creates parent dirs, so the wrong write succeeded and the success log (which printed `options.file` verbatim) was misleading. Swapped to `path.resolve` — absolute paths are honored, relative paths still resolve against cwd.

## Test plan

- [x] Added `pull.test.ts > "writes --file to an absolute path outside cwd"` — file lands at the absolute path, not nested under cwd; success log shows the resolved path.
- [x] Added `pull.test.ts > "writes --file with no package.json present (framework detection fails)"` — reproduces the exact scenario from the issue (non-framework dir + absolute `--file`) and verifies keys land at the requested path.
- [x] Existing 26 env-pull tests still green (28/28 pass).
- [x] `bun run format` / `bun run lint` / `bun run typecheck` / `bun run test` all pass locally (97/97).
- [x] Reviewer to spot-check `clerk env pull --file /tmp/foo.env` from a fresh dir against the prebuilt binary once CI publishes the preview build.

### DIFF
--- a/.changeset/env-pull-absolute-file-path.md
+++ b/.changeset/env-pull-absolute-file-path.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Fix `clerk env pull --file <path>` to honor absolute paths. Previously, absolute paths were silently nested under the current working directory (e.g. `--file /Users/u/clerk-dev.env` wrote to `<cwd>/Users/u/clerk-dev.env`), making the file appear missing at the expected location while the success message claimed it was written there. Both absolute and relative paths now resolve correctly.
--- a/packages/cli-core/src/commands/env/README.md
+++ b/packages/cli-core/src/commands/env/README.md
@@ -10,11 +10,11 @@ clerk env pull [--app <app_id>] [--instance dev|prod|<instance_id>] [--file <pat
 
 ### Options
 
-| Option            | Description                                                  |
-| ----------------- | ------------------------------------------------------------ |
-| `--app <id>`      | Application ID to target directly (works from any directory) |
-| `--instance <id>` | Instance to target (`dev`, `prod`, or a full instance ID)    |
-| `--file <path>`   | Target env file (default: auto-detect)                       |
+| Option            | Description                                                         |
+| ----------------- | ------------------------------------------------------------------- |
+| `--app <id>`      | Application ID to target directly (works from any directory)        |
+| `--instance <id>` | Instance to target (`dev`, `prod`, or a full instance ID)           |
+| `--file <path>`   | Target env file, relative to cwd or absolute (default: auto-detect) |
 
 ## Sequence Diagram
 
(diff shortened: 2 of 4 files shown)


rafa-thayto:
!snapshot

djgould (approved):

PR #274, opened 2026-05-12
feat: accept plural `orgs`/`users` as canonical `--for` values

kylemac:
## Summary

- Switches the canonical `--for` values on `clerk enable billing` and `clerk disable billing` from singular (`org`/`user`) to plural (`orgs`/`users`), since the plural forms read more naturally.
- Keeps the singular forms accepted as backward-compatible aliases so existing scripts and docs keep working.
- Updates help text, examples, error messages, tab-completion hints, READMEs, and the cross-reference in `clerk disable orgs` warnings, plus tests that cover both canonical and alias inputs.

## Test plan

- [ ] `bun run test packages/cli-core/src/commands/billing/index.test.ts`
- [ ] `bun run typecheck`
- [ ] `bun run lint`
- [ ] Manually verify `clerk enable billing --for orgs` and `--for org` both still work

### DIFF
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -626,10 +626,10 @@ Give AI agents better Clerk context: install the Clerk skills
       { command: "clerk enable orgs", description: "Enable organizations" },
       {
         command: "clerk enable orgs --force-selection --max-members 10",
-        description: "Enable organizations with options",
+        description: "Enable orgs with options",
       },
       {
-        command: "clerk enable billing --for org",
+        command: "clerk enable billing --for orgs",
         description: "Enable billing for organizations only",
       },
       {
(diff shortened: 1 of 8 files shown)


dmoerner (approved):

packages/cli-core/src/cli-program.ts
```diff
@@ -626,10 +626,10 @@ Give AI agents better Clerk context: install the Clerk skills
       { command: "clerk enable orgs", description: "Enable organizations" },
       {
         command: "clerk enable orgs --force-selection --max-members 10",
-        description: "Enable organizations with options",
+        description: "Enable orgs with options",
```
rafa-thayto: keep it as `organizations` here
```suggestion
        description: "Enable organizations with options",
```

PR #280, opened 2026-05-12
chore(test): pin fixture deps and manifest-drive the e2e harness

wyattjoh:
## Summary

Reproducibility, harness ergonomics, and a pre-commit-hook fix for the e2e suite.

The e2e fixture tooling moves from `bunx`/`bun install` onto `npx`/`npm ci` with a committed `package-lock.json` in each fixture, and every non-Clerk dependency is resolved to an exact version through `npm view` at refresh time so scaffolded projects no longer drift between runs. Scaffolders that change behavior across majors are pinned through an explicit `pinnedDependencyRanges` config, replacing the older `pinned: boolean` + `--force` flag, and a TanStack peer-dependency mismatch is now expressed declaratively via `packageJsonOverrides`. The resolution, override, and validation logic lives in a new `scripts/lib/fixture-deps.ts` module covered by its own unit tests.

The e2e harness consolidates around a single `createFixtureHarness(name)` entry point. Every fixture's config moves into `test/e2e/fixtures.manifest.ts` keyed by fixture name, so the refresh script can import the manifest directly and no longer needs to load test files or the `CLERK_REFRESH_FIXTURES` env-var shim. The harness handle carries the fixture config and directory, exposes a per-test users API that auto-cleans created users in `afterEach`, and gives `runFixtureTests`/`runFileExistsTest`/`runBrowserTests` a single place to read config from. A dedicated `test/e2e/tsconfig.json` brings the e2e tree under `bun run typecheck`, and e2e logging is silenced unless `CLERK_E2E_DEBUG=1`.

The custom `scripts/run-tests.ts` wrapper is dropped in favor of running `bun test --parallel` directly, which implies `--isolate` and gives module-state isolation between test files for free. Lint and format coverage extend over `test/e2e/`, with an `.oxfmtrc.json` that excludes fixture lockfiles and a no-console oxlint override scoped to the e2e tree where the logger is a deliberate console wrapper.

## Test plan

- [ ] `bun run format:check`
- [ ] `bun run lint`
- [ ] `bun run typecheck`
- [ ] `bun run test`
- [ ] `bun run test:e2e:op`


wyattjoh:
**Stack: wyattjoh/nextjs-pinned-range**

- **#280 👈 this PR**
  - #287

scripts/lib/fixture-deps.ts
```diff
@@ -0,0 +1,119 @@
...
+  pinnedDependencyRanges: Record<string, string> | undefined,
+): string[] {
+  if (!pinnedDependencyRanges) return [];
+
+  const warnings: string[] = [];
+
+  for (const [dep, range] of Object.entries(pinnedDependencyRanges)) {
+    const deps = pkg.dependencies;
+    const devDeps = pkg.devDependencies;
+    const target = deps?.[dep] !== undefined ? deps : devDeps?.[dep] !== undefined ? devDeps : null;
```
rafa-thayto: Nested ternary is hard to read. `Array.prototype.find` over the two candidate maps reads more directly and drops the need for the explicit `null` sentinel (the subsequent `if (!target)` already handles `undefined`).

```suggestion
    const target = [pkg.dependencies, pkg.devDependencies].find(
      (candidate) => candidate?.[dep] !== undefined,
    );
```

scripts/lib/fixture-deps.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      const resolved = await resolveVersion(name, spec);
+      const resolvedExact = semver.valid(resolved);
+      if (!resolvedExact) {
+        throw new Error(`${name}@${spec} resolved to non-exact version "${resolved}"`);
+      }
+
+      deps[name] = resolvedExact;
+    }
+  }
```
rafa-thayto, line 67: Each `npm view` is awaited serially, so refreshing a fixture with ~30 deps spends ~30 sequential round-trips against the registry. Collecting the resolutions into `Promise.all` shrinks that to a single batched wait while preserving the same writes.

```suggestion
  for (const field of DEPENDENCY_FIELDS) {
    const deps = pkg[field];
    if (!deps) continue;

    const tasks: Promise<void>[] = [];
    for (const [name, spec] of Object.entries(deps)) {
      if (EXACT_VERSION_EXCLUDED_PACKAGE_SCOPES.some((scope) => name.startsWith(scope))) {
        continue;
      }

      const exact = semver.valid(spec);
      if (exact) {
        deps[name] = exact;
        continue;
      }

      tasks.push(
        (async () => {
          const resolved = await resolveVersion(name, spec);
          const resolvedExact = semver.valid(resolved);
          if (!resolvedExact) {
            throw new Error(`${name}@${spec} resolved to non-exact version "${resolved}"`);
          }
          deps[name] = resolvedExact;
        })(),
      );
    }
    await Promise.all(tasks);
  }
```

rafa-thayto (changes requested):

rafa-thayto (approved):

PR #281, opened 2026-05-12
refactor(ci): inline bun setup, gate cache on snapshot

wyattjoh:
## Summary

Drops the `.github/actions/setup-bun` composite action and inlines its three operations (`oven-sh/setup-bun@v2`, `actions/cache/restore@v5`, `bun install --frozen-lockfile`) at all eight call sites. In `ci.yml`'s `build`, `lint`, `test`, and `test-e2e` jobs, the cache restore step is gated on `github.event_name != 'issue_comment'` so the snapshot path — which checks out PR-author-controlled code in the default branch's privileged context via `release.yml`'s `workflow_call` — never consumes the cache. The other workflows (`release.yml` versioning + canary-version, `notify-failure`, `notify-release`) inline directly without a gate because they only run on trusted triggers. Cache population continues to live in the standalone `warm-bun-cache.yml` workflow that runs solely on push to `main`.

## Test plan

- [ ] Local CI suite passes (`bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test`).
- [ ] CodeQL re-scan on the PR clears the three remaining `actions/cache-poisoning/poisonable-step` alerts.
- [ ] On merge to `main`, `Warm Bun cache` runs and the next PR build reads the warmed cache.


wyattjoh:
**Stack: wyattjoh/refactor-setup-bun-cache**

- **#281 👈 this PR**

rafa-thayto (approved):

PR #287, opened 2026-05-13
refactor(test): standardize log capture on useCaptureLog

wyattjoh:
## Summary

Standardize unit-test log capture on a single helper, and stabilize the integration harness.

Unit-test log capture is rebuilt on top of a `useCaptureLog()` lifecycle helper that installs a fresh buffer per test via `beforeEach`/`afterEach`, replacing the prior AsyncLocalStorage-backed `captureLog().run(...)` wrapper. With the slot installed automatically, test bodies no longer wrap their emitting calls. As part of this, `log.ui()` stops appending its own trailing newline so spinner cursor-control writes survive, the `cli-program` writeErr newline-stripping workaround that compensated for the old behavior is removed, and a latent double-newline bug in the spinner output is fixed by moving the trailing `\n` onto each emitting call site.

The integration harness's config load moves out of a top-level await so each test gets a fresh instance, and a new harness test covers the new shape.

## Test plan

- [ ] `bun run format:check`
- [ ] `bun run lint`
- [ ] `bun run typecheck`
- [ ] `bun run test`


wyattjoh:
**Stack: wyattjoh/nextjs-pinned-range**

- ~~#280~~
  - **#287 👈 this PR**

packages/cli-core/src/lib/log.ts
```diff
@@ -133,20 +132,23 @@ export type CapturedLogs = {
   stderr: string[];
 };
 
-const captureStorage = new AsyncLocalStorage<CapturedLogs>();
+let activeCapture: CapturedLogs | null = null;
 
-export function withCapturedLogs<T>(captured: CapturedLogs, fn: () => T): T {
-  return captureStorage.run(captured, fn);
+export function getActiveCapture(): CapturedLogs | null {
```
rafa-thayto: `getActiveCapture` has no callers anywhere in the branch (both `stubs.ts` and `harness.ts` only import `setActiveCapture`). Tests treat the buffer they passed into `setActiveCapture` as the source of truth, so a read-side accessor on the log module isn't needed. Worth dropping until a real call site appears.

```suggestion
export function setActiveCapture(captured: CapturedLogs | null): void {
  activeCapture = captured;
}
```

rafa-thayto (changes requested):

rafa-thayto (approved):

PR #288, opened 2026-05-13
refactor(errors): expose structured Clerk error fields on ApiError subclasses

wyattjoh:
## Summary

`ApiError` and its `PlapiError` / `FapiError` / `BapiError` subclasses now parse the Clerk error envelope at construction time and expose `code`, `long_message`, `meta`, and `clerk_trace_id` as typed instance fields. The base constructor uses the parsed `message` (or a short truncated fallback for non-JSON bodies) instead of the previous `API error (<status>): <raw body>` string, so default `error.message` rendering already shows the human-readable Clerk message. The global error handler in `cli-program.ts` reads those fields directly rather than re-parsing the body, which collapses the existing ad-hoc body-string branching.

Each subclass gains a `fromResponse(response: Response)` async factory that reads the body once and constructs the error, replacing the manual `response.text()` + `new XError(...)` pattern at every PLAPI / FAPI / KEYLESS / BAPI / users call site. The existing `fromBody` helpers stay for tests and fixtures that already have the body string in hand.

The parser also guards against JSON bodies whose top level is `null` or another non-object value, so a response body of `"null"` no longer crashes the global handler with a `TypeError` and falls back to the truncated-body path instead.

## Test plan

- [ ] `bun run typecheck`
- [ ] `bun run test`
- [ ] CI green

### DIFF
--- a/packages/cli-core/src/cli-program.test.ts
+++ b/packages/cli-core/src/cli-program.test.ts
@@ -3,6 +3,7 @@ import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { createProgram, formatApiBody } from "./cli-program.ts";
+import { ApiError } from "./lib/errors.ts";
 import { STANDARD_AGENT_DIRS, EXTRA_REL_PATHS } from "./lib/skill-detection.ts";
 
 test("registers users as a top-level command", () => {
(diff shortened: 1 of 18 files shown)


wyattjoh:
**Stack: api-errors-and-deploy**

- **#288 👈 this PR**
  - #260

rafa-thayto (approved):

rafa-thayto (approved):

PR #301, opened 2026-05-21
feat(fetch): identify CLI with a Clerk-CLI User-Agent header

rafa-thayto:
## Summary

- Sets `User-Agent: Clerk-CLI/<version> (Bun/<v>; <platform>-<arch>[; ci])` on every outbound request via the central `loggedFetch` helper, replacing Bun's default `Bun/<v>` so Clerk's edge can route or filter CLI traffic separately.
- Routes the lone direct `fetch()` in `api/catalog` through `loggedFetch`, picking up both the UA and standard debug-logging behavior (also fixes a pre-existing violation of `.claude/rules/debug-logging.md`).
- Callers can still override `User-Agent` via headers (verified in `fetch.test.ts`).

## Test plan

- [x] `bun run typecheck` — passes
- [x] `bun run build` — passes
- [x] New tests in `user-agent.test.ts` + `fetch.test.ts` cover UA shape, caller override, platform/arch, CI marker, printable-ASCII — all green
- [x] Updated `token-exchange.test.ts` to read headers via `Headers.get()` (the helper now passes a `Headers` instance through)
- [x] Reviewer: confirm UA shows up in Cloud Run logs as `Clerk-CLI/...` instead of `Bun/...`

### DIFF
--- a/packages/cli-core/src/lib/user-agent.ts
+++ b/packages/cli-core/src/lib/user-agent.ts
@@ -1,0 +1,20 @@
+/**
+ * Identifies the CLI in outbound HTTP calls so Clerk's edge can route or filter
+ * CLI traffic separately (e.g. to dedicated Cloud Run services). Without this
+ * we fall through to Bun's default `User-Agent: Bun/<version>`, which is
+ * indistinguishable from any other Bun-based client.
+ *
+ * Format: `Clerk-CLI/<version> (Bun/<bun-version>; <platform>-<arch>[; ci])`
+ *   - <platform>: darwin | linux | win32 | …  (process.platform)
+ *   - <arch>:     arm64 | x64 | …             (process.arch)
+ *   - `ci` segment is appended when running under a recognized CI environment.
+ */
+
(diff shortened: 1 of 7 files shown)


wyattjoh (approved):

packages/cli-core/src/lib/user-agent.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Format: `Clerk-CLI/<version> (Bun/<bun-version>; <platform>-<arch>[; ci])`
+ *   - <platform>: darwin | linux | win32 | …  (process.platform)
+ *   - <arch>:     arm64 | x64 | …             (process.arch)
+ *   - `ci` segment is appended when running under a recognized CI environment.
+ */
+
+import { DEV_CLI_VERSION, resolveCliVersion } from "./version.ts";
+
+export function buildUserAgent(): string {
+  const version = resolveCliVersion() ?? DEV_CLI_VERSION;
```
jfoshee, line 16: In the future it would be nice to encapsulate this as like `getCliVersion()` so callers don't need to worry about the fallback.

PR #304, opened 2026-05-21
feat(clerk-bird): add leaderboard, audio cue, and surface in help

rafa-thayto:
## Summary

- **Leaderboard**: a local top-10 leaderboard for the hidden `clerk bird` easter egg. From the GAME OVER screen, press `N` to submit a name and `L` to view the board; use `↑`/`↓` (or `j`/`k`) to select a row and `D` to delete it (with `Y`/`N` confirmation). Rankings live in `~/.flap-rankings.json`; the existing `~/.flap-best` file is unchanged.
- **Audio feedback**: each pipe-pass (`+1`) and the death event now emit a short bell tone via ASCII BEL (`\x07`). The host terminal handles cross-platform behavior on Windows, macOS, Linux, and any POSIX TTY — terminals with the bell disabled stay silent, respecting the user's own setting. No subprocess, no platform branching, no bundled audio asset.
- **Discoverability**: `clerk bird` is no longer hidden. The help formatter now renders easter-egg commands after the auto-generated `help` row so the easter egg surfaces in `clerk --help` without cluttering the top of the command list.

## Test plan

- [x] `bun run format && bun run lint && bun run typecheck` — all clean locally
- [x] `bun test packages/extras/src/clerk-bird/flap.test.ts` — 23 passing (includes new `beep` + `it.each` refactors of `removeRanking`)
- [x] `clerk --help` ends with the `bird` row below the `help` row
- [x] `clerk bird` launches the game; passing a pipe produces a `+1` popup **and** a bell tick on the same render frame; death produces a bell tick + screen shake
- [x] In macOS Terminal → Settings → Profiles → Advanced, disable "Audible bell" + "Visual bell"; relaunch the game and confirm scoring still works silently (we don't force audio when the user has disabled it)
- [x] On Windows Terminal / WSL: confirm BEL produces the system default beep on score (validates the no-platform-branching claim)

### DIFF
--- a/packages/extras/src/clerk-bird/flap.test.ts
+++ b/packages/extras/src/clerk-bird/flap.test.ts
@@ -1,0 +1,202 @@
+import { afterEach, describe, expect, it } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { beep, insertEntry, loadRankings, removeRanking } from "./flap.ts";
+
+const tmpDirs: string[] = [];
+
+function makeTmpDir(): string {
+  const dir = mkdtempSync(join(tmpdir(), "flap-rankings-"));
+  tmpDirs.push(dir);
+  return dir;
+}
+
+afterEach(() => {
+  while (tmpDirs.length > 0) {
+    const dir = tmpDirs.pop();
+    if (dir) rmSync(dir, { recursive: true, force: true });
+  }
+});
+
+describe("insertEntry", () => {
+  it("inserts the first entry at rank 1", () => {
+    const result = insertEntry([], { name: "alice", score: 10, ts: 1 });
+    expect(result.rank).toBe(1);
+    expect(result.list).toEqual([{ name: "alice", score: 10, ts: 1 }]);
+  });
+
+  it("sorts higher scores above lower scores", () => {
+    const list = [{ name: "alice", score: 10, ts: 1 }];
+    const result = insertEntry(list, { name: "bob", score: 20, ts: 2 });
+    expect(result.rank).toBe(1);
+    expect(result.list.map((e) => e.name)).toEqual(["bob", "alice"]);
+  });
+
+  it("breaks ties by earlier timestamp", () => {
+    const list = [{ name: "alice", score: 10, ts: 5 }];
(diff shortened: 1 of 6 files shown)


rafa-thayto:
!snapshot

packages/extras/src/clerk-bird/flap.ts
```diff
@@ -263,6 +297,107 @@ function saveBest(n: number): void {
...
+}
+
+function sanitizeEntry(raw: unknown): RankingEntry | null {
+  if (!raw || typeof raw !== "object") return null;
+  const r = raw as Record<string, unknown>;
+  if (typeof r["name"] !== "string") return null;
+  if (typeof r["score"] !== "number" || !Number.isFinite(r["score"])) return null;
+  if (typeof r["ts"] !== "number" || !Number.isFinite(r["ts"])) return null;
+  const name = r["name"].slice(0, MAX_NAME_LEN).replace(/[^\x20-\x7e]/g, "");
+  return { name, score: Math.floor(r["score"]), ts: Math.floor(r["ts"]) };
```
wyattjoh: This keeps any finite numeric `score` from `~/.flap-rankings.json`, including extremely large values. The leaderboard later formats `String(entry.score)` into a fixed-width box, and `padBox()` will call `" ".repeat(...)` with a negative count if a tampered or corrupted rankings file contains a long score such as `Number.MAX_VALUE`. Please either reject/clamp scores in `sanitizeEntry()` or truncate the rendered score string so a local rankings file cannot crash `clerk bird` when the leaderboard opens.

packages/extras/src/clerk-bird/flap.test.ts
```diff
@@ -0,0 +1,202 @@
+import { afterEach, describe, expect, it } from "bun:test";
```
wyattjoh, line 1: This new test file is not picked up by CI today. The root `test` script only runs `bun test 'packages/cli-core/src/' 'scripts/' ...`, and `packages/extras/package.json` does not define a `test` script, while CI calls only `bun run test`. Please add the extras tests to the root test command or add a package test script that CI runs, otherwise these leaderboard tests will silently rot after this PR.

wyattjoh (commented):
I found two issues worth fixing before merging.

wyattjoh (approved):

PR #305, opened 2026-05-21
feat: migrate prompts to Clack

wyattjoh:
## What

- Migrates CLI prompt wrappers, list selection, spinners, and human-facing command output to `@clack/prompts`.
- Routes command input through shared prompt helpers and adds Clack UI output capture for tests.
- Removes direct `@inquirer/*` dependencies and updates prompt-related tests and fixtures.

## Why

This moves the CLI to the new prompt UI model while keeping agent and machine-readable output paths stable.

## Testing

- `bun run typecheck`
- `bun run format:check`
- `bun run test`


packages/cli-core/src/commands/users/create-wizard.ts
```diff
@@ -101,11 +101,11 @@ async function loadSettings(
...
   const validate = required
-    ? (value: string) => value.trim().length > 0 || `${field.message} is required`
+    ? (value: string | undefined) => (value?.trim() ? undefined : `${field.message} is required`)
     : undefined;
   if (field.isPassword) {
     return password({ message, validate });
   }
-  const value = await input({ message, validate });
+  const value = await text({ message, validate });
   return value.trim();
```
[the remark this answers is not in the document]
wyattjoh: packages/cli-core/src/lib/prompts.ts types `text()` as `Promise<string>` after `unwrap()` converts Clack cancel symbols into `UserAbortError`. The `string | undefined` type is only for the validate callback input, matching Clack’s validator contract, not the prompt result. Empty optional fields come back as strings, so `.trim()` is safe here. Am I missing a runtime path where our wrapper returns `undefined`?

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -70,6 +71,8 @@ export async function billingEnable(options: BillingOptions): Promise<void> {
     billing.user_enabled = true;
   }
 
+  intro("Enabling billing");
```
rafa-thayto: `intro()` pushes the gutter prefix (via `pushPrefix()`), and only `outro()`/`pausedOutro()` pop it. Both `billingEnable` and `billingDisable` call `intro()` with no try/finally, so if `applyConfigPatch` throws (including `UserAbortError` when the user cancels the confirmation) the prefix is never popped and no closing bracket prints. The command exits with a dangling `|` gutter instead of the `Failed`/`Paused` close that `api`, `apps create`, `apps list`, and `config push` show on the error path. `config/pull.ts`, `config/schema.ts`, `env/pull.ts`, and `orgs/index.ts` have the same gap.

The guarded commands also copy the same ~15-line try/catch/finally + `closeStatus` block, so one helper in `lib/spinner.ts` fixes the leak and removes the duplication at once:

```ts
export async function withGutter<T>(
  title: string,
  fn: (set: (steps: readonly string[]) => void) => Promise<T>,
  opts?: { skip?: boolean },
): Promise<T> {
  if (opts?.skip) return fn(() => {});
  intro(title);
  let nextSteps: readonly string[] | undefined;
  try {
    const result = await fn((steps) => { nextSteps = steps; });
    outro(nextSteps);
    return result;
  } catch (error) {
    if (error instanceof UserAbortError || isPromptExitError(error)) pausedOutro();
    else outro("Failed");
    throw error;
  }
}
```

Then each command becomes `await withGutter("Enabling billing", async () => { ... })`, which guarantees the close on every path.

packages/cli-core/src/commands/apps/list.ts
```diff
@@ -1,9 +1,11 @@
 import { listApplications, type Application } from "../../lib/plapi.ts";
 import { withApiContext } from "../../lib/errors.ts";
 import { dim, cyan } from "../../lib/color.ts";
-import { withSpinner } from "../../lib/spinner.ts";
+import { UserAbortError, isPromptExitError } from "../../lib/errors.ts";
```
rafa-thayto: Two separate imports from `../../lib/errors.ts` (this line plus the `withApiContext` import two lines up). Merge them into one.

```suggestion
import { withApiContext, UserAbortError, isPromptExitError } from "../../lib/errors.ts";
import { dim, cyan } from "../../lib/color.ts";
```

rafa-thayto (changes requested):

packages/cli-core/src/commands/billing/index.ts
wyattjoh: Implemented this as a shared `withGutter` wrapper in `packages/cli-core/src/lib/spinner.ts` and routed the currently unguarded intro/outro commands through it. I left the existing manual wrappers in `api`, `apps create/list`, and `config push` because they carry command-specific close behavior, for example `api` treats handled `BapiError` as a failed close without rethrowing. Converting those cleanly would need a slightly different control surface than this simple helper.

rafa-thayto:
!snapshot

rafa-thayto (approved):

PR #307, opened 2026-05-25
feat(mcp): clerk mcp install/list/uninstall/run + doctor MCP check

rafa-thayto:
## Summary

Adds a `clerk mcp` command group that connects the Clerk remote MCP server (`https://mcp.clerk.com/mcp`) to 10 AI clients, plus an MCP health check in `clerk doctor`.

- **`clerk mcp install`** — register the server. Human mode: interactive multiselect over detected clients; agent mode / `--all` / `--client <id>` (repeatable): non-interactive. Install **always converges**: whatever entry sits under the name is replaced (remove-then-add through the client's CLI, or an overwrite for file-backed clients) — there is no `--force` and no conflict state.
- **`clerk mcp list`** — show every Clerk-flavored entry (named `clerk` or pointing at a `*.clerk.com` host). JSON output is `{ entries, failures }`, so an unreadable client config surfaces structurally instead of reading as "nothing installed".
- **`clerk mcp uninstall`** — remove the entry. Human mode prompts with the clients that actually have it (nothing pre-checked); agent mode / `--all` targets every client. Nothing matched → warm hint, exit 0.
- **`clerk mcp run`** — a built-in stdio↔Streamable-HTTP bridge that editors spawn (replaces `npx mcp-remote`; no npx dependency). Every client config stores the same URL-less descriptor `{ "command": "clerk", "args": ["mcp", "run"] }`; the bridge resolves its target at spawn time (`CLERK_MCP_URL` > env profile `mcpUrl` > hosted default).
- **`clerk doctor`** — MCP check probes each distinct configured URL via the `initialize` handshake when an entry is installed (skips otherwise; warns, never fails), and distinguishes "config unreadable" from "nothing installed".

## Registration is delegated to each client's own CLI

Clients that ship a **non-interactive** registration CLI are driven through it — the client owns its config format and write safety. The rest get direct file writes. There is **no file-write fallback** for CLI-backed clients: a missing binary fails that client with an actionable error (`mcp_client_cli_not_found` + docs link), and detection is PATH-based so the picker only offers clients we can actually drive.

| Client | `--client` id | Registered via | Removed via |
| --- | --- | --- | --- |
| Claude Code | `claude` | `claude mcp add --scope user` | `claude mcp remove` |
| Cursor | `cursor` | file write (no CLI exists) | file write |
| GitHub Copilot / VS Code | `vscode` (`copilot`) | `code --add-mcp '<json>'` | file write (add-only CLI) |
| Windsurf | `windsurf` | file write (no CLI exists) | file write |
| Gemini CLI | `gemini` | `gemini mcp add --scope user` | `gemini mcp remove` |
| Codex | `codex` | `codex mcp add` (TOML config) | `codex mcp remove` |
| opencode | `opencode` | file write (its `mcp add` is an interactive-only wizard) | file write |
| OpenClaw | `openclaw` | `openclaw mcp add --no-probe` | `openclaw mcp unset` |
| Warp | `warp` | file write to `~/.warp/.mcp.json` (no registration CLI) | file write |
| Hermes Agent | `hermes` | `hermes mcp add` (stdin-piped confirm + post-add verify) | `hermes mcp remove` |

All entries are **user-global**, so the server is available in every project.

## Design notes

- **Client CLIs are spawned safely:** stdin closed (a prompting CLI EOF-errors instead of hanging agent runs) + a 15s hard timeout. Windows `.cmd`/`.bat` shims are launched through `cmd.exe /c`.
- **Hermes hardening:** its `mcp add` ends in a confirm prompt and **exits 0 on cancel without saving** — so the affirmative answer is piped to stdin and the config is re-read after add; a silent no-op becomes `mcp_client_cli_failed`. Its YAML config is read-only to us (both mutations go through its CLI), so hand-written YAML is never rewritten.
- **Reads stay ours:** `list`, `doctor`, and the uninstall picker parse config files directly (JSON/TOML/YAML codecs, incl. OpenClaw's nested `mcp.servers` map). Legacy descriptor shapes (`{url}`, `{serverUrl}`, `mcp-remote` args, `clerk mcp run --url`) still round-trip.
- **Agent mode throughout:** non-TTY runs auto-emit JSON, never prompt, and every registration-blocking error carries a machine-readable `code` plus a `docsUrl` to the [Clerk MCP setup docs](https://clerk.com/docs/guides/ai/mcp/clerk-mcp-server) (agent mode gets the `.md` variant) for the manual fallback.
- **Partial failure is structural:** per-client failures land in a `failures: [{ client, error }]` array in install/uninstall/list JSON; the command exits non-zero only when every targeted client fails.
- **Auth (current limitation):** the bridge is transport-only. A 401/403 on the initial handshake exits with a clear error; on later requests it's surfaced to the editor as a per-request JSON-RPC error without killing the bridge.

## Test plan

- [x] `bun run format:check`, `bun run lint`, `bun run typecheck` — green
- [x] `bun run test` — 2095 tests green, including per-client argv contracts for all six CLI dialects, file shapes for the four file-backed clients, nested-topKey + YAML codec units, corrupt-config failure threading, Hermes verify-add, and Windows `.cmd`-shim spawning
- [x] Live verification (macOS): every CLI dialect verified against the real client CLI — `claude`/`gemini`/`codex`/`code` (duplicate-add/remove semantics), `hermes` (prompt + lying exit code discovered this way), `openclaw` (sandboxed `OPENCLAW_STATE_DIR`)
- [x] Real round-trip: `uninstall --all` → `install --all` across all detected clients with zero failures; opencode's own CLI reports `✓ clerk connected` and Hermes reports `clerk ✓ enabled`; bridge handshake against the hosted server returns `Clerk MCP Server`


rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

rafa-thayto:
!snapshot

wyattjoh (commented):
Solid, well-tested feature surface. The error handling, URL hardening, and per-client config abstraction are all carefully done. A few non-blocking observations below, mostly around concurrency edge cases in the stdio bridge and the shared `~/.claude.json` write path.

packages/cli-core/src/commands/mcp/clients/claude.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const claudeClient = makeJsonClient({
+  id: "claude",
+  displayName: "Claude Code",
+  scope: "user",
+  activation: "Restart Claude Code, then run `/mcp` to connect (`clerk` must be on your PATH).",
+  topKey: "mcpServers",
+  encode: (url) => ({ command: RUN_COMMAND, args: clerkRunArgs(url) }),
+  extractUrl: (d) => withLegacyUrl(d),
+  configPath: () => userPath(".claude.json"),
```
wyattjoh, line 22: `configPath: () => userPath(".claude.json")` targets Claude Code's primary state file, which Claude Code itself writes to frequently (project history, etc.) and which can be large. `writeJsonConfig` does a full read-modify-write with `JSON.stringify(config, null, 2)`, dropping any keys Claude Code added between our read and write if it writes concurrently, and reformatting the entire file. There is no locking or atomic rename. Consider writing via a temp file + atomic rename to at least avoid a torn/partial file, and document that a running Claude Code instance could lose interleaved writes. At minimum this risk is worth a comment so the next maintainer knows the blast radius of a write here is the user's whole Claude config, not just the `clerk` entry.

packages/cli-core/src/commands/mcp/run.ts
```diff
@@ -0,0 +1,343 @@
...
+
+  const session: Session = {};
+
+  // Serialize stdout writes: concurrent SSE drains and the server→client stream
+  // all emit, and a frame must never interleave with another.
+  let writeTail: Promise<void> = Promise.resolve();
+  const emit: Emit = (message) => {
+    captureProtocolVersion(message, session);
+    const line = JSON.stringify(message) + "\n";
+    writeTail = writeTail.then(() => writeRaw(line));
```
wyattjoh: `writeTail = writeTail.then(() => writeRaw(line))` serializes stdout writes by chaining onto a single promise. If `writeRaw` ever rejects (e.g. EPIPE when the parent editor closes the pipe), the rejection propagates to `writeTail` and every subsequent `emit` returns that same rejected promise, so no further frames are written and the awaited `writeTail` in the `finally` rejects. A `.catch` on the chained write (logging to stderr and continuing) would keep the bridge resilient to a transient write failure instead of wedging all output.

packages/cli-core/src/commands/mcp/run.ts
```diff
@@ -0,0 +1,343 @@
...
+  const inflight = new Set<Promise<void>>();
+  const track = (work: Promise<void>): void => {
+    const p = suppress(work).finally(() => inflight.delete(p));
+    inflight.add(p);
+  };
+
+  let serverStream: Promise<void> | undefined;
+  const ensureServerStream = (): void => {
+    if (serverStream) return;
+    serverStream = suppress(listenForServerMessages(url, session, emitPayload, abort.signal));
```
wyattjoh, line 93: On the `initialize` response the session id is read synchronously and `ensureServerStream()` fires, but the response body (carrying `result.protocolVersion`) is drained via `track(forwardJsonBody(...))` concurrently. `captureProtocolVersion` therefore runs asynchronously, after `dispatch` returns and the main loop may already be dispatching the next stdin frame. In the normal client flow (client awaits the initialize reply before sending `initialized`) this is fine, but a client that pipelines could send its next request before `session.protocolVersion` is set, omitting the `MCP-Protocol-Version` header. Worth a one-line note that ordering relies on the client awaiting the initialize reply, or capture the protocol version before forwarding the initialize body.

packages/cli-core/src/lib/input-json.ts
```diff
@@ -128,6 +128,38 @@ function hasStdinPipe(): boolean {
...
+// the `mcp run` detection below.
+const GLOBAL_FLAGS_WITH_VALUE = new Set(["--mode", "--input-json"]);
+
+function ownsRawStdin(argv: string[]): boolean {
+  // Drop flags (and the values of known global value-flags) so what remains is
+  // the runtime prefix followed by the command path. Subcommand option values
+  // like `--name mcp` stay attached to their own command and never get mistaken
+  // for the `mcp run` command itself.
+  const positionals: string[] = [];
+  for (let i = 0; i < argv.length; i++) {
```
wyattjoh: `ownsRawStdin` hand-maintains `GLOBAL_FLAGS_WITH_VALUE = new Set(["--mode", "--input-json"])` and the comment correctly warns it must track the root options in cli-program.ts. If a new global value-flag is added there and not mirrored here, its value token leaks into the positionals and could cause `clerk mcp run` detection to misfire (or fail to fire). This is a latent maintenance trap with no test or assertion linking the two lists. Consider deriving the set from the program's registered global options, or adding a test that fails if a root value-option is missing from this set.

packages/cli-core/src/commands/mcp/clients/toml-config.ts
```diff
@@ -0,0 +1,48 @@
...
+    throw new CliError(`Could not parse ${path} as TOML: ${errorMessage(error)}`, {
+      code: ERROR_CODE.MCP_CLIENT_CONFIG_INVALID,
+    });
+  }
+}
+
+export async function writeTomlConfig(path: string, config: ConfigRecord): Promise<void> {
+  log.debug(`mcp: write ${path}`);
+  await mkdir(dirname(path), { recursive: true, mode: 0o700 });
+  await Bun.write(path, stringify(config) + "\n");
```
wyattjoh: Codex config writes should get the same atomic-write treatment as the JSON clients. This direct `Bun.write` can truncate or partially write `~/.codex/config.toml` if the CLI is interrupted mid-write, and this is a user-global config file. Write to a sibling temp file, rename it into place, clean up the temp file on failure, then call `restrictPermissions(path)`.

packages/cli-core/src/commands/mcp/shared.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+import { getMcpUrl } from "../../lib/environment.ts";
+import { CliError, ERROR_CODE, errorMessage, throwUsageError } from "../../lib/errors.ts";
+import { log } from "../../lib/log.ts";
+import { isAgent } from "../../mode.ts";
+import { CLIENT_ALIASES, CLIENT_IDS, CLIENTS, detectInstalledClients } from "./clients/registry.ts";
+import type { ClientId, McpClient } from "./clients/types.ts";
+
+export type McpOptions = {
```
wyattjoh, line 14: Project TypeScript conventions require JSDoc on public exports. This module adds several exported types and helpers without per-export docs, which makes the new MCP API surface drift from the local convention. Either document the public exports with multi-line JSDoc or keep test-only helpers private and cover them through command/client APIs.

packages/cli-core/src/commands/mcp/install.ts
```diff
@@ -0,0 +1,101 @@
...
+// (and sign in, if the server requires it). Surface that for every client we
+// just wrote, so "added" doesn't read as "done and working".
+function installNextSteps(settled: ClientUpsert[]): string[] {
+  const activated = settled.filter(
+    ({ result }) => result.status === "added" || result.status === "updated",
+  );
+  if (activated.length === 0) return [];
+  return [
+    ...activated.map(({ client }) => `${client.displayName}: ${client.activation}`),
+    "If the server requires authentication, your editor opens a browser to sign in on first connect.",
```
wyattjoh: The install next step promises that the editor will open a browser to sign in when auth is required, but `clerk mcp run` is transport-only today and throws on a pre-session `401` or `403`. This will send users down the wrong path if they hit an auth-required server. Either implement auth in the bridge first, or change this copy and the README auth note to say auth-required servers are not supported yet.

packages/cli-core/src/commands/mcp/clients/clerk-run.test.ts
```diff
@@ -0,0 +1,38 @@
+import { describe, expect, test } from "bun:test";
+import { clerkRunArgs, extractClerkRunUrl, RUN_COMMAND } from "./clerk-run.ts";
+
+const URL = "https://mcp.clerk.com/mcp";
```
wyattjoh: should be exported by the code itself from a const, not redeclared here.

packages/cli-core/src/commands/mcp/clients/clerk-run.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe("clerk-run descriptor", () => {
+  test("clerkRunArgs builds the run invocation", () => {
+    expect(clerkRunArgs(URL)).toEqual(["mcp", "run", "--url", URL]);
```
wyattjoh: I kind of worry that the `--url` option will confuse users. Unlike the env vars which are very "clerk internal dev focused", the `--url` implies that our users could want to change it to something else, which doesn't make a lot of sense I think.

packages/cli-core/src/commands/mcp/clients/clerk-run.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/** The binary clients spawn. Must be on the user's PATH. */
+export const RUN_COMMAND = "clerk";
+
+export function clerkRunArgs(url: string): string[] {
+  return ["mcp", "run", "--url", url];
+}
+
+/** Recover the URL from a `clerk mcp run --url <url>` descriptor, or undefined. */
+export function extractClerkRunUrl(descriptor: unknown): string | undefined {
```
wyattjoh: This seems very fragile and doesn't support `bunx clerk@latest` which may be preferred. Another reason why dropping the `--url` may be a good idea.

packages/cli-core/src/commands/mcp/clients/gemini.ts
```diff
@@ -0,0 +1,38 @@
...
+  command: string;
+  args: [string, string, string, ...string[]];
+}
+
+// Match the exact legacy shape `{command: "npx", args: ["-y", "mcp-remote", <url>]}`.
+// Permissive matching would round-trip unrelated stdio bridges as Clerk entries.
+function isMcpRemoteEntry(value: unknown): value is McpRemoteEntry {
+  if (typeof value !== "object" || value === null) return false;
+  const candidate = value as { command?: unknown; args?: unknown };
+  if (candidate.command !== "npx") return false;
```
wyattjoh: why only npx? I don't quite understand the use case here for the auto-migration, seems fragile, maybe we just match on the url instead?

packages/cli-core/src/commands/mcp/index.ts
```diff
@@ -0,0 +1,113 @@
...
+    .setExamples([
+      { command: "clerk mcp install", description: "Install into all detected MCP clients" },
+      {
+        command: "clerk mcp install --client claude",
+        description: "Install into Claude Code only",
+      },
+      { command: "clerk mcp list", description: "Show registered Clerk entries" },
+      { command: "clerk mcp uninstall", description: "Remove the Clerk entry from all clients" },
+      {
+        command: "clerk mcp run --url https://mcp.clerk.com/mcp",
```
wyattjoh: This example `clerk mcp run --url https://mcp.clerk.com/mcp` is shown in `--help`, but `--url` was dropped from the public CLI — `mcp run` now registers only `--name`. Commander rejects unknown options, so anyone copying this example gets `error: unknown option '--url'`. Change it to just `clerk mcp run` (the URL resolves from `CLERK_MCP_URL` / the active env profile).

packages/cli-core/src/commands/mcp/run.ts
```diff
@@ -0,0 +1,351 @@
...
+  }
+
+  if (response.status === 401 || response.status === 403) {
+    // Before a session exists the server needs up-front auth we can't provide —
+    // fail clearly. After `initialize`, a scoped 401 is a per-request error, not
+    // a reason to kill the bridge and every other in-flight request with it.
+    if (session.id === undefined) {
+      throw new CliError(
+        `The MCP server at ${url} requires authentication, which \`clerk mcp run\` does not yet provide. ` +
+          "Point --url at a server that doesn't require auth, or wait for a release with built-in sign-in.",
```
wyattjoh: The pre-session 401/403 error says "Point `--url` at a server that doesn't require auth…", but `--url` is no longer accepted by the CLI, so this guidance is unactionable. Reword to reference the real override, e.g. "Set `CLERK_MCP_URL` to a server that doesn't require auth, or wait for a release with built-in sign-in."

packages/cli-core/src/commands/mcp/README.md
```diff
@@ -0,0 +1,139 @@
...
+Codex is the one TOML-backed client (`[mcp_servers.<name>]`); rewriting
+`config.toml` does not preserve comments.
+
+## How clients connect (the stdio bridge)
+
+Every client installs the same stdio descriptor — it launches `clerk mcp run
+--url <url>` rather than pointing the editor at the remote URL directly:
+
+```jsonc
+{ "command": "clerk", "args": ["mcp", "run", "--url", "https://mcp.clerk.com/mcp"] }
```
wyattjoh: The README still advertises `--url <url>` as an install/run flag (lines 8, 47, 78, 109, 139) and shows the installed descriptor as `{ "command": "clerk", "args": ["mcp", "run", "--url", "https://mcp.clerk.com/mcp"] }` here on line 50. But the code now stores URL-less args (`clerk-run.ts` `clerkRunArgs()` returns `["mcp", "run"]`), so this example is wrong on both the flag and the stored shape. Strip `--url` from the option tables, fix this descriptor to `["mcp", "run"]`, and drop the `--url >` segment from the documented resolution order. Per .claude/rules/commands.md the README must track actual options.

.changeset/mcp-install.md
```diff
@@ -0,0 +1,5 @@
+---
+"clerk": minor
+---
+
+Add `clerk mcp install`, `list`, `uninstall`, and `run` to connect the Clerk remote MCP server (`https://mcp.clerk.com/mcp`) to Claude Code, Cursor, GitHub Copilot (VS Code; `--client vscode` or `--client copilot`), Windsurf, Gemini, and Codex. Each client is configured to launch `clerk mcp run` — a built-in stdio bridge that forwards the editor's stdio JSON-RPC to the remote server over HTTP (the job `npx mcp-remote` did, now with no npx dependency), so `clerk` must be on your PATH. Entries are written to each client's user-global config (e.g. `~/.claude.json`, `~/.cursor/mcp.json`, `~/.codex/config.toml`), available across every project regardless of where you run the CLI. `clerk doctor` gains an MCP reachability check that probes each configured server via the MCP `initialize` handshake when an entry is installed. The URL resolves in order: `--url` > the `CLERK_MCP_URL` override (for local worker development) > the active env profile's `mcpUrl` field > the hosted server, so `clerk mcp install` works with no flags. The bridge is transport-only for now; against an auth-required server it surfaces a clear error rather than signing in.
```
wyattjoh: The changelog entry states the URL resolves in order `--url > CLERK_MCP_URL > ...`. Since `--url` is gone from the public CLI, this published changelog line is inaccurate. Drop the `--url >` segment so the documented order is `CLERK_MCP_URL > active env profile mcpUrl > hosted server`. (The minor bump itself is correct for a new feature.)

packages/cli-core/src/commands/mcp/run.ts
```diff
@@ -0,0 +1,351 @@
...
+  // never blocks the next request — head-of-line free.
+  const inflight = new Set<Promise<void>>();
+  const track = (work: Promise<void>): void => {
+    const p = suppress(work).finally(() => inflight.delete(p));
+    inflight.add(p);
+  };
+
+  let serverStream: Promise<void> | undefined;
+  const ensureServerStream = (): void => {
+    if (serverStream) return;
```
wyattjoh, line 92: `ensureServerStream()` guards with `if (serverStream) return` and never resets `serverStream`. The 404 handler (around line 169) deliberately clears `session.id`/`session.protocolVersion` to force the client to re-initialize, but leaves `serverStream` holding the old, resolved, expired-session promise. When the client re-initializes and a new session id arrives, `ensureServerStream()` short-circuits, so the new session never reopens its server→client GET/SSE channel — any server-initiated notifications after a session expiry are silently dropped for the rest of the process. Fix: set `serverStream = undefined` in the 404 handler alongside the session reset (the `suppress(...)` wrapper already makes a second GET safe).

packages/cli-core/src/commands/mcp/clients/json-config.ts
```diff
@@ -0,0 +1,98 @@
...
+    return parsed as ConfigRecord;
+  } catch (error) {
+    if (error instanceof CliError) throw error;
+    throw new CliError(`Could not parse ${path} as JSON: ${errorMessage(error)}`, {
+      code: ERROR_CODE.MCP_CLIENT_CONFIG_INVALID,
+    });
+  }
+}
+
+export async function writeJsonConfig(path: string, config: ConfigRecord): Promise<void> {
```
wyattjoh, line 80: `upsert`/`remove` read the entire config, mutate in memory, then rename a full rewrite over the original. The atomic temp+rename prevents a torn read but not a lost update: if a running client rewrites the file (e.g. Claude Code updates `~/.claude.json` frequently) between our read and our rename, our rename clobbers that write with stale data. This matches what `claude mcp add` itself does, so it's an accepted tradeoff — not a blocker — but the blast radius is another tool's primary state file. Worth documenting (README and/or the per-client `activation` copy) that install/uninstall is safest with the target client closed, or re-reading and merging immediately before the rename.

packages/cli-core/src/commands/mcp/install.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export async function mcpInstall(options: McpOptions = {}): Promise<void> {
+  const url = resolveUrl(options);
+  const name = resolveName(options);
+  const cwd = process.cwd();
+  const clients = await chooseClients(options, cwd);
+  const force = Boolean(options.force);
+  const json = wantsJson(options);
+
+  if (clients.length === 0) {
```
wyattjoh, line 72: When no clients are detected, the interactive picker returns `[]` and install prints "No MCP clients selected." — but nothing was deselected, none were detected. The non-interactive path throws the much more helpful `MCP_NO_CLIENT_DETECTED`, which lists the supported clients and the `--client` escape hatch. Consider surfacing that same guidance for the interactive zero-detection case so the two paths don't diverge for the same underlying state.

packages/cli-core/src/commands/mcp/run.ts
```diff
@@ -0,0 +1,351 @@
...
+    serverStream = suppress(listenForServerMessages(url, session, emitPayload, abort.signal));
+  };
+
+  try {
+    for await (const message of readJsonRpcLines(input, MAX_LINE_BYTES)) {
```
wyattjoh, line 100: The read loop `await dispatch(...)`s each frame, and `dispatch` awaits the POST through response headers before the next stdin frame is pulled. Response bodies drain concurrently via `track()`, but a slow plain `application/json` response blocks sending any following frame — including a `notifications/cancelled` meant to abort that very operation. SSE responses return headers promptly and are fine. Either `track()` the dispatch after `initialize` so independent requests pipeline, or soften the comment to say only response bodies are non-blocking.

packages/cli-core/src/commands/mcp/shared.ts
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Resolve the MCP server URL from options or environment (`CLERK_MCP_URL` >
+ * active env profile > hosted default), validating scheme and rejecting
+ * embedded credentials.
+ */
+export function resolveUrl(options: McpOptions): string {
+  // `getMcpUrl()` always resolves to a usable URL (Clerk's hosted server by
+  // default), so the only failure mode left is a `CLERK_MCP_URL` value that
+  // is malformed or uses a non-network scheme.
+  const candidate = options.url ?? getMcpUrl();
```
wyattjoh, line 35: `McpOptions.url` and the `options.url ??` branch in `resolveUrl` are now unreachable from the CLI (no command registers `--url`). The test `"targets the --url value"` (run.test.ts:331) passes `{ url }` directly into `mcpRun`, exercising a path no real invocation can reach, under a name implying a flag exists. Either re-introduce a hidden `--url` and wire it through `index.ts`, or remove `url` from `McpOptions`/`resolveUrl` and rename the test to reflect env/profile resolution — pick one so the option's status is unambiguous.

wyattjoh (commented):
Strong, well-tested PR. The main cleanup is leftover `--url` surface from the "drop --url from public CLI" commit (a help example that now errors, an unactionable error message, the README, and the changeset), plus a couple of narrow correctness edge cases noted inline.

packages/cli-core/src/commands/doctor/check-mcp.ts
```diff
@@ -0,0 +1,84 @@
...
+ * doesn't import `mcp/shared.ts` (env profiles, prompts) and the module cycle
+ * that comes with it. Imports only the light `collect`/`probe` helpers.
+ */
+
+import { collectEntries } from "../mcp/collect.ts";
+import { probeMcp, type McpProbeResult } from "../mcp/probe.ts";
+import type { ListEntry } from "../mcp/clients/types.ts";
+import type { CheckResult } from "./types.ts";
+
+const NAME = "MCP server";
```
wyattjoh: seems like for only being used twice that we don't need this!

packages/cli-core/src/commands/mcp/clients/clerk-run.ts
```diff
@@ -0,0 +1,82 @@
...
+ * list/uninstall.
+ *
+ * Priority:
+ * 1. Current format: `{ command: "clerk", args: ["mcp", "run"] }` — no URL in
+ *    args; resolves to `getMcpUrl()` so list/upsert see a comparable URL.
+ * 2. Legacy v1 format: `{ command: "clerk", args: ["mcp", "run", "--url", …] }` —
+ *    URL extracted from the `--url` arg.
+ * 3. Legacy v0 format: `{ url }` or `{ serverUrl }` — plain key lookup.
+ */
+export function withLegacyUrl(
```
wyattjoh: Maybe as this hasn't released yet we can remove this?

packages/cli-core/src/commands/mcp/clients/make-cli-client.ts
```diff
@@ -0,0 +1,142 @@
...
+   * case the CLI (which owns the format) gets the final say.
+   */
+  async function presenceOf(name: string, cwd: string): Promise<"present" | "absent" | "unknown"> {
+    try {
+      const entries = await base.list(cwd);
+      return entries.some((entry) => entry.name === name) ? "present" : "absent";
+    } catch {
+      return "unknown";
+    }
+  }
```
wyattjoh: could be pulled out of this cli client constructor I think!

packages/cli-core/src/commands/mcp/clients/make-client.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+function makeFileClient(spec: FileClientSpec, codec: ConfigCodec): McpClient {
+  const topKeyPath: readonly [string, ...string[]] =
+    typeof spec.topKey === "string" ? [spec.topKey] : spec.topKey;
+
+  /** Walk the key path, validating each level is an object (or absent → `{}`). */
+  function serversIn(config: ConfigRecord, configPath: string): Record<string, unknown> {
+    let node: Record<string, unknown> = config;
+    for (const key of topKeyPath) node = getServerMap(node, key, configPath);
```
wyattjoh: nit: Not really a fan of loops like this that don't wrap with `{ ... }`, maybe worth adding a lint rule for if others agree

packages/cli-core/src/commands/mcp/probe.ts
```diff
@@ -0,0 +1,107 @@
...
+const INITIALIZE_REQUEST = {
+  jsonrpc: "2.0",
+  id: 1,
+  method: "initialize",
+  params: {
+    protocolVersion: "2024-11-05",
+    capabilities: {},
+    clientInfo: { name: "clerk-cli", version: resolveCliVersion() ?? DEV_CLI_VERSION },
+  },
+};
```
wyattjoh: Would be amazing if we could add a dev-dep for this that would give us the mcp server types to assert against. Wouldn't be bundled or used internally, but would give us a hard TS gate to protect things.

packages/cli-core/src/commands/mcp/README.md
```diff
@@ -0,0 +1,223 @@
...
+[clerk/cloudflare-workers/workers/remote-mcp-server](https://github.com/clerk/cloudflare-workers/tree/main/workers/remote-mcp-server)).
+These subcommands register, list, and remove the Clerk entry per client, and
+probe the server via `clerk doctor`. Clients that ship a **non-interactive** MCP
+registration CLI (Claude Code, Gemini, Codex, VS Code, OpenClaw, Hermes) are
+registered by shelling out to it — the client owns its config format and write
+safety; for the rest (Cursor, Windsurf, Warp, opencode) we write the config
+file directly. opencode does ship an `mcp add` command, but it is an
+interactive wizard with no flag-driven stdio path (and no remove command), so
+it counts as file-backed. Reads (`list`, `doctor`, the uninstall picker)
+always parse the config files directly. The URL is resolved in order: the `CLERK_MCP_URL` environment
```
wyattjoh: Think the internal/dev comments related to it's usage should be in it's own section near the bottom instead of in the preamble (which here is already pretty long)

packages/cli-core/src/commands/mcp/run.test.ts
wyattjoh, line 1: Would strongly suggest we import/use the `@modelcontextprotocol/sdk` here so we can have typed tests and/or ones that integrate with the actual SDK to validate conformance.

packages/cli-core/src/commands/mcp/clients/make-client.ts
```diff
@@ -0,0 +1,223 @@
...
+      // uninstall's picker) settle per client, so one bad config warns there
+      // without sinking the other clients — and `doctor` can tell "unreadable
+      // config" apart from "no entries" instead of reporting a clean pass.
+      const config = await codec.read(configPath);
+      const servers = serversIn(config, configPath);
+      const entries: ListEntry[] = [];
+      for (const [name, descriptor] of Object.entries(servers)) {
+        const url = spec.extractUrl(descriptor);
+        if (!url) continue;
+        if (name === "clerk" || isClerkUrl(url)) {
```
wyattjoh: this only treats an entry as ours when it's named `clerk` or resolves to a clerk.com host. if someone installs with `--name foo` while `CLERK_MCP_URL` points at localhost (our local-dev flow), the entry falls out of list/doctor and can't be removed on the cli-backed clients. could we match on the descriptor shape (`isClerkRunEntry`) instead, so any `clerk mcp run` bridge is recognized regardless of name or resolved url?

packages/cli-core/src/commands/mcp/clients/make-cli-client.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    async remove(name: string, cwd: string): Promise<RemoveResult> {
+      if (!spec.removeArgs) return base.remove(name, cwd);
+      const configPath = base.configPath(cwd);
+      if ((await presenceOf(name, cwd)) === "absent") {
+        return { client: base.id, configPath, removed: false };
+      }
+      const bin = requireBinary();
+      await runOrThrow([bin, ...spec.removeArgs(name)], "remove the MCP entry");
+      return { client: base.id, configPath, removed: true };
```
wyattjoh, line 153: we return `removed: true` whenever the client cli exits 0, even if it no-op'd because the entry wasn't there. should we re-check presence afterwards (like `verifyAdd` does on the add side) so we don't report a removal that didn't actually happen?

packages/cli-core/src/commands/mcp/shared.ts
```diff
@@ -0,0 +1,188 @@
...
+    const client = clients[i]!;
+    if (outcome.status === "fulfilled") {
+      succeeded.push({ client, result: outcome.value });
+      continue;
+    }
+    errors.push(outcome.reason);
+    failed.push({ client: client.id, error: errorMessage(outcome.reason) });
+    log.warn(`${client.displayName}: ${errorMessage(outcome.reason)}`);
+  }
+  if (succeeded.length === 0 && errors.length > 0) throw errors[0];
```
wyattjoh: when every client fails this throws past the json output, so `--json` consumers get the error envelope instead of `{ results, failures }` — which is exactly the case where `failures` is most useful. could we emit the envelope and set the exit code separately instead of throwing here?

packages/cli-core/src/commands/mcp/install.ts
```diff
@@ -0,0 +1,96 @@
...
+  wantsJson,
+  type McpOptions,
+} from "./shared.ts";
+import { detectInstalledClients } from "./clients/registry.ts";
+import type { McpClient, UpsertResult } from "./clients/types.ts";
+
+async function chooseClients(options: McpOptions, cwd: string): Promise<McpClient[]> {
+  // `wantsJson` covers `--json` and agent mode, so non-interactive callers never
+  // block on the picker.
+  if (options.client?.length || options.all || wantsJson(options)) {
```
wyattjoh: folding `wantsJson` in here means `--json` with no `--client` installs into every detected client with no picker. `--json` reads as "same action, just different output" to me — should the install-all path only kick in for `isAgent()`, and keep the picker when we're on a tty?

packages/cli-core/src/commands/mcp/uninstall.ts
```diff
@@ -0,0 +1,109 @@
...
+export async function mcpUninstall(options: McpOptions = {}): Promise<void> {
+  const name = resolveName(options);
+  const cwd = process.cwd();
+  const json = wantsJson(options);
+  const explicit =
+    options.client && options.client.length > 0 ? resolveClients(options.client) : undefined;
+
+  // Non-interactive: explicit `--client`, `--all`, agent mode, or `--json`
+  // operate directly on the targeted clients.
+  if (json || explicit || options.all) {
```
wyattjoh: same thing as the install side, but riskier here — `uninstall --json` removes from all 10 clients with no picker or confirm, while the README only documents `--json` as an output format. can we gate the all-clients path on agent mode, so adding `--json` to inspect machine output doesn't wipe every client?

packages/cli-core/src/commands/mcp/clients/make-client.ts
```diff
@@ -0,0 +1,223 @@
...
+  ClientId,
+  ListEntry,
+  McpClient,
+  McpServerEntry,
+  RemoveResult,
+  Scope,
+  UpsertResult,
+} from "./types.ts";
+
+export function hasStringProp<K extends string>(
```
wyattjoh: nit: `hasStringProp` is a generic guard like `isRecord` — feels like it belongs over in `lib/objects.ts` rather than living in here.

packages/cli-core/src/commands/mcp/clients/toml-config.ts
```diff
@@ -0,0 +1,40 @@
...
+ * CLI handles both add and remove, so no code path needs a TOML write — and
+ * re-serializing would destroy the comments and formatting in a user's
+ * hand-maintained `config.toml`.
+ */
+
+import { isRecord } from "../../../lib/objects.ts";
+import { CliError, ERROR_CODE, errorMessage } from "../../../lib/errors.ts";
+import { readConfigText, refuseConfigWrite, type ConfigRecord } from "./json-config.ts";
+
+export async function readTomlConfig(path: string): Promise<ConfigRecord> {
```
wyattjoh, line 15: nit: now that the write half is shared via `refuseConfigWrite`, this read half is still basically identical to `readYamlConfig` — could fold the two into one read-only codec factory if we feel like tidying it.

packages/cli-core/src/commands/mcp/probe.ts
```diff
@@ -0,0 +1,107 @@
...
+  } catch {
+    return undefined;
+  }
+}
+
+// The streamable-HTTP transport answers `initialize` as either application/json
+// or a text/event-stream frame (`event: message\ndata: {…}`). Pull the JSON-RPC
+// payload out of whichever the server returned. For SSE, reassemble the first
+// event's `data:` lines (the spec allows a payload to span several).
+function parseHandshake(contentType: string, body: string): unknown {
```
wyattjoh, line 57: nit: this `data:` reassembly is duplicated in `run.ts`'s `emitEvent` — might be worth a small shared `parseSseEventData` helper both can use.

packages/cli-core/src/commands/doctor/check-mcp.ts
wyattjoh, line 1: heads up: the hosted server needs auth we don't send yet, so if it 401s the unauthenticated `initialize` probe, doctor will flag the just-installed entry as "not reachable (HTTP 401)" right after install. worth confirming against the live server so we're not false-warning.

packages/cli-core/src/commands/mcp/clients/json-config.ts
wyattjoh, line 1: one thing worth flagging: since we JSON.parse + re-serialize these, any comments or custom formatting in a hand-edited config get dropped on write, and a JSONC file (with comments) fails to read at all. probably fine given the client owns the file, but worth a note.

packages/cli-core/src/commands/mcp/run.test.ts
wyattjoh: Right now `run.test.ts` treats every emitted frame as `Record<string, unknown>` and only asserts it's parseable JSON:

```ts
.map((line) => JSON.parse(line) as Record<string, unknown>)
```

So the suite proves the bridge emits JSON, but not that it emits *spec-valid MCP*. A frame missing `jsonrpc: "2.0"`, or an `initialize` result with no `protocolVersion`, passes today. That's exactly the invariant a passthrough proxy should be guarding.

I'd add `@modelcontextprotocol/sdk` (current stable `^1.29.0`) as a **devDependency** only. The bridge itself stays dependency-free by design; this is test-side. Two additive tiers:

**1. Typed wire validation (small).** Swap the bare `JSON.parse` in `framesFrom` for the SDK's Zod schema, so every existing test doubles as a conformance check and the fixtures get real types:

```ts
import { JSONRPCMessageSchema } from "@modelcontextprotocol/sdk/types.js";

// .parse throws if the bridge wrote anything that isn't a valid JSON-RPC frame
.map((line) => JSONRPCMessageSchema.parse(JSON.parse(line)))
```

Result payloads can go a step further (`ListToolsResultSchema`, `InitializeResultSchema`) where we assert on them.

**2. Drive a real SDK `Client` through the bridge (conformance).** `mcpRun(options, streams)` already accepts injected `input`/`write` streams, which is the seam we need. A small `Transport` (just `start`/`send`/`close` + `onmessage`) can bind a real `Client` to the real bridge in-process, over the existing `stubFetch` upstream, no child process:

```ts
import { Client } from "@modelcontextprotocol/sdk/client/index.js";

class BridgeTransport implements Transport {
  onmessage?: (m: JSONRPCMessage) => void;
  // queue feeds mcpRun's `input`; the bridge's `write` calls back into onmessage
  async start() {
    void mcpRun({ url: URL }, {
      input: this.queue,
      write: (chunk) => {
        for (const line of chunk.split("\n").filter(Boolean))
          this.onmessage?.(JSONRPCMessageSchema.parse(JSON.parse(line)));
      },
    });
  }
  async send(m) { this.queue.push(JSON.stringify(m) + "\n"); }
  async close() { this.queue.end(); }
}

const client = new Client({ name: "test", version: "0" });
await client.connect(new BridgeTransport());
expect((await client.listTools()).tools[0].name).toBe("create_user");
```

If a genuine MCP client completes `initialize` + `listTools` through the bridge, that exercises session-id threading and protocol-version echo end to end, which is stronger than hand-rolled frame assertions.

I'd keep the current hand-rolled tests as-is for the adversarial cases (malformed SSE, oversized bodies, non-object frames), since those feed input a real SDK client would never send. Tier 1 is the quick win; Tier 2 as one or two focused tests on top.

Note: the `Transport` type is at `@modelcontextprotocol/sdk/shared/transport.js` in the 1.x line, and `queue` above is shorthand for a small async-pushable (resolvable-promise queue) rather than a real type.

wyattjoh (approved):

PR #309, opened 2026-05-25
refactor(cli): co-locate command registration in each command folder

rafa-thayto:
## What

Splits the command-registration wiring out of the monolithic `packages/cli-core/src/cli-program.ts` (1109 lines) into per-command `registerX(program)` functions co-located in each command's folder. `createProgram()` is now a short, ordered list of `registerX(program)` calls.

```
packages/cli-core/src/cli-program.ts | 860 +--------------------
```

This makes each command's surface (name, flags, examples, action) owned by its own folder — you add or change a command by editing `commands/<name>/index.ts`, not by editing one giant file. It follows the pattern already established by `registerExtras(program)`.

## How it works

- Each command folder exposes `export function registerX(program: Command): void` holding its own `.command()/.option()/.setExamples()/.action()` chain (mirrors the existing `registerExtras` precedent).
- New `commands/toggles/` owns the shared `enable`/`disable` parents, wiring the `orgs` + `billing` handlers under both (neither feature folder cleanly owns those parents).
- Shared option parsers (`parseIntegerOption`, `collectOptionValues`) extracted to `lib/option-parsers.ts`; the `users`-only `USER_LIST_ORDER_BY_CHOICES` moved into the `users` folder.
- Removed the now-dead `apps`/`users` aggregate exports that only existed to feed `cli-program.ts`.

## Behavior

**No user-facing change.** Registration order is preserved exactly, so `clerk --help` and every subcommand's help is unchanged. This was verified by capturing the full `--help` tree (39 command nodes) from `main` before the change and diffing it against the result after — **byte-identical**.

## Test plan

- [x] `bun run typecheck` — clean
- [x] `bun run build:compile` — compiles (336 modules)
- [x] `bun run test` — no new failures introduced (pre-existing local-env failures unchanged vs `main`; full verdict from CI)
- [x] `--help` tree byte-identical across all 39 nodes (init, auth, link, unlink, whoami, open, apps, users, env, config, enable/disable, api, doctor, switch-env, completion, skill, update, bird)
- [x] Multi-agent review (migration-fidelity + security/performance/TypeScript/code-quality) — cleared

Empty changeset included (internal refactor, no user-facing impact).


rafa-thayto:
!snapshot

packages/cli-core/src/cli-program.ts
```diff
@@ -164,779 +119,24 @@ Give AI agents better Clerk context: install the Clerk skills
...
-    .command("deploy")
-    .description("Deploy a Clerk application to production");
-  deployCmd.command("run", { isDefault: true, hidden: true }).action(deploy);
-  deployCmd
-    .command("status")
-    .description("Show production deploy status (read-only)")
-    .option("--wait", "Wait for DNS, SSL, and email DNS verification with retries")
-    .action(deployStatus);
-
+  registerInit(program);
```
wyattjoh: Maybe we can put these in an array and iterate over it instead? that way we have a common/standard registration loop and just a static array of registrants?

packages/cli-core/src/lib/option-parsers.ts
```diff
@@ -0,0 +1,25 @@
+import { throwUsageError } from "./errors.ts";
+
+/** Commander option reducer: accumulate repeated `--flag value` occurrences into an array. */
+export function collectOptionValues(value: string, previous: string[] = []): string[] {
+  return [...previous, value];
+}
+
+/** Parse and range-validate an integer option value, throwing a usage error on bad input. */
+export function parseIntegerOption(
```
wyattjoh, line 9: Silly, but can we add some unit testing for this? Just want to make sure it's covered.

packages/cli-core/src/cli-program.ts
```diff
@@ -155,750 +133,9 @@ export function createProgram() {
...
-    .description("Deploy a Clerk application to production");
-  deployCmd.command("run", { isDefault: true, hidden: true }).action(deploy);
-  deployCmd
-    .command("status")
-    .description("Show production deploy status (read-only)")
-    .option("--wait", "Wait for DNS, SSL, and email DNS verification with retries")
-    .action(deployStatus);
-
-  registerExtras(program);
+  for (const register of registrants) {
```
wyattjoh, line 142: I'd suggest that we export something like:

```ts
type Program = Command<[], { tester?: boolean, inputJson?: string, mode?: string, verbose?: boolean }>
```

from this file so that we can adjust the types on each of the registrants to accept an argument of type `Program` and not generic `Command` so that they can enforce the requirement/types for the global options.

packages/cli-core/src/cli-program.ts
```diff
@@ -43,50 +41,36 @@ import { clerkHelpConfig } from "./lib/help.ts";
...
+  registerDoctor,
+  registerSwitchEnv,
+  registerCompletion,
+  registerUpdate,
+  registerDeploy,
+  registerExtras,
+];
 
 export function createProgram() {
   const program = new Command()
```
wyattjoh, line 76: Should require that this conforms to the Program type

packages/cli-core/src/cli-program.ts
rafa-thayto: Done — refactored to use a `registrants` array iterated with a `for...of` loop.

packages/cli-core/src/lib/option-parsers.ts
rafa-thayto: Done — added `option-parsers.test.ts` with `test.each` coverage for both `collectOptionValues` and `parseIntegerOption` (valid inputs, non-integer rejection, and range validation).

packages/cli-core/src/cli-program.ts
rafa-thayto: Done — exported `Program = Command<[], { inputJson?: string; mode?: string; verbose?: boolean }>` (omitting `tester` since there's no `--tester` global flag). All registrant functions already accept `Program` instead of generic `Command`.

packages/cli-core/src/cli-program.ts
rafa-thayto: Done — added `as Program` cast at the end of the option chain so the `program` variable is typed as `Program` from the point of construction, not just validated at the `return` statement.

wyattjoh (approved):

PR #315, opened 2026-06-01
feat: remove bundled clerk-cli skill

wyattjoh:
## What

- Remove the bundled `clerk-cli` skill files from the CLI package.
- Remove the standalone `clerk skill install` command and its help-tip detection path.
- Keep `clerk init` installing agent context by requesting `clerk-cli` from `clerk/skills` alongside the default upstream skills.
- Move the shared `skills` CLI installer helpers into `lib` for `init` and billing to use.
- Add a major changeset for the removed command surface.

## Why

The `clerk-cli` agent skill now lives in the `clerk/skills` repository, so the CLI no longer needs to ship or stage a bundled copy. `clerk init` still installs the skill as part of the upstream agent skills flow, and shared installer helpers no longer need to live under a removed command directory.

## How

The init installer now treats `clerk-cli` as part of the upstream default skill list from `clerk/skills`. The bundled-skill text imports, staging helper, install command, skill detection helper, and related docs/tests were removed. The remaining `skills add` runner helpers moved to `packages/cli-core/src/lib/skills.ts` with tests beside them.

## Verification

- `bun run format:check`
- `bun run lint`
- `bun run typecheck`
- `bun run test`
- `bun run test:e2e:op`


packages/cli-core/src/lib/skills.ts
```diff
@@ -0,0 +1,135 @@
...
+export function buildSkillsArgs(
+  source: string,
+  skillNames: readonly string[],
+  interactive: boolean,
+  copy: boolean,
+): string[] {
+  const skillFlags = skillNames.flatMap((s) => ["--skill", s]);
+  const extraFlags = interactive ? [] : ["-y", "-g"];
+  const copyFlag = copy ? ["--copy"] : [];
+  return ["skills", "add", source, ...skillFlags, ...extraFlags, ...copyFlag];
```
rafa-thayto: `copy` is dead now that the bundled-skill staging is gone. The only path that passed `copy=true` was the removed `installClerkSkillCore`, which staged the embedded skill into a temp dir and needed `--copy` to avoid symlinking into a dir it deleted right after. Both remaining callers (`init/skills.ts` and `billing/index.ts`) pass `false`, so the `--copy` branch can never run. Drop the param from `buildSkillsArgs` and `runSkillsAdd` (and the `copy=true` test case plus the `copy` line in the docstring).

```suggestion
  return ["skills", "add", source, ...skillFlags, ...extraFlags];
```

rafa-thayto (changes requested):

rafa-thayto (approved):

PR #335, opened 2026-06-12
fix(deploy): use Frontend API URL for OAuth redirect URI in walkthrough

Railly:
## Problem

The OAuth walkthrough in `clerk deploy` prints `https://accounts.{domain}/v1/oauth_callback` as the redirect URI. That endpoint lives on the Frontend API (`clerk.{domain}`), not the Account Portal, so pasting the printed value into a provider console causes `redirect_uri_mismatch`. Affects every provider walkthrough since #260.

## Fix

PLAPI already returns the correct value as `frontend_api_url` on the domain object; it was being dropped. Thread it through deploy state (new-deploy and resume paths) into `showOAuthWalkthrough`, with a `https://clerk.{domain}` fallback since TS types do not validate runtime JSON.

## Tests

- New walkthrough test with a distinctive API-provided `frontend_api_url` proves the value is threaded, not string-built from the domain
- Reconcile-path walkthrough test now asserts the FAPI URI and the absence of the `accounts.` one
- `bun test`, `typecheck`, `lint`, `format:check` pass

### DIFF
--- a/packages/cli-core/src/commands/deploy/providers.ts
+++ b/packages/cli-core/src/commands/deploy/providers.ts
@@ -439,18 +439,23 @@ export function providerSetupIntro(provider: OAuthProvider | OAuthProviderDescri
 export async function showOAuthWalkthrough(
   provider: OAuthProvider | OAuthProviderDescriptor,
   domain: string,
+  frontendApiUrl?: string,
 ): Promise<void> {
   const descriptor = providerDescriptorFromInput(provider);
   const slug = descriptor?.provider ?? (provider as OAuthProvider);
   const label = descriptor?.label ?? providerLabel(slug);
   const docsUrl = descriptor?.docsUrl ?? providerDocsUrl(slug);
+  // The OAuth callback is served by the Frontend API (clerk.<domain>), not the
+  // Account Portal (accounts.<domain>). Prefer the API-reported URL over
+  // reconstructing it from the domain.
+  const callbackBase = frontendApiUrl?.replace(/\/+$/, "") || `https://clerk.${domain}`;
 
   log.info(`\nConfigure your ${bold(label)} OAuth app with these values:\n`);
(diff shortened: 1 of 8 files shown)


packages/cli-core/src/commands/deploy/providers.ts
```diff
@@ -439,18 +439,23 @@ export function providerSetupIntro(provider: OAuthProvider | OAuthProviderDescri
...
+  frontendApiUrl?: string,
 ): Promise<void> {
   const descriptor = providerDescriptorFromInput(provider);
   const slug = descriptor?.provider ?? (provider as OAuthProvider);
   const label = descriptor?.label ?? providerLabel(slug);
   const docsUrl = descriptor?.docsUrl ?? providerDocsUrl(slug);
+  // The OAuth callback is served by the Frontend API (clerk.<domain>), not the
+  // Account Portal (accounts.<domain>). Prefer the API-reported URL over
+  // reconstructing it from the domain.
+  const callbackBase = frontendApiUrl?.replace(/\/+$/, "") || `https://clerk.${domain}`;
```
wyattjoh: Maybe a good idea to extract this and others to a shared map?

wyattjoh:
I was more sort of thinking these domain subdomains:

https://github.com/clerk/cli/blob/41ad30e0f9809b33e5f291714e5634fb2c02e919/packages/cli-core/src/commands/deploy/copy.ts#L48

That way other code can just reference it properly instead of hardcoding `accounts.` which can too easily drift.

packages/cli-core/src/commands/deploy/providers.ts
Railly: Extracted to `oauthWalkthroughUrls`, returns `{ authorizedOrigins, redirectUri }` so the walkthrough just destructures 👍

wyattjoh (approved):

PR #340, opened 2026-06-16
fix(api): keep separator between path and summary in `clerk api ls`

### LINKED ISSUE #330: `clerk api ls`: long endpoint paths run into the description column with no separator
On 1.5.0, `clerk api ls --platform config` renders long endpoint paths with the summary glued on, no separator:

```
GET     /platform/applications/{applicationID}/instances/{envOrInsID}/configGet instance config
PUT     /platform/applications/{applicationID}/instances/{envOrInsID}/configReplace instance config
PATCH   /platform/applications/{applicationID}/instances/{envOrInsID}/configUpdate instance config
GET     /platform/applications/{applicationID}/instances/{envOrInsID}/config/schemaGet instance config schema
```

Looks like the path column has a fixed width and paths longer than it overflow straight into the description. A minimum one-space separator (or wrapping the description to the next line) would fix it.

rafa-thayto:
## Summary

`clerk api ls` rendered long endpoint paths with the summary glued directly onto the path, e.g.:

```
GET     /platform/applications/{applicationID}/instances/{envOrInsID}/configGet instance config
```

`pathWidth` is capped at 50, and `String.prototype.padEnd` is a no-op once the path already meets/exceeds that width, so no separator was emitted. This guarantees a two-space separator for over-long paths while leaving normal-width alignment untouched.

## Test plan
- `bun test src/commands/api/ls.test.ts` passes
- Existing `ls` tests are spacing-agnostic (substring assertions); no snapshot to update

Fixes #330

### DIFF
--- a/packages/cli-core/src/commands/api/ls.ts
+++ b/packages/cli-core/src/commands/api/ls.ts
@@ -35,7 +35,9 @@ function printTable(endpoints: EndpointInfo[]): void {
 
   for (const ep of endpoints) {
     const method = ep.method.padEnd(methodWidth);
-    const path = ep.path.padEnd(pathWidth);
+    // padEnd is a no-op once the path meets/exceeds pathWidth, so guarantee a
+    // separator for over-long paths instead of gluing the summary onto them.
+    const path = ep.path.length >= pathWidth ? `${ep.path}  ` : ep.path.padEnd(pathWidth);
     log.data(`${method}${path}${ep.summary}`);
   }
 }
--- a/.changeset/api-ls-path-separator.md
+++ b/.changeset/api-ls-path-separator.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Keep a separator between the path and description in `clerk api ls` so long endpoint paths no longer run straight into their summaries.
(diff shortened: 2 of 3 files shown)


packages/cli-core/src/commands/api/ls.ts
```diff
@@ -35,7 +35,9 @@ function printTable(endpoints: EndpointInfo[]): void {
 
   for (const ep of endpoints) {
     const method = ep.method.padEnd(methodWidth);
-    const path = ep.path.padEnd(pathWidth);
+    // padEnd is a no-op once the path meets/exceeds pathWidth, so guarantee a
+    // separator for over-long paths instead of gluing the summary onto them.
+    const path = ep.path.length >= pathWidth ? `${ep.path}  ` : ep.path.padEnd(pathWidth);
```
wyattjoh, line 40: This is the exact case the fix addresses, but `ls.test.ts` has no coverage for it. The test spec's longest path (`/users/{user_id}`) is well under the 50-char `pathWidth` cap, so the new `ep.path.length >= pathWidth` branch never executes and a regression here would go unnoticed. Consider adding a catalog entry with a path of 50+ characters and asserting the emitted line contains at least two spaces between the path and its summary (e.g. that the summary is not glued directly onto the path).

wyattjoh (commented):
Clean, well-scoped fix. The conditional correctly guarantees a separator for paths at or beyond `pathWidth` (where `padEnd` was a no-op), and the comment accurately explains why. One optional follow-up below.

dmoerner (approved):

wyattjoh (approved):

PR #341, opened 2026-06-16
fix(input-json): only read stdin with explicit `--input-json -`

### LINKED ISSUE #333: Agent mode parses any piped stdin as --input-json, breaking shell loops and preempting explicit flags
Found while scripting a batch of `clerk config patch` calls (CLI 1.5.0, agent mode). In agent mode, any non-empty piped stdin gets parsed as the global `--input-json` options payload, even when `--input-json` was never passed. If stdin isn't valid JSON, the command fails before doing anything:

```sh
$ echo 'not json' | clerk whoami
{"error":{"code":"invalid_json","message":"Invalid JSON in --input-json. Please provide valid JSON."}}
```

Where this bites in practice is shell loops, which is the bread-and-butter pattern for scripted/agent usage:

```sh
while read -r name app ins; do
  clerk config patch --app "$app" --instance "$ins" \
    --json '{"auth_attack_protection":{"bot_protection":{"captcha_enabled":false}}}' --yes
done < worklist.tsv
```

The `clerk` process inherits the loop's stdin, swallows the rest of the work list, fails with `invalid_json` despite the explicit and valid `--json` flag, and the loop silently terminates after one iteration. Workaround that makes it behave: `</dev/null` on the `clerk` call.

Two suggestions:

- Only read stdin as the options payload when `--input-json -` is passed explicitly. Implicit stdin consumption is surprising for every command that doesn't otherwise read stdin, and it preempts explicit flags.
- If implicit parsing stays, make the error say what happened ("stdin was interpreted as --input-json") so the failure mode is debuggable; as written it points at a flag the user never typed.

Happy to provide more environment detail if useful. Aside from this, the patch workflow itself was great: dry-run diffs made a 10-instance batch remediation very comfortable.

rafa-thayto:
## Summary

In agent mode the program-level `--input-json` expander consumed **any** non-TTY stdin and parsed it as the options payload, even when `--input-json` was never passed. That broke two common patterns:

```sh
# shell loops — the clerk process swallows the rest of the worklist
while read -r app ins; do clerk config patch --app "$app" --instance "$ins" --json '…' --yes; done < worklist.tsv

# commands that read their own stdin
cat payload.json | clerk api /users   # body was pre-empted and parsed as --input-json
```

Non-JSON stdin then failed with `invalid_json` before the command ran. This makes stdin read **only** with an explicit `--input-json -`; piped stdin is otherwise left untouched.

## Test plan
- `bun test src/lib/input-json.test.ts` and `src/test/integration/input-json.test.ts` pass
- Updated unit tests assert piped stdin is ignored without `--input-json -`

Fixes #333

### DIFF
--- a/packages/cli-core/src/lib/input-json.test.ts
+++ b/packages/cli-core/src/lib/input-json.test.ts
@@ -4,8 +4,6 @@ import { join } from "node:path";
 import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 
-const originalIsTTY = process.stdin.isTTY;
-
 describe("toKebabCase", () => {
   test("converts camelCase", () => {
     expect(toKebabCase("dryRun")).toBe("dry-run");
@@ -45,15 +43,6 @@ describe("toKebabCase", () => {
 });
 
 describe("expandInputJson", () => {
-  beforeEach(() => {
-    // Ensure stdin looks like a TTY so the auto-stdin path is not triggered
-    process.stdin.isTTY = true;
-  });
-
-  afterEach(() => {
-    process.stdin.isTTY = originalIsTTY;
-  });
-
   test("returns argv unchanged when --input-json is absent", async () => {
     const argv = ["clerk", "init", "--yes"];
     const result = await expandInputJson(argv);
--- a/packages/cli-core/src/lib/input-json.test.ts
+++ b/packages/cli-core/src/lib/input-json.test.ts
@@ -325,25 +325,16 @@ describe("expandInputJson", () => {
       expect(result.result).toContain("--yes");
     });
 
-    test("auto-detects piped stdin when --input-json is absent", async () => {
+    test("ignores piped stdin when --input-json is absent", async () => {
       const result = await expandViaStdin(["clerk", "init"], '{"framework":"next","yes":true}');
-      expect(result.result).toContain("--framework");
-      expect(result.result).toContain("next");
-      expect(result.result).toContain("--yes");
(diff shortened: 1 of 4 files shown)


packages/cli-core/src/lib/input-json.test.ts
```diff
@@ -325,25 +325,16 @@ describe("expandInputJson", () => {
       expect(result.result).toContain("--yes");
     });
 
-    test("auto-detects piped stdin when --input-json is absent", async () => {
+    test("ignores piped stdin when --input-json is absent", async () => {
```
wyattjoh, line 317: Now that the implicit auto-stdin path is gone, the `beforeEach`/`afterEach` hooks that stub `process.stdin.isTTY = true` (lines 48-55) and the `originalIsTTY` capture (line 7) no longer do anything. Their comment still reads "Ensure stdin looks like a TTY so the auto-stdin path is not triggered" — but that path was removed in this change. The in-process `expandInputJson` tests never touch stdin anymore when `--input-json` is absent, and the stdin-reading cases run in a subprocess via `expandViaStdin`, so nothing depends on the stub. Consider dropping the two hooks plus line 7, or at minimum updating the stale comment. Pure test cleanup, no functional impact.

wyattjoh (commented):
One minor test-cleanup note inline. The fix itself is correct: stopping `expandInputJson` from implicitly draining stdin is the right call, since that stdin is shared with commands that read their own body (`clerk api`, `clerk config push`).

wyattjoh (approved):

PR #342, opened 2026-06-16
fix(agent): don't hint at a sandbox for ordinary network failures

### LINKED ISSUE #331: Network-failure hint suggests a sandboxed run even on a plain host shell
When a request fails at the network layer in agent mode (stdout piped), 1.5.0 prints:

```
Host-only Clerk state or system capabilities may be unavailable in agent mode. This may be a sandboxed run.
Re-run this command on the host shell before trusting auth, link, env, or API failures.
```

Repro: `CLERK_PLATFORM_API_KEY=ak_x CLERK_PLATFORM_API_URL=https://localhost:9 clerk api --platform /platform/applications | cat`

The hedge is reasonable when the caller might actually be sandboxed, but it prints identically on a plain host shell where there is no sandbox, and there is no way to suppress it. In our case the host was genuinely unreachable (VPN-gated environment) and the hint sent us down the wrong path first. Suggestions: only show it when there is a positive sandbox/agent signal, show it once rather than two lines, or gate it behind a verbosity flag so scripted callers can opt out.

rafa-thayto:
## Summary

In agent mode (stdout not a TTY), any network failure printed a two-line hint:

```
Host-only Clerk state … This may be a sandboxed run.
Re-run this command on the host shell before trusting auth, link, env, or API failures.
```

`isLikelySandboxFailure` returned `true` for **all** network errors with no inspection, so it fired identically on a plain host shell where a host was simply unreachable (e.g. VPN-gated), sending users down the wrong path. Network failures now require a permission-like error (`EPERM`/`EACCES`/`sandbox`/…) before hinting — real seatbelt/seccomp blocks still warn; `ECONNREFUSED`/`ENOTFOUND`/`ETIMEDOUT` do not. The hint is also collapsed to a single line. `browser-launch` and `localhost-bind` keep their existing behavior.

## Test plan
- New `src/lib/host-execution.test.ts` covers: plain connectivity failure → no hint; permission-like failure → hint; browser-launch → hint; single line; human mode → no hint
- `doctor.test.ts` still passes

Fixes #331

### DIFF
--- a/packages/cli-core/src/lib/host-execution.ts
+++ b/packages/cli-core/src/lib/host-execution.ts
@@ -101,17 +101,33 @@ function warnAboutSandbox(detail?: string): void {
   }
 }
 
+const PERMISSION_PATTERNS = [
+  /\bEPERM\b/i,
+  /\bEACCES\b/i,
+  /operation not permitted/i,
+  /permission denied/i,
+  /sandbox/i,
+  /interaction is not allowed/i,
+  /access denied/i,
+];
+
+function matchesPermissionPattern(s: string): boolean {
+  return PERMISSION_PATTERNS.some((pattern) => pattern.test(s));
+}
+
 function isPermissionLikeFailure(error: unknown): boolean {
-  const message = errorMessage(error);
-  return [
-    /\bEPERM\b/i,
-    /\bEACCES\b/i,
-    /operation not permitted/i,
-    /permission denied/i,
-    /sandbox/i,
-    /interaction is not allowed/i,
-    /access denied/i,
-  ].some((pattern) => pattern.test(message));
+  if (!(error instanceof Error)) {
+    return matchesPermissionPattern(String(error));
+  }
+
+  if (matchesPermissionPattern(error.message)) return true;
+
+  const code = (error as NodeJS.ErrnoException).code;
+  if (code && matchesPermissionPattern(code)) return true;
+
+  if (error.cause instanceof Error && isPermissionLikeFailure(error.cause)) return true;
+
+  return false;
 }
 
 function isLikelySandboxFailure(capability: HostCapability, error: unknown): boolean {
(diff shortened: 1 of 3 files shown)


wyattjoh (commented):
Correct, well-scoped fix: network failures now require a permission-like error before hinting at a sandbox, while browser-launch and localhost-bind keep warning on any failure. CI checks (lint, format, typecheck, tests) and the changeset all pass.

packages/cli-core/src/lib/host-execution.ts
```diff
@@ -118,11 +115,11 @@ function isPermissionLikeFailure(error: unknown): boolean {
...
-  if (
-    capability === "network" ||
-    capability === "browser-launch" ||
-    capability === "localhost-bind"
-  ) {
+  // browser-launch and localhost-bind only ever fail for host-capability
+  // reasons, so any failure is a meaningful sandbox signal. Network failures
+  // are different: a plain unreachable host (VPN, DNS, ECONNREFUSED) is not a
+  // sandbox, so require a permission-like error before hinting at a sandbox.
+  if (capability === "browser-launch" || capability === "localhost-bind") {
```
wyattjoh, line 138: For the `network` capability this now defers to `isPermissionLikeFailure`, which runs `errorMessage(error)` (returns only `error.message`) against patterns like `EPERM`/`EACCES`/`operation not permitted`. In practice Bun's `fetch` collapses connectivity failures into a generic message (`"Unable to connect. Is the computer able to access the url?"`) and puts the discriminator on `error.code` (e.g. `ConnectionRefused`) and `error.cause`, neither of which this code inspects. So a genuine sandbox that blocks network at the socket layer may surface the same generic message and the hint would be suppressed. This is not a regression (the old code over-warned on every network error, which is exactly the bug being fixed) and `home-fs`/`browser-launch`/`localhost-bind` still carry sandbox detection, so no action is strictly required. Consider, as a follow-up, also checking `(error as NodeJS.ErrnoException).code` and `error.cause` so the permission patterns can match real Bun network errors. The new test uses a synthetic `new Error("operation not permitted")`, which does not reflect Bun's real fetch error shape.

wyattjoh (approved):

PR #343, opened 2026-06-16
fix(ui): print next steps before animating the "Next steps" header

rafa-thayto:
## Summary

The glimmering "Next steps" header animated by redrawing only its own line in place (`\r\x1b[K`) for ~450ms, and every caller printed the actual steps **after** the animation finished. The result was a lone, shimmering "Next steps" with nothing under it — it looked hung/broken.

`animateHeader` now takes an optional `body`: it prints the whole block (header + steps) up front so everything is visible immediately, then steps the cursor back up to the header and sweeps the highlight in place, restoring the cursor below the block. Backward compatible — without a `body` it behaves exactly as before.

Wired through the two call sites: `spinner.outro([steps])` (covers `clerk link`, `auth login`, `init`, and any gutter next-steps) and `clerk deploy`. `log.applyPrefix` is exported so the deploy body keeps its gutter prefix.

Verified the real escape sequence ordering:
```
<HIDE> header + all steps printed  →  <UP n>  →  header redraws (sweep)  →  <DOWN n>  →  <SHOW>
```

## Test plan
- New tests in `gradient.test.ts`: body printed in one write (non-interactive); body rendered before the first in-place redraw with correct cursor up/down (interactive)
- `spinner.test.ts`, deploy tests pass; non-interactive / `NO_COLOR` / piped output unchanged (single flat write, no cursor escapes)

### DIFF
--- a/packages/cli-core/src/lib/gradient.ts
+++ b/packages/cli-core/src/lib/gradient.ts
@@ -149,6 +149,14 @@ export async function animateHeader(options: AnimateHeaderOptions): Promise<void
   // header's own newline plus every newline inside the body.
   const rowsBelow = 1 + (body.match(/\n/g)?.length ?? 0);
 
+  // The cursor-up escape only lands on the header when the block fits on screen
+  // without scrolling. Fall back to a plain write on short terminals where the
+  // scroll would push the header out of reach.
+  if (process.stderr.rows != null && rowsBelow >= process.stderr.rows) {
+    write(`${prefix}${fallback(label)}\n${body}`);
+    return;
+  }
+
   hideCursor();
   try {
     // Print the whole block first so the body is visible immediately, then step
--- a/packages/cli-core/src/lib/gradient.ts
+++ b/packages/cli-core/src/lib/gradient.ts
@@ -114,29 +114,55 @@ interface AnimateHeaderOptions {
   prefix: string;
   label: string;
   fallback: (s: string) => string;
+  /**
+   * Text printed beneath the header (e.g. the next-steps list), including its
+   * own trailing newline(s). It is rendered up front so the whole block is on
+   * screen immediately; the highlight then sweeps across the header in place.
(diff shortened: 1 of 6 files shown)


packages/cli-core/src/lib/gradient.ts
```diff
@@ -114,29 +114,55 @@ interface AnimateHeaderOptions {
...
-    write(`${prefix}${fallback(label)}\n`);
+    write(`${prefix}${fallback(label)}\n${body}`);
     return;
   }
 
   const truecolor = supportsTruecolor();
   const span = Math.max(1, frames - 1);
+  // Rows the cursor ends below the header after the block is printed: the
+  // header's own newline plus every newline inside the body.
+  const rowsBelow = 1 + (body.match(/\n/g)?.length ?? 0);
```
wyattjoh, line 150: `rowsBelow = 1 + body newlines`, and after printing the whole block the code does `\x1b[${rowsBelow}A` to step back up to the header before sweeping. This is correct only when the block fits on screen without scrolling. The deploy `nextStepsBody` produces ~27 newlines, so on a terminal shorter than the block (or when the block is printed near the bottom of the viewport), printing it scrolls the buffer and the cursor-up no longer lands on the header line. The sweep then redraws onto a wrong line and the symmetric `\x1b[${rowsBelow}B` lands the cursor in the wrong place, corrupting the output. The lone-header animation this replaces only ever moved a single line in place, so this multi-row dance is new exposure. Worth at least a comment acknowledging the assumption; a more robust fix would cap the animation (or fall back to the plain write) when `rowsBelow` exceeds `process.stderr.rows`.

wyattjoh (commented):
Clean, well-tested fix for the lone "Next steps" header. One edge case worth noting around tall bodies and terminal scrolling, but it does not regress the prior behavior.

wyattjoh (approved):

Issue #329, opened 2026-06-11
Auth: support multiple platform environments (per-host credential profiles)


jacekradko:
While dogfooding the CLI (1.5.0, agent-driven) against a second Clerk platform environment, we hit a wall on auth: `CLERK_PLATFORM_API_URL` cleanly redirects requests, but the credential store only holds one OAuth token and `clerk auth login` has no notion of which host a token belongs to. Logging into the alternate environment would clobber the production login, so the only safe path is minting a platform API key for the second environment and threading `CLERK_PLATFORM_API_KEY` through every call.

That workaround is fine for CI but makes non-production work second-class for interactive use. A per-host credential store (key tokens by platform API URL) or an explicit `--profile` on `auth login`/`api`/`config` would make switching environments first-class. Bonus: `clerk doctor` could report which environment the stored token belongs to.

jfoshee:
What if there was an association w/ environment or token per directory?

```
pushd /a
clerk switch-env staging
clerk login # select workspace a
popd

pushd /b
clerk switch-env prod
clerk login # select workspace b
popd

pushd /c
clerk switch-env prod
clerk login # select workspace c
popd

cd /a
clerk whoami # user on staging env (workspace a)
```

To me, this seems easy to reason about.

jacekradko:
Correction after more digging: the core of this ask already ships in 1.5.0. `clerk switch-env` lists `production` and `staging` profiles (OAuth client ID, OAuth base URL, Platform API URL, and Backend API URL all switch together), and the credential store keys tokens per environment (`oauth-access-token:<envName>` in the keychain, `credentials.<envName>` for the file fallback), so logging into staging leaves the production token intact. I filed this issue having missed the command entirely.

Which points at what I think is the real residual issue: discoverability. `switch-env` doesn't appear in `clerk --help` (the closest entry is `env`, which manages environment variables), so the feature is invisible unless you already know its name. Suggestions:

- surface `switch-env` in the top-level help output
- have `clerk whoami` and/or `clerk doctor` report the active environment alongside the account

Happy for this to be closed in favor of a narrower "make switch-env discoverable" item, or retitled and kept; either works.

jfoshee:
We hid it because `switch-env` is not useful to users outside of Clerk.  Perhaps we can display it if we recognize the user has a Clerk email address.  That of course requires the user log in at least once.  wdyt?

PR #344, opened 2026-06-16
fix(env): surface `CLERK_PLATFORM_API_URL` credential mismatch

rafa-thayto:
## Summary

Addresses the immediate, concrete gap from #329. `CLERK_PLATFORM_API_URL` redirects API traffic to an arbitrary host, but credentials are keyed by **environment name** (via `switch-env`), not by URL — so when the override points somewhere else, the active environment's token is silently sent to that host with no credential isolation.

This is a Tier-1 / diagnostic fix:
- Warn in **human mode** when `CLERK_PLATFORM_API_URL` differs from the active environment's platform URL (agent/scripted stderr stays clean — it would otherwise corrupt machine-readable output; agents get the same info from `doctor`).
- `clerk doctor` now reports the active environment name and effective API URL in the "Logged in" check.

The broader request in #329 (user-definable environment profiles for full multi-env credential isolation) is a separate, larger change and is **not** included here — this PR makes the silent failure visible.

## Test plan
- New `src/lib/environment.test.ts`: warns on override mismatch in human mode; silent when unset, when equal, and in agent mode
- `doctor` tests pass

Refs #329

### DIFF
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -43,7 +43,7 @@ import {
   throwUsageError,
 } from "./lib/errors.ts";
 import { clerkHelpConfig, formatExamplesBlock, type Example } from "./lib/help.ts";
-import { isAgent } from "./mode.ts";
+import { isAgent, isHuman } from "./mode.ts";
 import { log } from "./lib/log.ts";
 import { maybeNotifyUpdate } from "./lib/update-check.ts";
 import { CURRENT_VERSION } from "./lib/version.ts";
@@ -147,12 +147,13 @@ export function createProgram(): Program {
       process.stderr.write(`[${activeEnv.toUpperCase()}]\n`);
     }
 
-    // Warn when CLERK_PLATFORM_API_URL routes requests to a different host than
-    // the active environment's platform URL. Credentials are keyed by environment
-    // name, so the active env's token will be sent to the override host. Emitted
-    // to stderr so it never corrupts stdout data output.
+    // Warn (human mode only) when CLERK_PLATFORM_API_URL routes requests to a
+    // different host than the active environment's platform URL. Credentials are
+    // keyed by environment name, so the active env's token will be sent to the
+    // override host. Agent/scripted mode stays clean — stray lines on stderr
+    // corrupt machine-readable output; those callers get this info from `doctor`.
     const override = isPlatformApiUrlOverridden();
-    if (override.overridden) {
+    if (override.overridden && isHuman()) {
       log.warn(
         `CLERK_PLATFORM_API_URL is routing requests to ${override.overrideUrl}, but credentials stay keyed to the "${override.envName}" environment — the "${override.envName}" token will be sent to that host.`,
       );
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -30,6 +30,7 @@ import {
   getCurrentEnvName,
   getAvailableEnvs,
   getPlapiBaseUrl,
+  warnIfPlatformApiUrlOverride,
 } from "./lib/environment.ts";
 import {
   CliError,
@@ -145,6 +146,8 @@ export function createProgram(): Program {
     if (activeEnv !== "production") {
       process.stderr.write(`[${activeEnv.toUpperCase()}]\n`);
     }
+
+    warnIfPlatformApiUrlOverride();
   });
 
   // Show update notification after each command, except for commands that
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -239,8 +239,26 @@ async function resolveArgv(
 ): Promise<{ argv: string[]; from: ParseFrom }> {
   const raw = args ?? process.argv;
   const effectiveFrom = from ?? (args === undefined ? "node" : "user");
+  // Honor an explicit `--mode` flag before `expandInputJson`, so a failure it
+  // throws (invalid JSON, missing file) is formatted in the right mode. The
+  // preAction hook re-applies and validates `--mode`, but it runs during
+  // parseAsync — too late for errors raised while resolving argv.
+  applyForcedModeFromArgv(raw);
   const argv = await expandInputJson([...raw]);
   return { argv, from: effectiveFrom };
+}
+
+/**
+ * Scan raw argv for an explicit `--mode human|agent` and force it early.
+ * Ignores an invalid value — the preAction hook reports that as a usage error.
+ */
+function applyForcedModeFromArgv(argv: string[]): void {
+  const idx = argv.indexOf("--mode");
+  if (idx === -1) return;
+  const value = argv[idx + 1];
+  if (value === "human" || value === "agent") {
+    setMode(value);
+  }
 }
 
 /**
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -30,7 +30,7 @@ import {
   getCurrentEnvName,
   getAvailableEnvs,
   getPlapiBaseUrl,
-  isPlatformApiUrlOverridden,
+  getPlatformApiUrlOverride,
 } from "./lib/environment.ts";
 import {
   CliError,
@@ -150,13 +150,14 @@ export function createProgram(): Program {
     // Warn (human mode only) when CLERK_PLATFORM_API_URL routes requests to a
     // different host than the active environment's platform URL. Credentials are
     // keyed by environment name, so the active env's token will be sent to the
-    // override host. Agent/scripted mode stays clean — stray lines on stderr
-    // corrupt machine-readable output; those callers get this info from `doctor`.
-    const override = isPlatformApiUrlOverridden();
-    if (override.overridden && isHuman()) {
-      log.warn(
-        `CLERK_PLATFORM_API_URL is routing requests to ${override.overrideUrl}, but credentials stay keyed to the "${override.envName}" environment — the "${override.envName}" token will be sent to that host.`,
-      );
+    // override host. Agent/scripted mode stays off stderr — stray lines there
+    // corrupt machine-readable output — but still gets the info via `log.debug`
+    // so a `--verbose` re-run of a failing script surfaces it.
+    const override = getPlatformApiUrlOverride();
+    if (override.overridden) {
+      const msg = `CLERK_PLATFORM_API_URL is routing requests to ${override.overrideUrl} instead of the "${override.envName}" environment's ${override.profileUrl} — the "${override.envName}" token will be sent to that host.`;
+      if (isHuman()) log.warn(msg);
+      else log.debug(`env: ${msg}`);
     }
   });
 
--- a/packages/cli-core/src/cli-program.ts
+++ b/packages/cli-core/src/cli-program.ts
@@ -30,7 +30,7 @@ import {
   getCurrentEnvName,
   getAvailableEnvs,
   getPlapiBaseUrl,
-  warnIfPlatformApiUrlOverride,
+  isPlatformApiUrlOverridden,
 } from "./lib/environment.ts";
 import {
   CliError,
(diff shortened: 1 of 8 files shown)


packages/cli-core/src/lib/environment.ts
```diff
@@ -136,6 +137,27 @@ export function getPlapiBaseUrl(): string {
...
+ * Human mode only: a per-command warning line would corrupt the machine-readable
+ * stderr that agent mode emits. Agent/scripted callers get the same information
+ * from the `clerk doctor` environment report instead.
+ */
+export function warnIfPlatformApiUrlOverride(): void {
+  if (!isHuman()) return;
+  const override = process.env.CLERK_PLATFORM_API_URL;
+  if (!override) return;
+  const envName = getCurrentEnvName();
+  if (override === getCurrentEnv().platformApiUrl) return;
```
wyattjoh: `override === getCurrentEnv().platformApiUrl` compares the two URLs as raw strings, so `CLERK_PLATFORM_API_URL=https://api.clerk.com/` (trailing slash) or a host-case difference against the profile's `https://api.clerk.com` will fire the mismatch warning even though both resolve to the same host. For a diagnostic that exists specifically to reduce surprise, a spurious warning undercuts the signal. Consider normalizing both sides before comparing, e.g. compare `new URL(override).href` against `new URL(profileUrl).href`, or at minimum strip a trailing slash. Wrap the `new URL()` calls so a malformed override string falls back to the raw comparison rather than throwing inside the preAction hook.

packages/cli-core/src/lib/environment.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    expect(captured.err).not.toContain("CLERK_PLATFORM_API_URL");
+  });
+
+  test("does not warn when the override equals the active env URL", () => {
+    const profileUrl = getPlapiBaseUrl(); // no override set → active env URL
+    process.env.CLERK_PLATFORM_API_URL = profileUrl;
+    warnIfPlatformApiUrlOverride();
+    expect(captured.err).not.toContain("CLERK_PLATFORM_API_URL");
+  });
+
```
wyattjoh, line 36: This case sets `process.env.CLERK_PLATFORM_API_URL = getPlapiBaseUrl()` and asserts no warning. Because `getPlapiBaseUrl()` returns the active env's `platformApiUrl` (no override set at call time), the test is effectively asserting `x === x` and would still pass even if the equality guard were removed and replaced with any branch that happens to not warn here. Pin the expected value to a concrete literal (the production profile URL, `https://api.clerk.com`) so the test fails if the guard logic regresses. Separately, the positive case only asserts the message contains `CLERK_PLATFORM_API_URL` and `production`; consider also asserting it names the override host so the warning's core claim (token goes to that host) stays covered.

wyattjoh (commented):
Clean, well-scoped diagnostic fix with good test coverage. One minor robustness note on the URL comparison and a small test-quality suggestion.

packages/cli-core/src/lib/environment.ts
```diff
@@ -136,6 +137,34 @@ export function getPlapiBaseUrl(): string {
...
+ * from the active environment's platform URL. Credentials are keyed by
+ * environment name, not by URL, so the active env's token is what gets sent to
+ * the override host — surface that so it isn't a silent surprise.
+ *
+ * Human mode only: a per-command warning line would corrupt the machine-readable
+ * stderr that agent mode emits. Agent/scripted callers get the same information
+ * from the `clerk doctor` environment report instead.
+ */
+export function warnIfPlatformApiUrlOverride(): void {
+  if (!isHuman()) return;
```
wyattjoh: Curious if we can challenge this. We emit the machine readable output to a single stream, maybe the other one it's ok to add warnings?

packages/cli-core/src/cli-program.ts
```diff
@@ -129,6 +130,8 @@ export function createProgram(): Program {
     if (activeEnv !== "production") {
       process.stderr.write(`[${activeEnv.toUpperCase()}]\n`);
     }
+
+    warnIfPlatformApiUrlOverride();
```
wyattjoh: An alternative to this design would have the environment lib export a function that says, `isPlatformApiUrlOverrided` Then, at this call site is where it decides to actually call the log.warn. That way, possibly we can use it to leverage the different stream that we want to use for the warnings emitting to the Agent.

wyattjoh (commented):
If the authentication information is tied to the URL being authenticated with, doesn't it make sense just to instead refuse to use that credential and consider the user as locked out?

packages/cli-core/src/cli-program.ts
```diff
@@ -135,6 +136,18 @@ export function createProgram(): Program {
...
       process.stderr.write(`[${activeEnv.toUpperCase()}]\n`);
     }
+
+    // Warn (human mode only) when CLERK_PLATFORM_API_URL routes requests to a
+    // different host than the active environment's platform URL. Credentials are
+    // keyed by environment name, so the active env's token will be sent to the
+    // override host. Agent/scripted mode stays clean — stray lines on stderr
+    // corrupt machine-readable output; those callers get this info from `doctor`.
+    const override = isPlatformApiUrlOverridden();
+    if (override.overridden && isHuman()) {
```
wyattjoh: This warning fires on effectively every integration test, and the workarounds in `input-json.test.ts` / `error-codes.test.ts` treat the symptom rather than the cause.

Two harness facts combine here:

1. `setupTest()` sets `CLERK_PLATFORM_API_URL=https://test-api.clerk.com` while the active environment stays `production` (`https://api.clerk.com`), so `isPlatformApiUrlOverridden()` returns `overridden: true` in every test.
2. The harness mocks `mode.ts` with a module-level `let _mode = "human"` that is never reset in `setupTest`/`teardownTest`.

So the default state for an integration test is human + overridden, and this warning becomes the **first line of stderr** for any command that reaches `preAction`. I confirmed it with a scratch test:

```
clerk.raw("doctor", "--json")
// stderr[0] = "CLERK_PLATFORM_API_URL is routing requests to https://test-api.clerk.com, ..."
```

Worse, because `_mode` leaks across tests within a file, whether the warning appears depends on test order. Three tests in one file:

```
A: clerk.raw("doctor", "--json")                    -> warning present
B: clerk.raw("--mode", "agent", "doctor", "--json") -> warning absent
C: clerk.raw("doctor", "--json")                    -> warning ABSENT
```

A and C are byte-identical commands with different stderr, purely because B ran in between and left `_mode` as `"agent"`.

Everything passes today, but this leaves a tripwire for any future test that asserts on stderr contents or the first stderr line. I'd rather fix it in the harness than in each consumer: reset `_mode = "human"` in `setupTest()` to kill the order-dependence, and either point `CLERK_PLATFORM_API_URL` at the active profile's URL or register a test env profile whose `platformApiUrl` is `https://test-api.clerk.com`, so the harness stops tripping a warning it doesn't mean to exercise. The JSON-parsing helpers in the two test files can then go away entirely.

wyattjoh (commented):
The core idea is right, and the normalized-URL comparison plus the human-mode gate are both good calls. The main thing I'd like resolved before merge is the integration-harness fallout: this warning now fires on essentially every integration test, and it landed as JSON-parsing workarounds in two test files rather than a fix at the source. Details inline.

packages/cli-core/src/test/integration/input-json.test.ts
```diff
@@ -234,16 +243,14 @@ test("--input-json is registered as a global option", async () => {
   // stderr is the agent-mode signature, which proves --mode agent was applied.
   const result = await clerk.raw("--input-json", '{"mode":"agent"}', "doctor", "--json");
   expect(result.exitCode).not.toBe(0);
-  const jsonOutput = result.stderr.split("\n").findLast((line) => line.startsWith("{"));
-  expect(jsonOutput).toBeDefined();
-  const parsed = JSON.parse(jsonOutput!);
+  const parsed = parseJsonFromStderr(result.stderr);
```
wyattjoh: This swap quietly changes semantics. The original code deliberately took the **last** JSON line:

```ts
const jsonOutput = result.stderr.split("\n").findLast((line) => line.startsWith("{"));
```

The new shared helper uses `.find(...)`, i.e. the **first** one. For a command like `doctor --json` that can emit several structured lines to stderr, first and last are not the same object, and the assertion would then be checking the wrong payload.

It passes right now only because nothing before the error line happens to start with `{` in this environment. Since the original author chose `findLast` on purpose, preserve that in the helper:

```ts
const jsonLine = stderr
  .split("\n")
  .map((l) => l.trim())
  .findLast((l) => l.startsWith("{"));
```

packages/cli-core/src/cli-program.ts
```diff
@@ -135,6 +136,18 @@ export function createProgram(): Program {
...
+      log.warn(
+        `CLERK_PLATFORM_API_URL is routing requests to ${override.overrideUrl}, but credentials stay keyed to the "${override.envName}" environment — the "${override.envName}" token will be sent to that host.`,
```
wyattjoh: I agree with keeping this off stderr in agent mode — the test fallout in this same PR is a good demonstration of why stray lines break machine-readable output. But the result is that agent/scripted mode gets *zero* signal, and that's arguably the higher-risk context: a CI job with `CLERK_PLATFORM_API_URL` set will ship the production token to the override host with nothing in the logs. The comment points those callers at `doctor`, but nothing obliges a script to run `doctor`.

A cheap middle ground that costs nothing in cleanliness: emit it as `log.debug` in the non-human branch. It stays out of normal output, but a `--verbose` re-run of a failing CI job surfaces it, which matches how the rest of the CLI handles diagnostics.

```ts
if (override.overridden) {
  const msg = `CLERK_PLATFORM_API_URL is routing requests to ${override.overrideUrl}, ...`;
  if (isHuman()) log.warn(msg);
  else log.debug(`env: ${msg}`);
}
```

packages/cli-core/src/lib/environment.ts
```diff
@@ -141,6 +141,42 @@ export function getPlapiBaseUrl(): string {
...
+/**
+ * Returns whether CLERK_PLATFORM_API_URL is set to a URL that differs from the
+ * active environment's configured platform URL, along with both URLs so the
+ * caller can surface a warning.
+ *
+ * Comparison normalises both URLs via `new URL().href` so trailing-slash and
+ * case differences are ignored; falls back to raw string comparison if either
+ * value is not a valid URL.
+ */
+export function isPlatformApiUrlOverridden():
```
wyattjoh: Two small things on this function:

**Naming.** An `is*` prefix reads as returning a boolean, but this returns a discriminated union. `getPlatformApiUrlOverride()` or `checkPlatformApiUrlOverride()` would signal the shape better. Minor, but it's a new exported API so it's cheap to get right now.

**Unused field.** `profileUrl` is returned but the only caller never uses it, so the warning tells the user where traffic is going without saying where it *should* be going. Including it makes the mismatch self-diagnosing: `...routing requests to ${overrideUrl} instead of the "${envName}" environment's ${profileUrl}...`. Otherwise, drop the field.

Also worth a test: the JSDoc documents a raw-string fallback when either value fails `new URL()`, but `environment.test.ts` never exercises that `catch` branch. A case with `CLERK_PLATFORM_API_URL="not a url"` would cover it.

packages/cli-core/src/test/integration/error-codes.test.ts
```diff
@@ -9,7 +9,12 @@ import { useIntegrationTestHarness, clerk, mockState } from "./lib/harness.ts";
 useIntegrationTestHarness();
 
 function parseJsonError(stderr: string): { code: string; message: string; docsUrl?: string } {
-  const parsed = JSON.parse(stderr);
+  const jsonLine = stderr
```
wyattjoh: This line-scanning helper is now copy-pasted here and in `input-json.test.ts` (as `parseJsonFromStderr`), with the same body and the same error message. If these survive the harness fix suggested elsewhere, hoist a single `parseJsonFromStderr` into `./lib/harness.ts` and import it in both files — otherwise the two copies will drift, and any future fix (like the `findLast` behavior) has to be applied twice.

packages/cli-core/src/test/integration/input-json.test.ts
wyattjoh: Should maybe just ensure that if you expect there only to be a single line, that we test for that.

PR #345, opened 2026-06-16
feat(api): add `clerk api --fapi` for the public Frontend API

### LINKED ISSUE #332: Feature: fetch an instance's public FAPI /v1/environment for config verification
After changing instance configuration (in our case toggling bot protection via the platform config API), the natural way to verify the change took effect is the instance's public FAPI `/v1/environment` payload, since that is what clerk-js actually consumes. The CLI speaks BAPI and PLAPI but has no FAPI surface, so verification falls back to curl plus manually decoding the FAPI domain out of the publishable key.

A small addition would close the loop, any of:

- `clerk api --fapi /environment --app <id> --instance <dev|prod>` (resolve the FAPI domain from the instance's pk, no auth needed for public endpoints)
- a `clerk config pull --effective` style command that fetches the public environment alongside the platform config

This would make config changes verifiable end to end with the CLI alone.

rafa-thayto:
## Summary

`clerk api` spoke only BAPI (default) and PLAPI (`--platform`). After a config change, the natural way to verify it took effect is the instance's public FAPI `/v1/environment` payload (what clerk-js actually consumes) — but that meant dropping to `curl` plus decoding the FAPI domain out of the publishable key by hand.

This adds `--fapi`:

```sh
clerk api --fapi /environment --app <id> --instance dev
```

- Resolves the FAPI host from the instance's publishable key (looked up via `--app`/`--instance` or the linked project through the Platform API), reusing the existing `lib/fapi.ts` client and `decodePublishableKey`.
- The request itself is **unauthenticated** — these endpoints are public.
- Paths are `/v1`-normalized like the other modes, so both `/environment` and `/v1/environment` work.
- `--fapi` and `--platform` are mutually exclusive.

Implementation reuses the request-target resolution: `api()` now builds a `{ baseUrl, runRequest }` pair, which also de-duplicates the BAPI/PLAPI branches. The local error handler was broadened from `BapiError` to `ApiError` so FAPI error bodies still print to stdout for piping.

## Test plan
- New tests in `src/commands/api/index.test.ts`: host resolution + no auth header; `--fapi`/`--platform` conflict; FAPI error body printed to stdout with exit 1
- Full api + completion + api-queries integration suites pass

Closes #332

### DIFF
--- a/packages/cli-core/src/commands/api/fapi.ts
+++ b/packages/cli-core/src/commands/api/fapi.ts
@@ -15,13 +15,10 @@ import {
   throwUsageError,
   withApiContext,
 } from "../../lib/errors.ts";
-import { decodePublishableKey } from "../../lib/fapi.ts";
+import { decodePublishableKey, CLERK_JS_API_VERSION } from "../../lib/fapi.ts";
 import { loggedFetch } from "../../lib/fetch.ts";
 import { fetchApplication, type ApplicationInstance } from "../../lib/plapi.ts";
 import type { BapiResponse } from "./bapi.ts";
-
-/** clerk-js API version FAPI shapes its `/v1/environment` payload for. */
-const CLERK_JS_API_VERSION = "5";
 
 interface ResolveOptions {
   app?: string;
--- a/packages/cli-core/src/commands/api/fapi.ts
+++ b/packages/cli-core/src/commands/api/fapi.ts
@@ -53,14 +53,14 @@ async function resolveInstance(options: ResolveOptions): Promise<ApplicationInst
   }
 
   const app = await withApiContext(fetchApplication(ctx.appId), "Failed to resolve instance");
-  const instance = app.instances.find((entry) => entry.instance_id === ctx.instanceId);
-  if (!instance) {
+  const resolved = resolveFetchedApplicationInstance(ctx.appId, app, ctx.instanceId);
+  if (!resolved.found) {
     throw new CliError(`Instance ${ctx.instanceId} not found in application.`, {
       code: ERROR_CODE.INSTANCE_NOT_FOUND,
       docsUrl: "https://clerk.com/docs/guides/development/managing-environments",
     });
   }
-  return instance;
+  return resolved.instance;
 }
 
 /** Resolve the instance's FAPI host from its publishable key. */
--- a/packages/cli-core/src/commands/api/fapi.ts
+++ b/packages/cli-core/src/commands/api/fapi.ts
@@ -1,0 +1,109 @@
+/**
+ * Frontend API (FAPI) passthrough for `clerk api --fapi`.
+ *
+ * FAPI is the public API that clerk-js consumes. Its host is per-instance and
+ * derived from the instance's publishable key, and the endpoints exposed here
+ * (e.g. `/v1/environment`) are public, so no auth header is sent.
+ */
+
+import { resolveAppContext, resolveFetchedApplicationInstance } from "../../lib/config.ts";
+import { normalizeBapiPath } from "../../lib/bapi-command.ts";
+import {
+  CliError,
+  ERROR_CODE,
+  FapiError,
+  throwUsageError,
+  withApiContext,
+} from "../../lib/errors.ts";
+import { decodePublishableKey } from "../../lib/fapi.ts";
+import { loggedFetch } from "../../lib/fetch.ts";
+import { fetchApplication, type ApplicationInstance } from "../../lib/plapi.ts";
+import type { BapiResponse } from "./bapi.ts";
+
+/** clerk-js API version FAPI shapes its `/v1/environment` payload for. */
+const CLERK_JS_API_VERSION = "5";
+
+interface ResolveOptions {
+  app?: string;
+  instance?: string;
+}
+
+async function resolveInstance(options: ResolveOptions): Promise<ApplicationInstance> {
+  if (options.app) {
+    const app = await withApiContext(fetchApplication(options.app), "Failed to resolve instance");
+    const resolved = resolveFetchedApplicationInstance(options.app, app, options.instance);
+    if (!resolved.found) {
+      throw new CliError(`Instance ${resolved.instanceId} not found in application.`, {
+        code: ERROR_CODE.INSTANCE_NOT_FOUND,
+        docsUrl: "https://clerk.com/docs/guides/development/managing-environments",
+      });
+    }
+    return resolved.instance;
+  }
+
+  let ctx: Awaited<ReturnType<typeof resolveAppContext>>;
+  try {
+    ctx = await resolveAppContext({ app: options.app, instance: options.instance });
+  } catch (error) {
+    if (error instanceof CliError && error.code === ERROR_CODE.NOT_LINKED) {
+      throwUsageError(
+        "No instance found. Link a project with `clerk link`, or pass --app <app_id>.",
+        "https://clerk.com/docs/guides/development/managing-environments",
+        ERROR_CODE.NOT_LINKED,
+      );
+    }
+    throw error;
+  }
+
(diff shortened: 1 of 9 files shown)


packages/cli-core/src/commands/api/index.ts
```diff
@@ -61,17 +93,8 @@ export async function api(
...
-
-    if (options.platform) {
-      secretKey = await getAuthToken();
-      baseUrl = getPlapiBaseUrl();
-    } else {
-      secretKey = await resolveBapiSecretKey(options);
-      baseUrl = getBapiBaseUrl();
-    }
+    // 3. Resolve the request target (base URL + executor)
+    const { baseUrl, runRequest } = await resolveApiTarget(options);
```
wyattjoh: `resolveApiTarget` runs at step 3, before the `options.dryRun` check at step 4. For `--fapi` this means `resolveFapiHost` -> `resolveInstance` -> `fetchApplication` makes a real Platform API request (and requires PLAPI auth) even when the user asked only to preview the request with `--dry-run`. That contradicts the `--dry-run` help text ("Show the request without executing it") and surprises users who run dry-run precisely to avoid network/auth. Consider resolving the host lazily, or special-casing dry-run for fapi so the base URL can be shown (or noted as unresolved) without the round-trip.

packages/cli-core/src/commands/api/fapi.ts
```diff
@@ -0,0 +1,109 @@
...
+  CliError,
+  ERROR_CODE,
+  FapiError,
+  throwUsageError,
+  withApiContext,
+} from "../../lib/errors.ts";
+import { decodePublishableKey } from "../../lib/fapi.ts";
+import { loggedFetch } from "../../lib/fetch.ts";
+import { fetchApplication, type ApplicationInstance } from "../../lib/plapi.ts";
+import type { BapiResponse } from "./bapi.ts";
```
wyattjoh: `CLERK_JS_API_VERSION = "5"` here duplicates the identical constant (same value, same intent, near-identical comment) already exported context in `src/lib/fapi.ts`. Two copies will drift when the clerk-js major version is bumped. Import and reuse the existing constant from `src/lib/fapi.ts` rather than redeclaring it in the command module.

packages/cli-core/src/commands/api/fapi.ts
```diff
@@ -0,0 +1,109 @@
...
+
+/** clerk-js API version FAPI shapes its `/v1/environment` payload for. */
+const CLERK_JS_API_VERSION = "5";
+
+interface ResolveOptions {
+  app?: string;
+  instance?: string;
+}
+
```
wyattjoh, line 18: The `--app` branch duplicates the pattern in `resolveBapiSecretKey` (fetch app, call `resolveFetchedApplicationInstance`, throw INSTANCE_NOT_FOUND when not found), and the fallback branch hand-rolls `app.instances.find(entry => entry.instance_id === ctx.instanceId)` instead of reusing `resolveFetchedApplicationInstance`. Both paths could route through the shared helper so instance-matching logic (alias handling, dev default, not-found messaging) lives in one place and stays consistent with the BAPI command.

packages/cli-core/src/commands/api/index.ts
```diff
@@ -25,12 +26,43 @@ export interface ApiOptions {
...
 }
 
 const MUTATING_METHODS = new Set(["POST", "PUT", "PATCH", "DELETE"]);
 
+type RunRequest = (req: { method: string; path: string; body?: string }) => Promise<BapiResponse>;
+
+/** Resolve the API surface (base URL + request executor) from the flags. */
+async function resolveApiTarget(
+  options: ApiOptions,
+): Promise<{ baseUrl: string; runRequest: RunRequest }> {
```
wyattjoh, line 53: In `--fapi` mode the request is intentionally unauthenticated, so a `--secret-key` passed alongside `--fapi` is silently dropped with no warning. `--fapi` + `--platform` correctly errors as a usage conflict; consider the same treatment (or at least a `log.warn`) for `--secret-key` so users do not believe their key is being used.

packages/cli-core/src/commands/api/index.test.ts
```diff
@@ -476,6 +476,72 @@ describe("api command", () => {
     );
   });
 
+  // --- --fapi mode ---
```
wyattjoh, line 479: The added tests cover the `--app` happy path, the `--platform` conflict, and an API error. Two documented behaviors are untested: the no-`--app` path that errors with the `clerk link`/`--app` NOT_LINKED guidance, and the README's claim that both `/environment` and `/v1/environment` resolve to the same FAPI URL. Adding a case for each would lock in the host-resolution fallback and the path-normalization contract.

wyattjoh (commented):
Clean, well-scoped addition that mirrors the existing BAPI/PLAPI passthrough patterns. A few follow-ups around dry-run semantics, a duplicated version constant, and silently-ignored flags.

packages/cli-core/src/lib/fapi.ts
```diff
@@ -15,7 +15,7 @@ const PK_LIVE_PREFIX = "pk_live_";
  * The clerk-js client version FAPI's `/v1/environment` payload is shaped for.
  * Bump when consuming response fields introduced in a later major version.
  */
-const CLERK_JS_API_VERSION = "5";
+export const CLERK_JS_API_VERSION = "5";
```
dmoerner: Should we not target 6 here?

packages/cli-core/src/commands/api/index.ts
```diff
@@ -216,6 +245,10 @@ export function registerApi(program: Program): void {
     .option("--secret-key <key>", "Override the secret key")
     .option("--instance <id>", "Instance to target (dev, prod, or instance ID)")
     .option("--platform", "Use Platform API instead of Backend API")
+    .option(
+      "--fapi",
+      "Use the instance's public Frontend API (no auth; host resolved from the publishable key)",
```
dmoerner: I'd like it to be a bit clearer that this only can be used with unauthenticated endpoints. "no auth" might make it seem like it skips auth instead.

dmoerner (approved):
I think we should just be clearer that this only hits public endpoints.

Because it is possible to use the command line to hit FAPI with curl, and use auth, you just need to use a cookie jar. This is how I do it in my local skill for testing:

```sh
FAPI=""
jar=$(mktemp)

# 1. Create a dev browser; capture its token.
tok=$(curl -s -c "$jar" -X POST "$FAPI/v1/dev_browser" -H "Origin: $FAPI" | jq -r '.token')

# 2. Inspect what the instance has enabled (strategies, factors).
curl -s -b "$jar" -c "$jar" "$FAPI/v1/environment?__clerk_db_jwt=$tok" -H "Origin: $FAPI" \
  | jq '.auth_config | {first_factors, identification_strategies, password}'

# 3. The client state (sessions, sign_in, sign_up).
curl -s -b "$jar" -c "$jar" "$FAPI/v1/client?__clerk_db_jwt=$tok" -H "Origin: $FAPI" \
  | jq '.response | {sessions: (.sessions|length)}'
```

packages/cli-core/src/commands/api/fapi.ts
```diff
@@ -0,0 +1,106 @@
...
+  const rawBody = await response.text();
+  let body: unknown;
+  try {
+    body = JSON.parse(rawBody);
+  } catch {
+    body = rawBody;
+  }
+
+  return { status: response.status, headers: response.headers, body, rawBody };
+}
```
wyattjoh: Feels like this should live in `lib/fapi.ts`? wdyt?

rafa-thayto: sure, makes sense!

opened this one to move bapi as well: https://github.com/clerk/cli/pull/359

PR #351, opened 2026-06-17
fix(cli): show all API errors and surface plain error/message bodies

wyattjoh:
## Summary

When an API request fails, the CLI now parses the response body so every error in a multi-error response is shown, joined by newlines, instead of only the first one. Response bodies that carry a plain `error` or `message` field instead of a Clerk `errors` array are surfaced directly rather than printed as raw JSON. The error classes and their structured fields are unchanged, so existing consumers that read `code` and `meta` (such as the deploy command) keep working.

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bun run test` (1679 pass)

### DIFF
--- a/.changeset/api-error-formatting.md
+++ b/.changeset/api-error-formatting.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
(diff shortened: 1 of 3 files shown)


wyattjoh:
**Stack: wyattjoh/errors-eval**

- **#351 👈 this PR**

PR #354, opened 2026-06-17
ci: pin GitHub workflows to Node.js 24

wyattjoh:
## What

- Pin every GitHub Actions `actions/setup-node` step to Node.js 24.

## Why

This keeps CI, release publishing, snapshot publishing, and fixture refresh jobs on the requested Node.js major version.


rafa-thayto (approved):
nice!

PR #360, opened 2026-06-22
fix(e2e): bound network and setup steps to kill flaky 300s timeouts

rafa-thayto:
## Problem

The E2E suite flaked intermittently with a single opaque signature — the **Nuxt** fixture's `beforeAll` hitting the 300s budget:

```
(fail) Nuxt with TypeScript > (unnamed) [300001.43ms]
       ^ a beforeEach/afterEach hook timed out
```

Reading the raw CI logs of two failing runs showed the same signature came from **two different stalled steps**:

| Run | Last step logged | Hung in |
|---|---|---|
| `27758855853` | `git init done` | `clerk link` — an untimed `fetch()` to the production Clerk API |
| `27721361169` | `fixture copied` | `git init` — a local subprocess |

Root cause: `loggedFetch` (the wrapper for **every** CLI network call) called native `fetch()` with no `AbortSignal`, so a stalled connection hung forever. `--parallel` (8 workers on the 8-vCPU runner) amplified it by bursting ~18 simultaneous PLAPI calls at startup. And no setup step had its own timeout, so *any* stall consumed the full 300s with no hint which step was to blame.

## Fix

- **Root cause (production):** default 60s timeout in `loggedFetch`, composed with any caller signal via `AbortSignal.any` so tighter budgets (e.g. `keyless.ts`'s 15s) still win. A stalled connection now fails fast across every CLI command with a self-diagnosing error (`plapi: request timed out after 60000ms`), not just in tests.
- **Defense-in-depth (e2e):** each setup step (`git init` / `clerk link` / `clerk init` / `npm ci`) runs under a per-step timeout that fails with a labeled error instead of silently eating the 300s budget — covers the local-`git` stall too.
- **Contention:** cap e2e `--parallel=4`; add an explicit `afterEach` cleanup budget and `npm ci --no-audit --no-fund`.
- **Noise:** drop success-path debug traces (the per-step labeled timeouts replace their diagnostic role); keep the failure diagnostics.

## Test plan

- [x] New unit tests for the timeout (fires after the budget; a shorter caller signal isn't masked; fast requests unaffected)
- [x] `bun run test` — 1705 pass / 0 fail
- [x] `typecheck` / `lint` / `format:check` clean
- [ ] E2E suite is secret-gated — confirmed by CI on this PR

> Note: refactored out of a shared working branch; this PR is flake-fix-only (the `--fapi` work lives in its own PRs).

https://claude.ai/code/session_01V1YkHZ2Ad1okwkX9bxTYsd


test/e2e/lib/test-user.ts
```diff
@@ -53,8 +53,6 @@ export async function createTestUser(configDir: string, target: TestUserTarget):
     skip_password_checks: true,
   });
 
-  log(`creating test user: ${email}`);
```
wyattjoh, line 56: These logs should only render when the debug is turned on, is there a reason you've removed them here?

test/e2e/lib/fixture-setup.ts
```diff
@@ -42,6 +42,20 @@ async function copyFixture(fixtureDir: string, projectDir: string): Promise<void
...
+ * npm's default fetch-timeout is 300s, so one stalled registry connection in
+ * `clerk init`'s SDK install or `npm ci` can consume the entire 300s beforeAll
+ * budget. Cap it low so a stalled fetch aborts fast and retries on a fresh
+ * connection — and stays well under the withRetry step budgets. Both npm runs
+ * use projectDir as cwd, so a project-level `.npmrc` covers them.
+ */
+async function writeNpmrc(projectDir: string): Promise<void> {
+  await Bun.write(
+    join(projectDir, ".npmrc"),
+    "fetch-timeout=20000\nfetch-retries=2\nfetch-retry-mintimeout=1000\nfetch-retry-maxtimeout=8000\n",
```
wyattjoh, line 55: Maybe a good idea to extract these timeouts into a constant above, just so that we can have it a little bit more readable.

PR #366, opened 2026-06-26
feat(webhooks): `clerk webhooks` toolkit (V1)

rafa-thayto:
## Summary

Adds the **`clerk webhooks`** command group (V1) — a PLAPI-free local webhooks toolkit. Every command runs with **no auth, no linked project, and no Clerk backend**, so a developer can test webhook delivery and signature verification before any dashboard setup.

Three subcommands:

- **`clerk webhooks listen --forward-to <url>`** — opens a standalone Svix relay tunnel and forwards each delivery to a local handler. `--forward-to` is required. Without `--token` the banner warns that the auto-generated relay token isn't a guaranteed-stable handle and prints the exact `--token` to pin next time; `--token <c_…>` pins an explicit, shareable URL that survives restarts (persisted in config). Flags: `--forward-to` (required), `--token`, `--headers`, `--json`.
- **`clerk webhooks verify`** — verifies a webhook signature offline (HMAC-SHA256), from a saved `listen` event line (`--delivery`) or the four explicit header values. No network calls.
- **`clerk webhooks token`** — generates a valid relay token (`c_` + 10 base62 chars) for `listen --token`. Prints the bare token to stdout so it pipes: `clerk webhooks listen --token "$(clerk webhooks token)"`.

## Agent mode

`listen` emits NDJSON (`ready` / `event` / `reconnecting` lines) under `--json` or in agent mode; `verify` and `token` emit machine-readable output too. Usage/validation errors surface as JSON in agent mode (e.g. a malformed `--token` or missing `--forward-to`).


rafa-thayto:
!snapshot

packages/cli-core/src/commands/webhooks/forward.ts
```diff
@@ -0,0 +1,91 @@
...
+    const key = colonIndex === -1 ? "" : trimmed.slice(0, colonIndex).trim();
+    if (!key) {
+      throwUsageError(
+        `Invalid --headers pair "${trimmed}". Expected key:value. Commas separate pairs and can't appear in values — use multiple --headers flags instead.`,
+      );
+    }
+    headers[key] = trimmed.slice(colonIndex + 1).trim();
+  }
+  return headers;
+}
```
wyattjoh, line 52: Instead, you should enforce a single `--header` or `-H` flag to pass headers one at a time, then using the same first-colon indexing trick you used here, use the `Headers` object:

```ts
const raw = ["Content-Type: text/html", "Set-Cookie: a=1", "Set-Cookie: b=2"];

const headers = new Headers();
for (const header of raw) {
  const colon = header.indexOf(':');
  headers.append(header.slice(0, colon).trim(), header.slice(colon + 1).trim());
}
```

Which will handle/parse the headers more correctly.

packages/cli-core/src/commands/webhooks/forward.ts
```diff
@@ -0,0 +1,91 @@
...
+      method: args.method,
+      headers: args.headers,
+      body: args.body,
+      signal: AbortSignal.timeout(30_000),
+    });
+    const bodyText = await response.text();
+    const headers: Record<string, string> = {};
+    response.headers.forEach((value, key) => {
+      headers[key] = value;
+    });
```
wyattjoh, line 104: While not super relevant here, this approach does lose duplicate `Set‑Cookie` headers because they are emitted one at a time, and you only retain the last one.

packages/cli-core/src/commands/webhooks/listen.test.ts
```diff
@@ -0,0 +1,310 @@
...
+  token: string;
+  onEvent: EventHandler;
+  onTokenRotated: (token: string) => Promise<void>;
+  onReconnect: () => void;
+}
+
+const relayClients: FakeRelayClient[] = [];
+const lastClient = () => relayClients.at(-1);
+
+class FakeRelayClient {
```
wyattjoh: Ideally, when we're creating fake implementations like this, it would be best for us to use or add an `implements` keyword here so that we can ensure that the `RelayClient`, in this case here, that's being exported, is implementing the complete interface and we don't have drift.

packages/cli-core/src/commands/webhooks/relay-client.ts
```diff
@@ -0,0 +1,160 @@
...
+      this.clearProbe();
+      if (this.stopped) return;
+
+      if (event.code === RELAY_CLOSE_TOKEN_COLLISION) {
+        // Another listener holds this token: rotate, persist, redial. Reconnect
+        // through the same backoff as a normal drop so a relay that rejects
+        // every fresh token can't spin a zero-delay reconnect storm.
+        this.token = generateRelayToken();
+        log.debug("relay: token collision (1008), rotating token");
+        void this.options.onTokenRotated(this.token).finally(() => {
```
wyattjoh: `onTokenRotated` returns the `setRelayEntry` promise, but it's chained here as `void ....finally(...)`. `.finally()` passes a rejection through to the promise it returns, and that promise is discarded with `void`, so if persisting the rotated token fails (a config write error) it surfaces as an unhandled promise rejection rather than being handled anywhere.

The reconnect itself is fine (the `.finally` callback always runs), so the only issue is the swallowed-but-unhandled rejection. A `.catch()` makes the intent explicit and keeps the redial:

```ts
void this.options
  .onTokenRotated(this.token)
  .catch(() => log.debug("relay: failed to persist rotated token"))
  .finally(() => {
    if (this.stopped) return;
    setTimeout(() => this.connect(), RELAY_RECONNECT_DELAY_MS);
  });
```

packages/cli-core/src/commands/webhooks/forward.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Delivery headers plus `--headers` extras. Extras may override non-svix
+ * delivery headers, but the delivery's `svix-*` headers always win — they are
+ * what `verify` (and the user's handler) authenticate against.
+ */
+export function buildForwardHeaders(
+  eventHeaders: Record<string, string>,
+  extraHeaders: Record<string, string>,
+): Headers {
+  const headers = new Headers(eventHeaders);
```
wyattjoh: `new Headers(eventHeaders)` forwards every delivery header to the local handler verbatim, including the original `host`. I checked Bun's `fetch`: it does recompute `content-length` (so a stale length from the relay is safe), but it forwards `host` as-is rather than deriving it from the `--forward-to` URL. The handler will see the relay's `Host` instead of `localhost:<port>`.

Harmless for most dev handlers, but frameworks that do host-based routing or build absolute URLs from `Host` can behave oddly. Worth stripping `host` (and the other hop-by-hop framing headers like `connection`/`transfer-encoding`) before forwarding so the request looks like it originated against the forward target.

packages/cli-core/src/commands/webhooks/listen.ts
```diff
@@ -0,0 +1,219 @@
...
+  let client: RelayClient | undefined;
+  let shuttingDown = false;
+
+  // Deliveries can arrive the moment the relay handshake completes, but the
+  // ready banner/line must print first. Gate processing until setup is done;
+  // the SIGINT path also resolves it so the drain can never hang.
+  let resolveSetupGate!: () => void;
+  const setupGate = new Promise<void>((resolve) => {
+    resolveSetupGate = resolve;
+  });
```
wyattjoh: You can actually use the new API here:

```suggestion
  const { promise: setupGate, resolve: resolveSetupGate } = Promise.withResolvers<void>()
```

packages/cli-core/src/commands/webhooks/listen.ts
```diff
@@ -0,0 +1,219 @@
...
+  // Own SIGINT handling, registered BEFORE the socket opens. The global handler
+  // (cli.ts) is a cleanup-free exit(130) and would fire first, so remove it:
+  // close the socket, drain in-flight forwards, then exit 130.
+  process.removeListener("SIGINT", cliSigintHandler);
+  process.on("SIGINT", () => {
+    void (async () => {
+      shuttingDown = true; // MUST precede resolveSetupGate so processDelivery short-circuits
+      resolveSetupGate(); // gated deliveries must settle or the drain hangs
+      client?.stop();
+      await Promise.allSettled([...inFlight, ...(tokenRotationTask ? [tokenRotationTask] : [])]);
```
wyattjoh: The SIGINT path awaits `Promise.allSettled([...inFlight, ...])` before `process.exit(130)`, and each in-flight forward is bounded only by `forward.ts`'s `AbortSignal.timeout(30_000)`. So if the local handler is hung, Ctrl+C blocks the CLI for up to 30s before it exits.

The handler is also registered with `process.on` (not `once`) and has no re-entrancy guard, so a second Ctrl+C just starts another drain awaiting the same `inFlight` promises rather than force-quitting — the usual "press Ctrl+C again to force quit" escape hatch is missing. Consider racing the drain against a short (~2s) timeout before exiting, and/or force-exiting immediately on a second SIGINT.

packages/cli-core/src/commands/webhooks/listen.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    renderForwardResult(outcome, event.method, forwardPath(forwardTo));
+    renderForwardDiagnostics(outcome, svixId);
+  }
+
+  client = new RelayClient({
+    token,
+    onEvent: (event, reply) => {
+      const task = processDelivery(event, reply).catch((error) => {
```
wyattjoh, line 207: `processDelivery` rejections are only `log.debug`'d here, which makes one failure mode silent. `buildForwardHeaders` does `new Headers(event.headers)`, and `Headers` throws on a header name/value it considers invalid. That throw happens before `forwardDelivery`'s own try/catch, so the rejection lands here: the delivery is dropped with no response frame back to the relay (telemetry shows a hung delivery instead of the synthetic 502 you carefully send elsewhere), and nothing prints to the terminal without `--verbose`.

Since the inbox URL is publicly POST-able, a crafted or malformed delivery can reach this path. Consider surfacing a `log.warn` when a delivery is dropped, or building the forward headers defensively (skip/sanitize an individual bad header) so one bad header doesn't silently sink the whole delivery.

packages/cli-core/src/commands/webhooks/relay-protocol.ts
```diff
@@ -0,0 +1,112 @@
...
+      id: frame.data.id,
+      method: frame.data.method ?? "POST",
+      headers: frame.data.headers ?? {},
+      bodyB64: frame.data.body ?? "",
+    },
+  };
+}
+
+export function decodeEventBody(event: RelayEventFrame): string {
+  return Buffer.from(event.bodyB64, "base64").toString("utf8");
```
wyattjoh, line 94: `decodeEventBody` decodes the body as `Buffer.from(bodyB64, "base64").toString("utf8")`, and `forwardDelivery` then re-encodes that string when sending it on (the same pattern is in `verify`'s `parseDeliveryLine` + `verifyWebhookSignature`, which HMACs the decoded string with `.update(str, "utf8")`).

For any body that isn't valid UTF-8, this round-trip is lossy, so the bytes the local handler receives — and the bytes `verify` HMACs — differ from what Svix signed. The handler's own signature check, or `clerk webhooks verify`, would then fail even with the correct secret. Clerk webhook bodies are UTF-8 JSON today so this is theoretical, but given how byte-exact the rest of `verify` is (the trailing-newline footgun note), carrying the body as raw bytes (`Buffer`/`Uint8Array`) end-to-end would make it robust.

packages/cli-core/src/commands/webhooks/listen.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (ndjson) {
+    log.data(buildReadyLine(readyInfo));
+  } else {
+    renderReadyBanner(readyInfo);
+    if (!options.token) renderUnpinnedTokenHint(client.token);
+  }
+  resolveSetupGate();
+
+  // listen never exits 0: it ends via SIGINT (130) or an unrecoverable error (1).
+  await new Promise<never>(() => {});
```
wyattjoh, line 243: I don't necessarily like this approach here, but until we support a more comprehensive abort controller style system, I don't want to make any changes. This is more of a note to myself :)

packages/cli-core/src/commands/webhooks/relay-client.ts
```diff
@@ -0,0 +1,160 @@
...
+  start(): Promise<void> {
+    const FIRST_CONNECT_TIMEOUT_MS = this.options.firstConnectTimeoutMs ?? 30_000;
+    const opened = new Promise<void>((resolve, reject) => {
+      this.rejectFirstOpen = reject;
+      this.resolveFirstOpen = () => {
+        clearTimeout(this.startTimeoutId);
+        this.startTimeoutId = undefined;
+        resolve();
+      };
+    });
```
wyattjoh: Same as mentioned previously, we could probably use the `Promise.withresolvers` here.

packages/cli-core/src/commands/webhooks/shared.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/** Flags inherited from the `webhooks` group. V1 only carries `--json`. */
+export interface WebhooksGlobalOptions {
+  json?: boolean;
+}
+
+/** JSON on stdout when `--json` is set or we're in agent mode. */
+export function shouldOutputJson(options: { json?: boolean }): boolean {
+  return Boolean(options.json) || isAgent();
+}
```
wyattjoh, line 11: I feel like this is used elsewhere in the code base already?

packages/cli-core/src/commands/webhooks/token.ts
```diff
@@ -0,0 +1,30 @@
...
+export async function webhooksToken(options: WebhooksTokenOptions = {}): Promise<void> {
+  const token = generateRelayToken();
+  if (options.json) {
+    log.data(JSON.stringify({ token }));
+    return;
+  }
+  log.data(token);
+  if (isAgent()) return;
+  await outro([
+    `Pin it: clerk webhooks listen --token ${token} --forward-to <url>`,
```
wyattjoh: `Pin it` isn't quite clear, can we reword this?

packages/cli-core/src/commands/webhooks/verify.ts
```diff
@@ -0,0 +1,239 @@
...
+      return await Bun.file(path).text();
+    } catch (err) {
+      const reason = err instanceof Error ? `: ${err.message}` : "";
+      throw new CliError(`Could not read ${path}${reason}`, { code: ERROR_CODE.FILE_NOT_FOUND });
+    }
+  }
+  return throwUsageError(
+    `${flag} takes @file or - for stdin (inline values get mangled by shells).`,
+  );
+}
```
wyattjoh, line 123: Is this something that we should abstract for the general input field rather than having this be isolated to the webhooks command?

packages/cli-core/src/cli.ts
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bun
 import { createProgram, runProgram } from "./cli-program.ts";
-import { EXIT_CODE } from "./lib/errors.ts";
-process.on("SIGINT", () => process.exit(EXIT_CODE.SIGINT));
+import { cliSigintHandler } from "./lib/signals.ts";
```
wyattjoh: For global handlers like this, I think we should have it all uppercase instead.

```suggestion
import { CLI_SIGINT_HANDLER } from "./lib/signals.ts";
```

.oxlintrc.json
```diff
@@ -10,6 +10,8 @@
       "files": [
         "packages/cli-core/src/cli.ts",
         "packages/cli-core/src/cli-program.ts",
+        "packages/cli-core/src/lib/signals.ts",
+        "packages/cli-core/src/commands/webhooks/listen.ts",
```
wyattjoh, line 14: It's for reasons like this, where we essentially disabled the lint rules, that makes me want to think the promise that never resolves is a bad long-term solution.

No changes required here. I just wanted to call this out.

wyattjoh (approved):

rafa-thayto:
!snapshot

PR #373, opened 2026-07-03
feat: add clerk impersonate command with instance-aware targeting

rafa-thayto:
## Summary

- Add `clerk impersonate` (alias `clerk imp`) to create a short-lived actor token and print the sign-in URL that logs you in as another user, plus `clerk impersonate revoke` to cancel a pending token. Both require `clerk login` and stamp the token's actor as `cli:<email>[+context]` for auditing.
- User resolution accepts a `user_...` ID, an exact email (`email_address` filter), or a fuzzy term (`query`), with an interactive picker for ambiguous or omitted input. Zero-match errors name the searched app and instance (e.g. `No user found matching "x" on My Application (development).`).
- Never silently default to the development instance in interactive flows: when the resolved app has more than one instance and no `--instance` flag pins one, human mode prompts "Select an instance to use:" (choices shown as `ins_xxx (production|development)`). Agent mode never prompts and keeps the dev default.
- Instances targeted by raw ID are now labeled by their environment type, so the production impersonation warning (MFA bypass / quota) fires even with `--instance ins_xxx`.
- Human-mode guardrails: confirmation prompt defaulting to No, an explicit warning on production instances, and friendly errors for plan-gating (402) and quota exhaustion (422).

## Test plan

- [x] `bun run format` / `bun run lint` / `bun run typecheck`
- [x] `bun run test` — 1748 tests pass (unit + integration, including new coverage for the resolver, instance prompt, actor stamping, revoke, and shell completion)
- [x] Verified against a real app: exact-email lookup on the wrong instance now reports the searched instance; `--instance production` resolves the user and reaches the actor-token API
- [ ] E2E (`bun run test:e2e`) runs in CI

https://claude.ai/code/session_01HoHwrFq6AhrW13j2PkBQjc


packages/cli-core/src/commands/impersonate/impersonate.ts
```diff
@@ -0,0 +1,177 @@
...
+    const limit = typeof meta.limit === "number" ? meta.limit : undefined;
+    const used = typeof meta.used === "number" ? meta.used : undefined;
+    const quota =
+      limit !== undefined && used !== undefined
+        ? ` (used ${used}/${limit} this billing period)`
+        : "";
+    return new CliError(`Impersonation limit exceeded${quota}.`, {
+      code: ERROR_CODE.IMPERSONATION_LIMIT_EXCEEDED,
+    });
+  }
```
wyattjoh: nit: This seems like something that we should create a standardized error class in the future for errors related to billing/usage.

packages/cli-core/src/commands/impersonate/impersonate.ts
```diff
@@ -0,0 +1,177 @@
...
+    if (!proceed) {
+      throwUserAbort();
+    }
+  }
+
+  const payload = {
+    user_id: userId,
+    actor: { sub: actor.sub, iss: actor.iss },
+    expires_in_seconds: expiresIn,
+  };
```
wyattjoh: This should have a named type to it and not just an arbitrary construction, because it depends on an API contract.

Maybe we should elevate this to a library function in BAPI.

packages/cli-core/src/commands/impersonate/resolve-user.ts
```diff
@@ -0,0 +1,93 @@
...
+  } else {
+    searchParams.set("query", user);
+  }
+  searchParams.set("limit", String(CANDIDATE_LIMIT + 1));
+
+  const response = await bapiRequest({
+    method: "GET",
+    path: `/users?${searchParams}`,
+    secretKey,
+  });
```
wyattjoh: Another good candidate for elevating to a named function. I don't like having these library-esque function calls nested inside commands.

packages/cli-core/src/commands/users/interactive/instance-context.test.ts
```diff
@@ -22,15 +27,21 @@ mock.module("../../../lib/config.ts", () => ({
...
     },
-    _instance?: string,
+    instance?: string,
   ) => {
-    const development = app.instances.find((i) => i.environment_type === "development");
-    if (!development) return { found: false, instanceId: "unknown", instanceLabel: "unknown" };
+    // Mirrors the real resolver: an env-type or instance-id hint wins,
+    // otherwise fall back to the development instance. Labels use the
+    // matched instance's environment type.
+    const hint = instance ?? "development";
```
wyattjoh, line 35: I don't think we should be defaulting to development here, because I thought the whole point of the interactive instance picker was to not default.

rafa-thayto: this case is when the user only has the dev instance, so they don't need to select

wyattjoh: I'm not super comfortable with a flag that is data dependent like this, where the default depends on the presence of the production instance. It breaks a repeatability guarantee that running one command will produce the same kind of result. I think it should be explicit here to prevent confusion.

rafa-thayto: ok, that's fair I changed my mind!

wyattjoh (approved):

PR #374, opened 2026-07-06
fix(init): scaffold ssr.noExternal for React Router v8 projects

manovotny:
## Summary

`clerk init` on a React Router 8 project scaffolds an app that is broken in dev out of the box: every request fails during SSR with `useNavigate() may be used only in the context of a <Router> component.` This came out of root-causing a public [field report](https://x.com/Compurnicus/status/2073828066649117008) where an agent couldn't one-shot a Clerk + React Router app.

RR8 ships development/production conditional exports. `react-router dev` externalizes `@clerk/react-router` for SSR, so Node resolves react-router's *production* build for Clerk while the app code gets the *development* build through Vite — two module instances, two Router contexts, and Clerk's `useNavigate()` call inside `ClerkProvider` throws. Upstream: remix-run/react-router#15232. The scaffolder already handles the v7/v8 split for the `v8_middleware` flag but never touched the vite config.

## Changes

- Scaffold `ssr: { noExternal: ["@clerk/react-router"] }` into `vite.config.{ts,js,mts,mjs}` when `react-router` is v8+ (or unparseable, matching how the `v8_middleware` gate already treats unknown versions). v7 has no conditional exports and is left alone.
- Uses magicast with a string-replace fallback, mirroring the astro scaffolder; appends to an existing `ssr.noExternal` array instead of clobbering it; skips when the config already references `@clerk/react-router`.
- Emits a manual post-instruction (with the verbatim error and upstream link) when the vite config is missing or uses a function form that can't be safely modified.
- 6 new unit tests covering: fresh add, append-to-existing-array, already-present skip, v7 no-op, missing-config instruction, and function-form instruction.

## Verification

- Reproduced in a scratch app (`create-react-router@latest` → RR 8.0.0 + `@clerk/react-router@3.5.5`): dev SSR 500s without the entry, HTTP 200 with it; `react-router build` unaffected.
- `bun run format`, `lint`, `typecheck` clean; `react-router.test.ts` 21/21 pass. (Full `bun run test` has 363 pre-existing failures in credential-store/host-execution on clean `main` in my sandbox — identical count with and without this change.)

## Not planned (follow-up candidates)

- The e2e fixture (`test/e2e/fixtures/react-router`) pins `react-router@7.15.0`, so CI never exercises v8 — worth adding an RR8 fixture so the next regression of this class is caught.

## References

- Upstream issue: https://github.com/remix-run/react-router/issues/15232
- SDK v8 support (where Clerk's own integration template gained this same workaround): https://github.com/clerk/javascript/pull/8972
- Companion PRs: clerk/skills#55 and a clerk-docs quickstart tweak.

### DIFF
--- a/packages/cli-core/src/commands/init/frameworks/react-router.test.ts
+++ b/packages/cli-core/src/commands/init/frameworks/react-router.test.ts
@@ -434,3 +434,116 @@ export default [index("routes/home.tsx")] satisfies RouteConfig;
     'route("($locale)/sign-up/*", "routes/($locale).sign-up.tsx")',
   );
 });
+
+test("adds ssr.noExternal to vite.config.ts for React Router 8", async () => {
+  await mkdir(join(tempDir, "app"), { recursive: true });
+  await Bun.write(
+    join(tempDir, "vite.config.ts"),
+    `import { reactRouter } from "@react-router/dev/vite";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [reactRouter()],
+});
+`,
+  );
+
+  const plan = await reactRouter.scaffold(makeCtx({ deps: { "react-router": "8.0.0" } }));
+  const viteAction = plan.actions.find((action) => action.path === "vite.config.ts");
+
+  expect(viteAction?.type).toBe("modify");
+  if (viteAction?.type !== "modify") throw new Error("Expected vite config modify action");
+
+  expect(viteAction.content).toContain("noExternal");
+  expect(viteAction.content).toContain("@clerk/react-router");
+});
+
+test("appends to an existing ssr.noExternal array for React Router 8", async () => {
+  await mkdir(join(tempDir, "app"), { recursive: true });
+  await Bun.write(
+    join(tempDir, "vite.config.ts"),
+    `import { defineConfig } from "vite";
+
+export default defineConfig({
+  ssr: {
+    noExternal: ["some-other-pkg"],
+  },
+});
+`,
+  );
+
+  const plan = await reactRouter.scaffold(makeCtx({ deps: { "react-router": "8.0.0" } }));
+  const viteAction = plan.actions.find((action) => action.path === "vite.config.ts");
+
+  expect(viteAction?.type).toBe("modify");
+  if (viteAction?.type !== "modify") throw new Error("Expected vite config modify action");
+
+  expect(viteAction.content).toContain("some-other-pkg");
+  expect(viteAction.content).toContain("@clerk/react-router");
+});
+
+test("skips vite config when @clerk/react-router is already in ssr.noExternal", async () => {
+  await mkdir(join(tempDir, "app"), { recursive: true });
+  await Bun.write(
+    join(tempDir, "vite.config.ts"),
+    `import { defineConfig } from "vite";
+
+export default defineConfig({
+  ssr: {
+    noExternal: ["@clerk/react-router"],
+  },
+});
+`,
+  );
+
+  const plan = await reactRouter.scaffold(makeCtx({ deps: { "react-router": "8.0.0" } }));
+  const viteAction = plan.actions.find((action) => action.path === "vite.config.ts");
+
+  expect(viteAction?.type).toBe("skip");
+});
+
+test("does not touch vite config for React Router 7", async () => {
+  await mkdir(join(tempDir, "app"), { recursive: true });
+  await Bun.write(
+    join(tempDir, "vite.config.ts"),
+    `import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [],
+});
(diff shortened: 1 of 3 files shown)


manovotny:
Companion PRs, now all open:

- clerk/skills#55 — skill updated for React Router v8 (nonexistent `ClerkApp` removed, workaround documented)
- clerk/clerk#2872 — quickstart callout now names the verbatim error so searches land on the fix (migrated from clerk/clerk-docs#3504 as part of the docs repo move)

packages/cli-core/src/commands/init/frameworks/react-router.ts
```diff
@@ -382,6 +382,81 @@ function scaffoldConfig(ctx: ProjectContext): Promise<FileAction | null> {
...
+
+    if (!defaultExport.ssr) defaultExport.ssr = {};
+    if (!defaultExport.ssr.noExternal) defaultExport.ssr.noExternal = [];
+    if (!Array.isArray(defaultExport.ssr.noExternal)) return content;
+
+    defaultExport.ssr.noExternal.push("@clerk/react-router");
+    return mod.generate().code;
+  } catch {
+    if (content.includes("defineConfig")) {
+      return content.replace(
```
rafa-thayto: Heads up, this fallback runs for every `defineConfig({...})` config, not just rare ones, because the AST branch above throws when it sets `.ssr` on the call expression. So when a config already has an `ssr` block, this injects a second one, and since the later key wins, `@clerk/react-router` silently never applies.

The fix is to reach into the call's argument with magicast instead of string-injecting, so you always update the existing `ssr` rather than duplicating it. This drops the regex fallback entirely:

```ts
function addClerkNoExternal(content: string): string {
  try {
    const mod = parseModule(content);
    const def = mod.exports.default as any;
    if (!def || typeof def !== "object") return content;
    // export default defineConfig({...}) -> operate on the call's first arg
    const config = def.$type === "function-call" ? def.$args[0] : def;
    if (!config || typeof config !== "object") return content;
    if (!config.ssr) config.ssr = {};
    if (config.ssr.noExternal === undefined) config.ssr.noExternal = [];
    if (!Array.isArray(config.ssr.noExternal)) return content; // boolean/string, leave it and let the caller emit the manual instruction
    if (!config.ssr.noExternal.includes("@clerk/react-router")) {
      config.ssr.noExternal.push("@clerk/react-router");
    }
    return mod.generate().code;
  } catch {
    return content;
  }
}
```

Verified against `defineConfig` with and without an existing `ssr`, a plain object export, a boolean `noExternal`, and a config that already lists `@clerk/react-router`. Each produces a single `ssr` block with no duplicates.

packages/cli-core/src/commands/init/frameworks/react-router.test.ts
```diff
@@ -434,3 +434,116 @@ export default [index("routes/home.tsx")] satisfies RouteConfig;
...
+  );
+
+  const plan = await reactRouter.scaffold(makeCtx({ deps: { "react-router": "8.0.0" } }));
+  const viteAction = plan.actions.find((action) => action.path === "vite.config.ts");
+
+  expect(viteAction?.type).toBe("modify");
+  if (viteAction?.type !== "modify") throw new Error("Expected vite config modify action");
+
+  expect(viteAction.content).toContain("some-other-pkg");
+  expect(viteAction.content).toContain("@clerk/react-router");
```
rafa-thayto, line 483: This stays green even when the output ends up with two `ssr` keys and the clerk entry gets shadowed, since it only checks substrings. Assert there's a single `ssr` block too so a duplicate-key regression fails here.

```suggestion
  expect(viteAction.content).toContain("some-other-pkg");
  expect(viteAction.content).toContain("@clerk/react-router");
  expect(viteAction.content.match(/ssr\s*:/g)?.length).toBe(1);
```

rafa-thayto (changes requested):

packages/cli-core/src/commands/init/frameworks/react-router.ts
manovotny: Confirmed — reproduced both halves: the set trap throws on the function-call proxy (so the fallback ran for every `defineConfig`), and the injected duplicate `ssr` loses to the original, silently dropping the entry. Applied your version verbatim and verified it against defineConfig with/without `ssr`, plain object export, boolean `noExternal`, function-form, and already-present entry — single `ssr` block in every case. Fixed in 4d247bc.

packages/cli-core/src/commands/init/frameworks/react-router.test.ts
manovotny: Applied, and mirrored it into the no-existing-`ssr` test as well. Fixed in 4d247bc.

rafa-thayto (approved):

PR #382, opened 2026-07-14
fix(impersonate): revoke the live session when the actor token was already accepted

rafa-thayto:
## Summary

`clerk imp revoke act_...` failed with `(400): cannot revoke` once the impersonation sign-in URL had been opened: BAPI's `POST /v1/actor_tokens/{id}/revoke` only revokes **pending** tokens, and accepting the ticket flips the token to `accepted` while the live impersonation persists as a session. There was no way to end that session from the CLI.

- `clerk imp revoke` now detects the 400 and falls back to session revocation: `GET /v1/sessions?user_id=<user>&status=active`, filtered to sessions whose `actor.sub` matches the operator's `cli:<email>` stamp (prefix match, so `--actor` `+<context>` variants also match), then `POST /v1/sessions/{id}/revoke` for each — sessions started by other operators are never touched.
- New `--user <user_id>` option on `clerk imp revoke` (required for the fallback — BAPI has no `GET /v1/actor_tokens/{id}`, so the target user can't be recovered from the token ID). Without it, the fallback errors with the new `actor_token_already_accepted` code and shows how to re-run; no matching session errors with `impersonation_session_not_found`.
- The `Revoke with:` hint printed at creation now includes `--user <user_id>` so the copied command works in both the pending and accepted states.
- Agent mode emits `{ "id": "<act_id>", "status": "accepted", "revokedSessionIds": ["sess_..."] }` when the fallback runs, and never prompts.
- New `lib/sessions.ts` BAPI client (tolerates both plain-array and paginated `{ data }` list shapes), `isActorTokenNotRevocableError` helper in `lib/actor-tokens.ts`, README + changeset (minor) updates.

## Test plan

- [x] 8 new unit tests in `revoke.test.ts` (TDD, watched fail first): fallback with/without `--user`, actor-stamp filtering incl. `+<context>` and foreign/plain sessions, agent-mode JSON payload, no-match error, paginated list shape, non-400 passthrough, `--user` format validation
- [x] `impersonate.test.ts` asserts the creation hint includes `--user <user_id>`
- [x] `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test` (1904 pass) all green
- [x] `bun run test:e2e:op` — could not run from this environment (1Password vault unavailable); CI runs `test:e2e`

### DIFF
--- a/.changeset/imp-revoke-fix.md
+++ b/.changeset/imp-revoke-fix.md
@@ -1,0 +1,5 @@
+---
+"clerk": minor
+---
+
+Revoke live impersonation sessions with `clerk imp revoke <act_id> --user <user_id>`. When the actor token was already accepted (the sign-in URL was opened), the token itself can no longer be revoked — the command now falls back to finding and revoking the active session(s) your impersonation created. The "Revoke with" hint printed by `clerk imp` includes `--user` so the command works in both states.
--- a/packages/cli-core/src/lib/sessions.ts
+++ b/packages/cli-core/src/lib/sessions.ts
@@ -1,0 +1,64 @@
+/**
+ * Backend API (BAPI) sessions client.
+ *
+ * Sessions created from an actor token carry the initiating actor in
+ * `session.actor` (the CLI stamps `cli:<login-email>`), which is how
+ * `clerk impersonate revoke` finds and ends a live impersonation once the
+ * actor token itself was accepted and can no longer be revoked.
+ */
+
+import { bapiRequest } from "./bapi.ts";
+
+/** The actor claim stamped onto sessions created from an actor token. */
+export type SessionActor = {
+  sub?: string;
+  iss?: string;
+};
+
+/** The subset of BAPI's Session object the CLI consumes. */
+export type Session = {
+  id: string;
+  status?: string;
+  actor?: SessionActor | null;
+};
+
+/** Result of revoking a session. Fields are optional — BAPI may echo them. */
+export type RevokedSession = {
+  id?: string;
+  status?: string;
+};
+
+export async function listUserSessions(
+  secretKey: string,
+  query: { userId: string; status?: string },
+): Promise<Session[]> {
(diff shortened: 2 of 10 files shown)


.changeset/imp-revoke-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+"clerk": minor
+---
+
+Revoke live impersonation sessions with `clerk imp revoke <act_id> --user <user_id>`. When the actor token was already accepted (the sign-in URL was opened), the token itself can no longer be revoked — the command now falls back to finding and revoking the active session(s) your impersonation created. The "Revoke with" hint printed by `clerk imp` includes `--user` so the command works in both states.
```
wyattjoh, line 5: I'd replace `accepted` with `consumed` to be clear that the actor token was one time use.

packages/cli-core/src/lib/sessions.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  // BAPI list endpoints exist in two shapes: a plain array and a paginated
+  // `{ data: [...] }` envelope. Accept both.
+  const body = response.body;
+  if (Array.isArray(body)) {
+    return body as Session[];
+  }
+  const data = (body as { data?: Session[] } | null)?.data;
+  return Array.isArray(data) ? data : [];
+}
```
wyattjoh, line 50: This list endpoint only exists in a single state, let's simplify this here.

.changeset/imp-revoke-fix.md
rafa-thayto: going to keep `accepted` since that's the actual BAPI token status, and it's already in the error code (`actor_token_already_accepted`) and the agent JSON output. Renaming it only in the changeset would make the changelog say something different from what ships. It also matches how invitations name this state (`pending | accepted | revoked | expired`).

wyattjoh (approved):

PR #390, opened 2026-07-23
fix(orgs): enable organization creation defaults with `--auto-create`

manovotny:
## Summary

Found while reviewing clerk/clerk#2961, where the docs work around it by telling agents to patch the parent flag manually.

`clerk enable orgs --auto-create` patched only `organization_creation_defaults.automatic_organization_creation.enabled`. The API gates auto-creation on the umbrella `organization_creation_defaults.enabled` too, so on instances where creation defaults were never enabled the command succeeded — and no organization was ever auto-created. 

Perhaps it's this way for some reason, but it tripped me up when verifying the docs contribution.

## Changes

- `--auto-create` now sends both flags in one patch.
- Sending both together is safe: the API's enable transition seeds defaults only for sub-settings not in the request, so the explicit auto-create value wins and the remaining sub-settings get the same seeds the Dashboard flow produces. Re-sending `enabled: true` on an already-enabled instance is a no-op, so the command stays idempotent.
- Payload test now asserts both flags, and the command README documents the behavior.

## References

- clerk/clerk#2961

### DIFF
--- a/.changeset/orgs-auto-create-creation-defaults.md
+++ b/.changeset/orgs-auto-create-creation-defaults.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Fix `clerk enable orgs --auto-create` so organizations are actually auto-created for new users. The flag now also enables organization creation defaults, which the API requires before honoring the automatic creation setting.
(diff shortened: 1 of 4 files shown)


rafa-thayto (approved):
nice!

PR #394, opened 2026-07-24
feat(init): close the framework scaffolding gaps for expo, express, fastify, ios, and android

rafa-thayto:
## Summary
- **Expo**: wrap the expo-router root layout with `<ClerkProvider>` + secure token cache (existing layouts are modified in place, missing ones created), and add an Expo starter to the bootstrap registry so `clerk init --starter --framework expo` scaffolds a new app via `create-expo-app`
- **Express / Fastify**: wire `clerkMiddleware()` / `clerkPlugin` into the server entry file (ESM and CJS, including inline-require and type-annotated forms) via a shared `node-server.ts` scaffolder; TypeScript Express apps also get the `@clerk/express/env` request type augmentation
- **iOS / Android**: detect native platforms without a package.json (`*.xcodeproj`/`*.xcworkspace` directory bundles, `AndroidManifest.xml`), accept `--framework ios|android`, skip the npm bootstrap/install paths, and print the exact SDK quickstart steps (validated against the official Clerk docs, including the SwiftUI `.environment(Clerk.shared)` step)
- **Hardening** (second commit): comment/string masking so quoted or commented-out creation statements can't hijack insertions, ClerkProvider wrapping scoped to the default export (expo-router `ErrorBoundary` exports are no longer wrapped), multi-line-import-safe insertion, directory-only wildcard markers, and style-consistent generated imports — every fix has a regression test written first

## Test Plan
- [x] `bun run format:check` / `bun run lint` / `bun run typecheck` / `bun run test` — 2201 tests pass (28 new regression tests on this branch)
- [x] Hands-on verification: real fixture apps scaffolded by the actual code, dependencies installed, servers started and curled — Express ESM/CJS and Fastify respond 200 with `x-clerk-auth-status` headers stamped by the live middleware
- [x] Real `create-expo-app` run with the registry's exact flags, chained into the scaffolder; `tsc --noEmit` delta vs pristine template = zero; `expo export --platform web` bundles and renders with clerk-js booting in-browser
- [x] Real CLI runs for iOS/Android fixtures in an isolated unauthenticated environment: correct detection, zero file writes, correct guidance, exit 0; negative fixtures (bare `Package.swift`, `build.gradle.kts`, stray `.xcodeproj` file) correctly not detected
- [x] Scaffold output diffed against the official Clerk quickstarts for all five platforms (matches; the one gap found — the iOS SwiftUI environment step — is fixed here)

## Notes
No releasing changes; changeset included (`minor`). Native platforms are instruction-only by design — their SDKs (SPM/Gradle) can't be safely installed by a JS package manager, so init links the app, pulls keys, and prints the documented setup steps instead.


wyattjoh (commented):
Nice work on the masking layer in `source-scan.ts` and the shared `node-server.ts` scaffolder — the test coverage on the Express/Fastify entry shapes is genuinely thorough. Two things I'd like to see addressed before merge: the ErrorBoundary fix doesn't hold for arrow-function default exports, and the iOS/Android path writes a live secret key into a repo that almost certainly doesn't gitignore it.

packages/cli-core/src/commands/init/frameworks/expo.ts
```diff
@@ -0,0 +1,225 @@
...
+ */
+function findDefaultExportBody(
+  content: string,
+  masked: string,
+): { start: number; end: number } | null {
+  let fnIdx = masked.search(/export\s+default\s+(?:async\s+)?function\b/);
+
+  if (fnIdx === -1) {
+    // `export default RootLayout;` referencing a function declared elsewhere.
+    const ref = /export\s+default\s+(\w+)/.exec(masked);
```
wyattjoh: `findDefaultExportBody` only resolves two shapes: `export default function ...` and `export default Ident` where `Ident` is a **function declaration**. The very common React Native shape — an arrow function — resolves to `null` on both counts, and the caller then falls back to `from = 0, to = content.length` and runs `lastIndexOf("return (")` over the whole file. That is exactly the failure this commit set out to fix, just via a different door.

Running `wrapLastReturnWithProvider` against:

```tsx
import { Slot } from "expo-router";

const RootLayout = () => {
  return (
    <Slot />
  );
};

export default RootLayout;

export function ErrorBoundary({ error }) {
  return (
    <Text>{error.message}</Text>
  );
}
```

produces the provider inside `ErrorBoundary`, leaving `RootLayout` unwrapped:

```tsx
export function ErrorBoundary({ error }) {
  return (
    <ClerkProvider publishableKey={publishableKey} tokenCache={tokenCache}>
      <Text>{error.message}</Text>
    </ClerkProvider>
  );
}
```

Same result for `export default () => { ... }`. The user gets a layout that looks scaffolded (imports and the `publishableKey` block are added at the top) but every Clerk hook throws at runtime because the tree was never wrapped.

Two fixes worth considering. Extend the reference lookup to also match `(?:const|let|var)\s+${ref[1]}\s*=` and, when the initializer is an arrow, take the body from the `=>`'s `{` (or treat a concise `=> (` body as the return site). And make the whole-file fallback fail closed: if the default export can't be resolved, return `null` so the caller emits the "unsupported shape" skip instead of wrapping whatever return happens to be last in the file. The existing test at `expo.test.ts:238` should get an arrow-function sibling.

packages/cli-core/src/lib/framework.ts
```diff
@@ -112,6 +126,43 @@ export const FRAMEWORK_MAP: FrameworkInfo[] = [
...
+// must still detect as Expo).
+export const NATIVE_FRAMEWORK_MAP: NativeFrameworkEntry[] = [
+  {
+    dep: "ios",
+    name: "iOS (Swift)",
+    sdk: "ClerkKit",
+    envVar: "CLERK_PUBLISHABLE_KEY",
+    envFile: ".env",
+    ecosystem: "swift",
+    // Xcode project/workspace bundles only — a bare Package.swift could be a
```
wyattjoh, line 147: With `envFile: ".env"` on the native entries, a default `clerk init` in an Xcode or Android Studio project runs the `authenticate` strategy, which calls `pull()`. `pull()` writes both `CLERK_PUBLISHABLE_KEY` **and** `CLERK_SECRET_KEY` (`commands/env/pull.ts` adds the secret whenever `matched.secret_key` is present), and nothing in the CLI adds `.env` to `.gitignore` — `addGitignoreEntry` in `lib/keyless.ts` is only wired to the keyless `.clerk/` path.

For the npm frameworks this is survivable because the generators emit a `.gitignore` covering `.env*`. GitHub's Swift and Android `.gitignore` templates do not mention `.env` at all, so on the platforms this PR adds, the default outcome is a live secret key sitting in a tracked file.

It's also a key these projects don't need. The quickstart steps this scaffolder prints only ever reference the publishable key (`Clerk.configure(publishableKey:)` / `Clerk.initialize(..., publishableKey =)`); a mobile client has no server to use `sk_` from.

Suggested fix: for `ecosystem !== "npm"`, either skip the secret key entirely on the pull, or at minimum append `.env` to `.gitignore` the way the keyless flow does and warn when the file isn't ignored. A post-instruction telling the user to gitignore `.env` would be the cheap version, but silent-by-default is the wrong failure mode for a credential.

packages/cli-core/src/commands/init/frameworks/expo.ts
```diff
@@ -0,0 +1,225 @@
...
+      type: "create",
+      path: `${ctx.srcDir ? "src/" : ""}app/_layout.${ext}`,
+      content: newLayoutContent(ctx.envFile),
+      description: "Create root layout with ClerkProvider and token cache",
+    };
+  }
+
+  const content = await Bun.file(join(ctx.cwd, layoutPath)).text();
+
+  if (content.includes("ClerkProvider")) {
```
wyattjoh: This is the one anchor search in the Expo scaffolder that still runs against raw `content` rather than the masked source. A layout with a commented-out or documented `<ClerkProvider>` — e.g. a `// TODO: wrap in <ClerkProvider>` left over from a manual attempt — makes init skip the file entirely with "Already has ClerkProvider", and the user gets no wiring and no warning that it bailed.

Given the rest of the commit is specifically about not letting comments and string literals hijack scaffolding decisions, this one should use the mask too:

```ts
if (maskCommentsAndStrings(content).includes("ClerkProvider")) {
```

(The trade-off is one extra mask pass on a file that's about to be masked anyway inside `wrapLastReturnWithProvider`, so it's close to free if you hoist it.)

packages/cli-core/src/commands/init/frameworks/node-server.ts
```diff
@@ -0,0 +1,187 @@
...
+ */
+type ServerEntry = { action: FileAction | null; wired: boolean };
+
+/** Entry file candidates for Node server projects, most specific first. */
+const ENTRY_BASENAMES = ["index", "server", "app", "main"];
+const ENTRY_EXTS = ["ts", "mts", "js", "mjs", "cjs"];
+
+function entryCandidates(): string[] {
+  const names = ENTRY_BASENAMES.flatMap((base) => ENTRY_EXTS.map((ext) => `${base}.${ext}`));
+  return [...names.map((name) => `src/${name}`), ...names];
```
wyattjoh: `entryCandidates()` puts every `src/`-prefixed name ahead of every root-level name, so the ordering is `src/index`, `src/server`, `src/app`, `src/main`, then `index`, `server`, `app`, `main`. A project whose real entry is a root `index.js` but that also happens to have a `src/app.ts` module resolves to `src/app.ts`, and `clerkMiddleware()` lands in a file that never runs.

The `package.json#main` check covers this when it's set, but plenty of Express/Fastify apps use `scripts.start` or a bare `node index.js` and leave `main` at its default. Interleaving by basename instead of by directory matches intent better:

```ts
return ENTRY_BASENAMES.flatMap((base) =>
  ENTRY_EXTS.flatMap((ext) => [`src/${base}.${ext}`, `${base}.${ext}`]),
);
```

That keeps "most specific basename first" (the stated rule) while no longer letting an unrelated `src/app.ts` outrank a root `index.js`.

packages/cli-core/src/commands/init/frameworks/expo.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    const inner = content.slice(openIdx + 1, closeIdx - 1);
+    return content.slice(0, openIdx) + wrapJsx(inner) + content.slice(closeIdx);
+  }
+
+  // Single-line form: `return <Slot />;` — multi-line JSX without parens is
+  // invalid JS (ASI), so the statement always ends on the same line.
+  const singleLine = /return\s+(<.*>)\s*;?\s*$/gm;
+  let match: RegExpExecArray | null = null;
+  for (const m of content.slice(from, to).matchAll(singleLine)) {
```
wyattjoh: Unlike the `return (` branch above it, the single-line branch iterates `content.slice(from, to)` rather than the masked slice — the mask is only used afterward for the `masked[from + m.index] === "r"` guard. That works, but it means the greedy `<.*>` can run into a trailing comment.

`return <Slot />; // TODO: see docs -> here` matches nothing (the greedy `<.*>` swallows through `->` and then `\s*;?\s*$` fails; backtracking to `<Slot />` fails on the trailing comment), so `wrapLastReturnWithProvider` returns `null` and the layout is skipped as "unsupported shape" purely because of a comment. Matching against the masked slice — where the comment is blanked — makes both the search and the guard consistent and fixes the case.

packages/cli-core/src/commands/init/frameworks/expo.ts
```diff
@@ -0,0 +1,225 @@
...
+}
+
+/**
+ * Locate the body of the default-exported function so return-wrapping never
+ * targets a sibling export — expo-router layouts commonly export additional
+ * components (e.g. the documented `ErrorBoundary`) from the same file.
+ * Returns null when the default export isn't a resolvable function.
+ */
+function findDefaultExportBody(
+  content: string,
```
wyattjoh: `findDefaultExportBody(content, masked)` never reads `content` — every lookup in the body goes through `masked`. Drop the parameter (and the argument at the call site) so the signature stops implying the function needs both.

packages/cli-core/src/commands/init/frameworks/node-server.ts
```diff
@@ -0,0 +1,187 @@
...
+  ctx: ProjectContext,
+  config: ServerFrameworkConfig,
+): Promise<ScaffoldPlan> {
+  const { action, wired } = await scaffoldServerEntry(ctx, config);
+
+  return {
+    actions: action ? [action] : [],
+    postInstructions: [
+      ...(wired ? [] : [`${config.manualWiring} See: ${config.docsUrl}`]),
+      `Ensure ${ctx.framework.envVar} and CLERK_SECRET_KEY are set in your ${ctx.envFile} (pulled via \`clerk env pull\`), and load them before Clerk imports — e.g. \`node --env-file=${ctx.envFile} index.js\``,
```
wyattjoh: The instruction always suggests `node --env-file=<envFile> index.js` even though the scaffolder just resolved the actual entry path, which is frequently `src/server.ts` or `src/index.ts`. For a TypeScript project the suggested command is wrong twice over. `scaffoldServerEntry` already knows `entryPath` — threading it out alongside `wired` and interpolating it here would make the example copy-pasteable.

packages/cli-core/src/commands/init/frameworks/express.ts
```diff
@@ -0,0 +1,58 @@
...
+  creationPattern:
+    /(?:const|let|var)\s+(\w+)(?:\s*:\s*[\w$.]+(?:<[^>]*>)?)?\s*=\s*(?:express|require\(\s*["']express["']\s*\))\s*\(\s*\)/,
+  frameworkPackage: "express",
+  attachStatement: (appVar) => `${appVar}.use(clerkMiddleware());`,
+  description: "Add clerkMiddleware() to Express app",
+  docsUrl: "https://clerk.com/docs/expressjs/getting-started/quickstart",
+  manualWiring: "Add `app.use(clerkMiddleware())` from @clerk/express to your server entry file.",
+};
+
+const TYPES_REFERENCE_PATH = "types/globals.d.ts";
```
wyattjoh, line 19: `types/globals.d.ts` matches the official quickstart, but the reference only takes effect if the project's `tsconfig.json` actually picks the file up. A tsconfig with `"include": ["src"]` (common enough in Express templates) never loads it, so the `Request` augmentation silently doesn't apply and the user hits type errors on `req.auth` with a file on disk that looks like it should have fixed it.

Either co-locate the reference under the source root when `src/` is in use, or add a post-instruction noting the file must be covered by `include`.

packages/cli-core/src/commands/init/frameworks/ios.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Docs: https://clerk.com/docs/ios/getting-started/quickstart
+ */
+export const ios: FrameworkScaffold = {
+  name: "iOS (Swift)",
+  dep: "ios",
+
+  matches: (ctx) => ctx.framework.dep === "ios",
+
+  async scaffold(ctx: ProjectContext): Promise<ScaffoldPlan> {
```
wyattjoh, line 21: `installSkills` isn't gated by `isNpmFramework`, so an iOS or Android run reaches the "Install agent skills?" prompt and then shells out to `npx skills add ...` in a repo that may have no Node toolchain at all. The failure is non-fatal (`runSkillsAdd` warns rather than throws), so this isn't a blocker, but prompting a Swift developer to run npx is odd.

Related: `FRAMEWORK_SKILL_MAP` has no `ios` / `android` entries, so native users get only the default set. If `clerk/skills` ships mobile skills alongside `mobile/clerk-expo`, they're worth adding here.

packages/cli-core/src/commands/init/frameworks/android.ts
```diff
@@ -0,0 +1,33 @@
...
+  name: "Android (Kotlin)",
+  dep: "android",
+
+  matches: (ctx) => ctx.framework.dep === "android",
+
+  async scaffold(ctx: ProjectContext): Promise<ScaffoldPlan> {
+    return {
+      actions: [],
+      postInstructions: [
+        'Add the Clerk Android SDK to app/build.gradle.kts: `implementation("com.clerk:clerk-android-ui:<latest-version>")` and `implementation("androidx.lifecycle:lifecycle-viewmodel-compose:2.9.2")` (requires minSdk 24+ and Java 17+; latest version: https://github.com/clerk/clerk-android/releases)',
```
wyattjoh: Looks like you missed a `<latest-version>`:


```suggestion
        'Add the Clerk Android SDK to app/build.gradle.kts: `implementation("com.clerk:clerk-android-ui:<latest-version>")` and `implementation("androidx.lifecycle:lifecycle-viewmodel-compose:<latest-version>")` (requires minSdk 24+ and Java 17+; latest version: https://github.com/clerk/clerk-android/releases)',
```

wyattjoh (approved):

PR #395, opened 2026-07-24
feat(keyless): make keyless the default and let agents configure a keyless instance

rafa-thayto:
## What

Restores keyless mode as the default for unauthenticated `clerk init` (reversing #268), then extends keyless so an **agent can bootstrap and configure Clerk end to end without a Clerk account** — no login, no `CLERK_PLATFORM_API_KEY`.

**No backend changes required.** Every route used here was verified live against `api.clerk.dev` with an unclaimed keyless application's `sk_test_` key before being written against.

## Init

| Flag / case | Behavior |
| --- | --- |
| unauthenticated + keyless-capable framework | keyless again (human bootstrap + all agent runs); existing-project human runs still log in |
| `--keyless` | forces keyless **even when signed in**, existing projects included |
| `--login` | forces the authenticated flow |
| `--template <b2b-saas\|b2c-saas\|native\|waitlist>` | pre-configures the application at creation |

Usage errors: `--keyless --login`, `--keyless --app`, `--template --login`, and agent-mode `--login` while signed out (agents can't complete browser OAuth).

`-y` means confirmations only — it neither forces nor bypasses keyless.

## Operating a keyless instance

`clerk config pull` / `patch` fall back to the Backend API using the instance secret key. The payload names BAPI resources 1:1 rather than translating the Platform API's document shape:

```sh
clerk config patch --json '{"instance":{"support_email":"dev@acme.com"},"organization_settings":{"enabled":true}}'
```

Groups: `instance`, `communication`, `restrictions`, `organization_settings`, `protect`, `oauth_application_settings`, `instance_settings`. Any other key is a usage error naming the supported ones.

Also now account-free: **`clerk enable/disable orgs`**, **`clerk whoami`** (reports the instance identity), **`clerk env pull`** (writes locally-held keys). Keys are additionally discovered from **`.clerk/.tmp/keyless.json`**, the file Clerk SDKs write when they mint their own application — previously invisible to the CLI.

**Account credentials are deliberately not part of the keyless decision.** Only `--app` or a linked profile selects the account path, so these work with or without a platform key/login. When credentials do exist and the directory simply isn't linked, the command still runs and warns that linking gives the full configuration.

## Known limits (documented, not worked around)

- **Billing** and **auth strategies / social providers** have no secret-key-reachable route — both exit with a clear `auth_required` explanation. RFC drafted separately.
- `restrictions` and `instance_settings` are write-only (no GET route).
- `GET /v1/instance` returns fewer fields than `PATCH` accepts, so the pre-write diff is best-effort; the write is unaffected.
- Backend API has no dry run, so `--dry-run` previews locally and sends nothing.
- Multi-group writes aren't atomic (no transaction); a mid-sequence failure names the groups already applied.

## Test plan

- `bun run format:check`, `bun run lint`, `bun run typecheck` — clean
- `bun run test` — **1947 pass, 0 fail**
- E2E not run locally (1Password vault unavailable); relies on CI
- Verified live against a real keyless app: config pull/patch across groups, `enable orgs` (membership limit change confirmed via Frontend API), `whoami`, `env pull` on an SDK-created app with no `.env`, `--template b2b-saas` (orgs enabled on creation vs. off without it), and every guard rail
- Dev servers booted on all five keyless-capable frameworks; the Clerk `<SignIn />` widget rendered live from a keyless app

## Notes for reviewers

- All account/keyless dispatch is centralized in `commands/config/io.ts` over a typed `InstanceTarget`; no command branches on it directly.
- `init/index.test.ts` crossed 1000 lines, so it was split into `index.test.ts` (plumbing) and `strategy.test.ts` (strategy selection), with the shared spy harness extracted to `src/test/lib/init-harness.ts`.
- Branch is based on an older `main`; the diff is clean against the merge base but a rebase before merge is reasonable.


.changeset/keyless-mode-default.md
wyattjoh, line 1: These changsets are included in the release notes, this is a little verbose for a minor release, maybe we can trim it down a bunch?

packages/cli-core/src/commands/billing/index.ts
```diff
@@ -56,9 +56,26 @@ function describeTargets(targets: Target[]): string {
...
+  const target = await resolveInstanceTarget(options);
+  if (target.kind === "keyless") {
+    throw new CliError(
+      "Billing can only be configured on a claimed application — Clerk's Backend API has no billing settings, so an unclaimed keyless application can't reach them.\n" +
+        "Run `clerk auth login` to claim this application, then re-run the command.",
+      { code: ERROR_CODE.AUTH_REQUIRED },
+    );
+  }
+  return target;
+}
```
wyattjoh, line 72: nit: I'd almost prefer instead to have something like `assertBillingTarget` after calling `resolveInstanceTarget`, as really, the resolving here has nothing to do with billing, only asserting that the instance supports billing.

packages/cli-core/src/commands/config/apply-patch.ts
```diff
@@ -21,45 +28,51 @@ export interface ApplyPatchOptions {
...
 
   if (!hasConfigChanges(current, payload, true)) {
     log.info(dryRun ? "[dry-run] No changes detected" : "No changes detected");
     return false;
   }
 
   const headline = dryRun
-    ? `[dry-run] Proposing PATCH on ${ctx.appLabel} (${ctx.instanceLabel}):`
-    : `${verb} on ${ctx.appLabel} (${ctx.instanceLabel}):`;
+    ? `[dry-run] Proposing PATCH on ${target.label}:`
```
wyattjoh: shouldn't this use `${verb}` here too?

packages/cli-core/src/commands/config/apply-patch.ts
```diff
@@ -21,45 +28,51 @@ export interface ApplyPatchOptions {
...
   if (!dryRun && isHuman() && !yes) {
     const ok = await confirm({ message: "Proceed?" });
     if (!ok) throwUserAbort();
   }
 
   const spinnerMsg = dryRun
-    ? `[dry-run] Validating config on ${ctx.appLabel} (${ctx.instanceLabel})...`
-    : `${verb} on ${ctx.appLabel} (${ctx.instanceLabel})...`;
+    ? `[dry-run] Validating config on ${target.label}...`
+    : `${verb} on ${target.label}...`;
```
wyattjoh: Same as before, seems worthwhile to use the same message and just instead use the prefix

packages/cli-core/src/commands/config/keyless.ts
```diff
@@ -0,0 +1,368 @@
...
+  password: "which authentication strategies are enabled",
+  phone_number: "which authentication strategies are enabled",
+  username: "which authentication strategies are enabled",
+  email_address: "which authentication strategies are enabled",
+  passkey: "which authentication strategies are enabled",
+  social: "social sign-in providers",
+  oauth: "social sign-in providers",
+  second_factors: "multi-factor authentication policy",
+  application_name: "the application's name and branding",
+};
```
wyattjoh, line 113: aside: think we should consider with how much copy we have like this, possibly extracting it to a common framework so we could support i18n later

packages/cli-core/src/commands/config/schema.ts
```diff
@@ -13,6 +14,14 @@ interface ConfigSchemaOptions {
 
 export async function configSchema(options: ConfigSchemaOptions): Promise<void> {
   await withGutter("Fetching configuration schema", async () => {
+    if (await resolveKeylessTarget(options)) {
```
wyattjoh: Other code used the resolve for instance, then checking the keyless type/kind on it, is there a reason we're using this one here instead when the following line is just `resolveAppContext` anyways? Seems like we're double-resolving the target.

packages/cli-core/src/commands/open/keyless-claim.ts
```diff
@@ -0,0 +1,118 @@
...
+ * launcher — which on Windows goes through `cmd /c start "" "<url>"`, where a
+ * `file:`/`javascript:` scheme or an embedded quote stops being just a URL.
+ */
+function isHttpsUrl(value: string): boolean {
+  try {
+    return new URL(value).protocol === "https:";
+  } catch {
+    return false;
+  }
+}
```
wyattjoh: seems like something that could be moved inline as it's only used once

packages/cli-core/src/commands/users/open.ts
```diff
@@ -64,6 +65,31 @@ async function resolveKnownUserDashboardTarget(options: UsersOpenOptions): Promi
...
+  const keyless = await resolveKeylessTarget({ cwd: process.cwd() });
+  if (!keyless) return;
+
+  throw new CliError(
+    `This directory holds an unclaimed keyless application (secret key from ${keyless.source}), which has no Dashboard page — a dashboard link needs an application ID, and one is only assigned when the application is claimed.\n` +
+      `Run \`clerk auth login\` to claim it, then \`clerk users open ${userId}\` will work.\n` +
+      `To inspect the user right now, \`clerk api /users/${userId}\` reads it straight from the instance.`,
+    { code: ERROR_CODE.INSTANCE_NOT_FOUND },
+  );
+}
```
wyattjoh: nit: personally prefer the approach of keeping the happy path on the same level as the starting block indentation. This means for this that the cli error would only be thrown within the `if (keyless) {` block. Improves readability!

packages/cli-core/src/lib/bapi-command.ts
```diff
@@ -58,14 +79,24 @@ export async function resolveBapiSecretKey(options: ResolveBapiSecretKeyOptions)
...
-  if (process.env.CLERK_SECRET_KEY) {
-    validateKeyPrefix(process.env.CLERK_SECRET_KEY, "sk_");
-    return process.env.CLERK_SECRET_KEY;
+  // An unclaimed keyless application keeps its only secret key on disk (env var,
+  // .env.local, or the SDK's own keyless.json) — the same resolution `whoami`,
+  // `config`, and `env pull` already share. The shipped binary is compiled with
+  // --no-compile-autoload-dotenv, so without this every `users`/`api` command
+  // would report "no secret key" on a perfectly live keyless project the moment
+  // it wasn't run via `bun run` (which autoloads .env.local for us in dev).
+  const keyless = await resolveKeylessTarget({ instance: options.instance, cwd: options.cwd });
```
wyattjoh, line 100: Routing this through `resolveKeylessTarget` changes `CLERK_SECRET_KEY` precedence in two ways that aren't in the changeset. `resolveKeylessTarget` returns `undefined` whenever the directory is linked (`if (await resolveProfile(cwd)) return undefined`) and throws a usage error whenever `--instance` is set, so the old step-3 env-var check no longer happens in either case.

Running the real code confirms both:

**Linked project + `CLERK_SECRET_KEY`.** Resolution falls through to `resolveAppContext` and returns the Platform API's key for the linked profile's default instance instead of the env var. That's a silent wrong-target, not a failure: `clerk users delete <id>` in a linked checkout can now hit a different instance than the one the operator named via the environment. Without account credentials it fails outright instead.

**Unlinked project + `CLERK_SECRET_KEY` + `--instance`.** Exits code 2 with ``--instance is not supported for an unclaimed keyless application: the secret key in CLERK_SECRET_KEY env var already targets its own instance.`` Previously the env key was returned and `--instance` was a no-op. So `clerk api --instance dev /users` and `clerk users list --instance prod` now break, and the message talks about a keyless application to someone who deliberately set a key for a claimed one.

The test that would have caught this is gone. `bapi-command.test.ts:201` asserts `resolveBapiSecretKey({ instance: "dev" })` resolves to `sk_env_123`, but it only passes because `resolveKeylessTargetSpy.mockResolvedValue({...})` stubs out the very guard that throws — the old test exercising the real env-var path was deleted.

Either restore env-var precedence ahead of the profile lookup, or, if the new precedence is intended, stop forwarding `instance` into `resolveKeylessTarget` here and add unmocked tests for both scenarios. A mocked `resolveKeylessTarget` can't validate this function's contract.

packages/cli-core/src/commands/doctor/context.ts
```diff
@@ -44,6 +52,49 @@ export function createDoctorContext(): DoctorContext {
       return appPromise;
     },
 
+    getKeylessTarget() {
```
wyattjoh, line 56: `resolveKeylessTarget` throws a `CliError` when the key it finds doesn't start with `sk_`, and every keyless-aware check calls `ctx.getKeylessTarget()` without catching it. The per-check handler in `doctor/index.ts:46` turns each throw into `{ name: "Unknown check", message: "Check crashed: …" }`, so the check names are lost.

Reproduced with a `.env.local` containing `CLERK_SECRET_KEY=oops_not_a_secret_key`:

```
✗ Check crashed: Expected a secret key starting with `sk_` in .env.local, found something else.
✗ Check crashed: Expected a secret key starting with `sk_` in .env.local, found something else.
✗ Check crashed: Expected a secret key starting with `sk_` in .env.local, found something else.
✗ Check crashed: Expected a secret key starting with `sk_` in .env.local, found something else.
✗ Check crashed: Expected a secret key starting with `sk_` in .env.local, found something else.
```

Five checks — Logged in, Authentication valid, Project linked, Application reachable, Instance IDs — all report as "Unknown check", and the user learns nothing about which parts of their setup are actually fine. A malformed key is precisely the misconfiguration doctor exists to diagnose, so this should be one named failing check rather than five anonymous crashes.

Catching the `INVALID_KEY_FORMAT` error inside `getKeylessTarget()` and caching it as a diagnosable state (rather than letting it propagate into every caller) would let `checkLoggedIn` report it once, by name, with a remedy.

packages/cli-core/src/commands/whoami/index.ts
```diff
@@ -1,21 +1,75 @@
...
+  // question in a way that can't throw, so an expired session falls through to
+  // the keyless path instead of aborting it.
+  const token = await getValidToken().catch((error: unknown) => {
+    log.debug(`credentials: stored session unusable (${errorMessage(error)})`);
+    return null;
+  });
+
+  // No usable account, but the directory may still hold a working keyless
+  // application. Report what it is instead of a flat "not logged in".
   if (!token) {
```
wyattjoh, line 75: `resolveIdentity` only consults `resolveKeylessTarget` when there's no usable token, so a valid account session short-circuits the keyless branch entirely. But `resolveBapiSecretKey` prefers the local keyless key whenever the directory is unlinked, regardless of account credentials.

In an unlinked directory holding both a valid session and a `CLERK_SECRET_KEY`, that means `clerk whoami` reports the account while `clerk api /users` and `clerk users list` operate on the instance the local secret key addresses. I confirmed both sides against the real resolvers.

That's a bad split for the one command people run to answer "what will my next command touch?" — `config patch` at least warns via `warnKeylessCoversLess`, but `whoami` states the account with no hint that the Backend API surface is pointed somewhere else.

Worth either having `whoami` mention the local key when one is present and the directory is unlinked, or aligning the two resolvers on a single precedence rule.

---

**Separately, on `render()` further down this file:** the keyless branch ends with `printNextSteps(NEXT_STEPS.WHOAMI)`, and `lib/next-steps.ts:63` defines that as ``["Run `clerk link` to connect this directory to an application"]``.

So the last line a user sees after `clerk whoami` in a keyless project tells them to run `clerk link`, which can't succeed — there is no application in any account to link to. That's the exact dead end the rest of this PR is built around avoiding: `open/README.md` says the errors "deliberately do not send the user solely to `clerk link`", `users/open.ts:79` refuses to give that advice for the same reason, and `doctor/checks.ts:89` routes through `claimHint()` to say `clerk auth login` or "Claim it from the Clerk Dashboard" instead.

The test at `whoami/index.test.ts:357` only asserts on `"unclaimed keyless application"` and `".env.local"`, so it passes either way. A `WHOAMI_KEYLESS` entry pointing at `clerk auth login`, plus an `expect(captured.err).not.toContain("clerk link")` assertion mirroring the one in `open/index.test.ts`, would close it.

.changeset/keyless-mode-default.md
```diff
@@ -0,0 +1,38 @@
+---
+"clerk": minor
```
wyattjoh: The changeset covers this area with: *"`clerk users`, `clerk api`, and the rest of the Backend API command surface now resolve a keyless project's own secret key from its env files."* That describes the addition but not the subtraction — `CLERK_SECRET_KEY` no longer wins over a linked profile, and combining it with `--instance` is now a hard usage error (see the `bapi-command.ts` comment; both verified by running the code).

Both break invocations that work on `main`, so someone reading the release notes to decide whether to upgrade gets no warning. `.claude/skills/changesets/references/policy.md` maps incompatible CLI behavior changes to `major` and gates that behind explicit sign-off.

Either revert the precedence change — which makes `minor` correct and is my preference — or disclose it in the changeset body and take the bump type to a maintainer.

packages/cli-core/src/lib/credential-store.test.ts
```diff
@@ -154,6 +154,21 @@ describe("credential-store", () => {
     expect(await getStoredSession()).toBeNull();
   });
 
+  test("getValidToken treats a 429 refresh failure as transient, not an expired session", async () => {
```
wyattjoh, line 157: `isUnrecoverableRefreshFailure()` makes a non-429 4xx from the token endpoint throw `sessionExpiredError()` while deliberately *leaving the stored credentials in place* — unlike the `invalid_grant` branch above it, which calls `deleteToken()`. That difference is the whole point of the new branch, and it's the mechanism behind the changeset's claim that commands no longer leak `Token refresh failed (401)`.

The only test added is the negative one (429 treated as transient). Nothing asserts the positive case: that `ApiError(401, '{"error":"invalid_client"}')` produces an `AuthError` *and* that `getStoredSession()` still returns the session afterward. A future edit that moved `deleteToken()` into the new branch, or widened the 429 exclusion to all 4xx, would keep the suite green.

`whoami/index.test.ts:330` looks like coverage but isn't — it stubs `mockGetValidToken.mockRejectedValue(new CliError("Token refresh failed (401)"))`, bypassing the classifier.

A sibling to the 429 test asserting `rejects.toBeInstanceOf(AuthError)` plus `expect(await getStoredSession()).toEqual(session)` would pin it.

packages/cli-core/src/commands/whoami/index.test.ts
```diff
@@ -8,13 +8,29 @@ import {
...
+const mockHasKeyPairMismatch = mock();
+const mockBapiRequest = mock();
 
 mock.module("../../lib/credential-store.ts", () => ({
   ...credentialStoreStubs,
   getValidToken: (...args: unknown[]) => mockGetValidToken(...args),
+  hasStoredCredentials: (...args: unknown[]) => mockHasStoredCredentials(...args),
+}));
+
+mock.module("../../lib/keyless-target.ts", () => ({
```
wyattjoh, line 28: This PR adds a comment to `test/lib/stubs.ts:129` — "Mocking a module replaces it wholesale, so this must cover **every** export of lib/config.ts — a missing name is an import error in any consumer, not just the one under test" — and fills in seven missing `configStubs` entries accordingly.

The same lesson isn't applied to the new module. `lib/keyless-target.ts` exports five runtime values; the mocks supply three here, one in `open/index.test.ts`, one in `users/open.test.ts`, and one in `doctor/context.test.ts`. `resolveInstanceTarget` — imported by `config/push.ts`, `config/pull.ts`, `orgs/index.ts`, and `billing/index.ts` — is missing from all four.

Nothing fails today because `bun run test` passes `--parallel`, but per `.claude/rules/testing.md` `mock.module` registrations are process-lifetime, which is exactly the hazard the stubs comment describes.

A `keylessTargetStubs` export in `test/lib/stubs.ts` covering all five, spread into each file the way `configStubs` now is, would keep this from biting later.

packages/cli-core/src/commands/orgs/index.ts
```diff
@@ -69,17 +69,26 @@ export async function orgsEnable(options: OrgsOptions): Promise<void> {
...
-      withApiContext(
-        fetchInstanceConfig(ctx.appId, ctx.instanceId, ["billing", "organization_settings"]),
-        "Failed to fetch config",
-      ),
-    );
+    // Billing lives only in the account-level config document, so a keyless run
+    // can't read it. Skip the pre-flight fetch entirely rather than half-run it:
+    // applyConfigPatch reads the group it needs on its own.
+    const current =
+      target.kind === "account"
```
wyattjoh, line 79: On the keyless path `current` is `undefined`, so `orgBillingOn` is always `false` and the guard below (`if (orgBillingOn && isAgent()) throwUsageError(...)`) can never fire. That's a deliberate safety guard being disabled, documented in `orgs/README.md` as "can't run its organization-billing pre-flight check".

`config/keyless.test.ts:695` covers `enable orgs` end to end in a keyless project, but there's no matching `disable orgs` case — so neither the `current === undefined` path nor the bypassed refusal is pinned. A later refactor reintroducing an unconditional `current.billing` access would `TypeError` on the keyless path with the suite still green.

Adding a `disable orgs` case alongside the existing `enable orgs` one, asserting the PATCH lands and that no Platform API config fetch is attempted, would cover it.

packages/cli-core/src/commands/config/README.md
```diff
@@ -149,9 +157,108 @@ clerk config put --file full-config.json --dry-run
...
+| `instance_settings`          | `/v1/beta_features/instance_settings`     | No       | `test_mode`, `progressive_sign_up`, `from_email_address`, `restricted_to_allowlist` |
+
+Any other top-level key exits with a usage error naming the supported ones. Most of them (`session`, `sign_up`, `auth_email`, …) are genuinely account-only and the error points at `clerk auth login`. A handful — `enterprise_connections`, `saml_connections`, `oauth_applications`, `domains` — are BAPI resource _collections_ reachable on an unclaimed application today; they're just not part of this config document, so the error points at `clerk api /<resource>` instead of a login that wouldn't add them anyway.
+
+`instance_settings` is backed by a beta route, and is the only way to reach those four auth-config fields without an account — which is why it's included.
+
+`clerk config pull` returns the same envelope. `restrictions` is omitted because the Backend API has no read route for it; asking for it by name prints a warning rather than failing.
```
wyattjoh: `KEYLESS_GROUPS` in `config/keyless.ts` marks two groups `readable: false` — `restrictions` and `instance_settings` — and `pullKeylessConfig` filters on `isReadable` for both. The changeset gets this right ("`restrictions` and `instance_settings` are omitted"), and this README's own payload-shape table marks both as Readable: No. Only this sentence is out of step.

It matters because someone who patches `instance_settings.test_mode` and then pulls to confirm will find it absent and reasonably conclude the write failed. Worth naming both groups here.

packages/cli-core/src/commands/doctor/README.md
```diff
@@ -37,6 +37,30 @@ clerk doctor --fix       # Offer to auto-fix issues
...
+### Keyless applications
+
+The Authentication token, Token validity, and Project linkage checks resolve
+the same keyless fallback the rest of the CLI uses (`lib/keyless-target.ts`):
+a project with no account session and no linked profile, but a `sk_...` key
+on disk (or in `CLERK_SECRET_KEY`/framework env var), is running on an
+**unclaimed keyless application** — a legitimate, healthy state, not a broken
+one.
+
+- No token, keyless key present → **pass**, naming the instance and noting
```
wyattjoh: The new keyless section says: "No token, keyless key present → **pass**, naming the instance and noting `clerk auth login` claims it." `claimHint()` only says that when `hasClaimBreadcrumb()` is true. Otherwise — an SDK-minted `.clerk/.tmp/keyless.json`, or a hand-copied `CLERK_SECRET_KEY`, both of which this PR explicitly adds support for — the message is the opposite: ``"Claim it from the Clerk Dashboard — `clerk auth login` only claims applications `clerk init` created."``

That distinction is the entire reason `claimHint` exists, so it's worth stating both outcomes here and what selects between them.

packages/cli-core/src/commands/init/index.ts
```diff
@@ -330,10 +455,52 @@ async function authenticateAndLink(
...
+ * `--fresh` is the explicit "I know, replace it anyway" escape hatch.
+ */
+async function shouldKeepExistingKeyless(
+  cwd: string,
+  skipConfirm: boolean,
+  fresh: boolean,
+): Promise<boolean> {
+  if (fresh) return false;
+
+  const existing = await readKeylessBreadcrumb(cwd);
```
wyattjoh, line 487: This guard only looks at `.clerk/keyless.json` — the breadcrumb `clerk init` writes itself. But this same PR teaches every other keyless path to also recognise an app the SDK minted for itself in `.clerk/.tmp/keyless.json` (`findLocalSecretKey`, `findKeylessClaimUrl`, `resolveKeylessTarget`). `shouldKeepExistingKeyless` was not brought along, so the two disagree about whether an application exists.

Verified in a directory holding only an SDK-minted app:

```
init guard sees (readKeylessBreadcrumb):    undefined
but an app IS here (findLocalSecretKey):    { secretKey: "sk_test_sdkapp",
                                              source: ".clerk/.tmp/keyless.json" }
```

So in a project where the keyless app came from running the dev server rather than from `clerk init`, `shouldKeepExistingKeyless` returns `false` and `setupKeylessApp` mints a replacement and overwrites the env keys — orphaning the SDK's app along with anything configured on it or any users created in it. That is exactly the harm this function's own docstring describes: "Re-running init must not silently mint a replacement and orphan that app."

It also bypasses the consent design deliberately built here: because no breadcrumb is found, the `skipConfirm` branch is never reached, so neither agent mode nor `-y` gets the "keep by default" protection, and the human confirm prompt never appears either. The replacement is silent in every mode.

Reading the SDK file too — the app is identifiable via `findLocalSecretKey`, which already knows how — would close it. The prompt has no `createdAt` to show for that source, so it would need slightly different wording.

wyattjoh (approved):

PR #404, opened 2026-07-31
fix: dual-era MCP server detection in doctor probe + 2026-07-28 header contract in mcp run

rafa-thayto:
Fixes AIE-1355

## What

Implements MCP 2026-07-28 support in the CLI's hand-rolled MCP bridge (SDK remains a type-only import).

### `clerk doctor` probe (`probe.ts`)

Implements the spec's backward-compatibility algorithm:

- Probes with a modern `server/discover` first (per-request `_meta` + `MCP-Protocol-Version`/`Mcp-Method` headers).
- On a 400, inspects the response body: a recognized modern JSON-RPC error (`-32020` HeaderMismatch, `-32021` MissingRequiredClientCapability, `-32022` UnsupportedProtocolVersion) proves a modern server and never falls back — `-32021` (capability-gated) counts as reachable, and `-32022` is retried once with an advertised supported version. Reserved MCP codes on non-400 statuses (proxy error pages) are not trusted and still fall back.
- Anything else falls back to the legacy `initialize` handshake, so legacy-only servers keep probing correctly.
- `401`/`403` at either stage still reads as "reachable, authentication required".

### `clerk mcp run` relay (`run.ts`)

Every relayed POST now carries the 2026-07-28 request-metadata headers, derived from the body so strict servers never reject with `HeaderMismatch` (-32020):

- `Mcp-Method` from the body `method`.
- `Mcp-Name` from `params.name`/`params.uri` on `tools/call`/`resources/read`/`prompts/get`, with Base64 sentinel encoding (`=?base64?…?=`) for non-header-safe values.
- `MCP-Protocol-Version` from the request's own `_meta` when a modern client declares it, else the legacy-negotiated version.
- `Mcp-Param-*` mirroring for `x-mcp-header` tool parameters, learned from `tools/list` responses; tools with spec-invalid annotations are excluded from forwarded `tools/list` results (with a stderr warning), as the spec requires.
- Session tracking no longer treats "no `Mcp-Session-Id` yet" as "connection never worked" — 2026-07-28 servers are stateless and never mint one, so a scoped 401 mid-run no longer kills the bridge.

Shared header/encoding/annotation logic lives in a new `headers.ts`.

## How verified

- TDD: probe decision table (one test per era/fallback/auth case), header-contract and mirroring tests for the relay, 51 unit tests for the new header helpers.
- `bun run format`, `bun run lint`, `bun run typecheck`, `bun run test` (2525 pass) all green locally.
- `bun run test:e2e:op` could not run locally (no access to the "AI Enablement" 1Password vault from this machine); e2e does not touch the MCP code paths and runs in CI on this PR.

## Docs

- `commands/mcp/README.md` — bridge header contract + dual-era reachability note.
- `commands/doctor/README.md` — MCP check description.
- Changeset: `patch`.


rafa-thayto:
Closing: decided the CLI stays modern-only (2026-07-28). Dual-era client detection is optional per the spec (versioning: dual-era support is a MAY), and we don't need the doctor probe fallback. The uncommitted -32020/21/22 error-surfacing fix from the conformance run will move to a standalone PR. Ref: AIE-1355.

PR #410, opened 2026-08-06
feat(telemetry): CLI usage telemetry

djgould:
Adds usage telemetry to the CLI ([GROW-1200](https://linear.app/clerk/issue/GROW-1200/implement-cli-anonymous-telemetry)): one `CLI_COMMAND_EXECUTED` event per command run, sent to `clerk-telemetry.com` (BigQuery, `sdk = 'clerk-cli'`).

## What an event contains

| Field | Example |
| --- | --- |
| `command` | `users list` — subcommand names only, never arguments |
| `flags` | `json,limit` — flag names only, never values |
| `outcome` / `exit_code` / `error_code` | `error` / `1` / `not_linked` |
| `duration_ms` | `412` |
| `machine_uuid` | random per-machine UUID |
| `install_method` | `npm_global`, `homebrew`, `npx`, … |
| `ai_agent` | `claude_code`, `cursor`, `gemini_cli`, … — empty for humans |
| `terminal_program` | `ghostty`, `warp`, … |
| `mode` | `human` / `agent` (TTY detection) |
| `os` / `arch` / `ci` / `in_tmux` / `in_screen` | `darwin` / `arm64` / `false` / … |
| `env` | `production` — which Clerk API the CLI targets (`clerk switch-env`); non-production means Clerk-internal usage, so analysts can filter it |
| `workspace_id` / `app_id` | linked project ids; `null` when unlinked |

Never collected: command arguments, option values, file paths, error messages, or personal data.

## Controls

- `clerk telemetry status|disable|enable` — persisted opt-out; a `disable` run sends nothing itself
- `CLERK_TELEMETRY_DISABLED` / `DO_NOT_TRACK` env vars (any non-false value)
- The first run only shows a disclosure notice and sends nothing
- `--verbose` prints each event before the POST; failures never affect the command (1s cap)
- The User-Agent gains an `AIAgent/<agent>` segment so API logs can classify agent traffic — omitted when telemetry is opted out

**Release gate:** telemetry-service drops iid-less events (`workers/telemetry-service/src/index.ts:91`) — needs the one-line `sdk === "clerk-cli"` exemption in cloudflare-workers before this ships, then an end-to-end BigQuery check.


packages/cli-core/src/cli-program.ts
```diff
@@ -231,7 +239,15 @@ export async function runProgram(
...
     await program.parseAsync(argv, { from });
+    // Some commands report failure via process.exitCode instead of throwing —
+    // read it back so telemetry doesn't record them as successes.
+    const softExitCode = Number(process.exitCode ?? EXIT_CODE.SUCCESS);
+    await finalizeAndSendTelemetry({
+      outcome: softExitCode === EXIT_CODE.SUCCESS ? "success" : "error",
+      exitCode: softExitCode,
+    });
   } catch (error) {
+    await finalizeAndSendTelemetry(telemetryResultForError(error));
```
wyattjoh: `await finalizeAndSendTelemetry(...)` is the first statement in the catch block, before any error rendering. The user waits on a network POST before learning their command failed.

Measured with `CLERK_TELEMETRY_URL` pointed at a blackholed address, on a purely local ~0.1s failure:

```
$ bun packages/cli-core/src/cli.ts env pull
[+1.08s] {"error":{"code":"not_linked",...}}
```

A full second before the "not linked" message appears. Rendering the error first and finalizing afterwards (or from a `finally` after the output block) fixes it without changing what gets collected.

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+        workspace_id: resolved?.profile.workspaceId ?? null,
+        app_id: resolved?.profile.appId ?? null,
+      },
+    };
+
+    log.debug(`telemetry: event ${JSON.stringify(event)}`);
+
+    const url = process.env.CLERK_TELEMETRY_URL ?? DEFAULT_TELEMETRY_ENDPOINT;
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), TELEMETRY_TIMEOUT_MS);
```
wyattjoh: Two issues here.

First, the cost is real on the success path too. Measured on `clerk telemetry status` with a hanging endpoint: **0.105s → 1.120s**. That is a uniform latency tax on every invocation.

Second, the `AbortController` only covers the `loggedFetch` below it. `getTelemetryStatus()`, `ensureMachineUuid()` (config read plus a first-run write), and `resolveProfile()` (config read plus up to two `git` subprocess spawns for commands that never otherwise touch git) all run before this timer starts and are unbounded. So "failures never affect the command (1s cap)" is stronger than what the code guarantees.

Either detach the send from the exit path, or lower `TELEMETRY_TIMEOUT_MS` and put the entire body inside the abort budget rather than just the request.

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+    try {
+      await loggedFetch(url, {
```
wyattjoh: `loggedFetch` wraps the request in `withNetworkAccess`, whose `observeHostCapabilityFailure` calls `warnAboutSandbox()` on a permission-like failure in agent mode. Since telemetry is now the last network call on every command, a sandboxed run whose other work was purely local can surface:

> Host-only Clerk state or capabilities may be unavailable in agent mode (possible sandboxed run)…

after an otherwise successful command. The `try/catch` here swallows the error, but the `log.warn` has already been emitted — which cuts against this module's own "Telemetry must never affect the command" docstring, and misattributes a telemetry failure to the user's actual command.

Since `.claude/rules/debug-logging.md` rules out calling `fetch` directly, an opt-out flag on `loggedFetch` that skips capability observation seems like the cleanest fix.

packages/cli-core/src/lib/env-signals.ts
```diff
@@ -0,0 +1,73 @@
...
+  return env.TERM_PROGRAM ?? "";
+}
+
+/**
+ * How the CLI binary was installed/invoked. The npm wrapper and package
+ * runners leave `npm_*` vars in the child env; direct binary installs are
+ * classified by executable path.
+ */
+export function detectInstallMethod(env: EnvLike, execPath: string): string {
+  if (env.CLERK_INSTALL_METHOD) return env.CLERK_INSTALL_METHOD;
```
wyattjoh: `CLERK_INSTALL_METHOD` here, plus `LC_TERMINAL` (line 35) and `TERM_PROGRAM` (line 42), are returned unmodified and land in the payload. Every other signal in this module is a closed enum, which is what makes these stand out — they are unbounded-cardinality analytics dimensions and an arbitrary-string path into the pipeline.

It also puts pressure on the README's "never collects command arguments, option values, file paths" claim: `CLERK_INSTALL_METHOD` can hold anything, including a path.

Truncating to a small fixed length and either allowlisting known values or slugging them (lowercase, non-alphanumerics collapsed) would keep the signal without the open end.

packages/cli-core/src/commands/telemetry/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+}
+
+export async function telemetryStatus(): Promise<void> {
+  const status = await getTelemetryStatus();
+  if (isAgent()) {
+    log.data(JSON.stringify(status));
+    return;
+  }
+  log.data(`Telemetry is ${status.enabled ? "enabled" : "disabled"}`);
```
wyattjoh: `log.data("Telemetry is enabled")` writes UI prose to stdout. Per `.claude/rules/logging.md`, `log.data()` is reserved for pipeable machine-readable output and human status text belongs on stderr via `log.info`. The agent-mode JSON branch just above is correct as-is.

Switching the human branch to `log.info` needs a matching update to the integration test that asserts `enabled.stdout`.

wyattjoh (commented):
Solid implementation overall, and the mid-PR docs revision to drop "anonymous" once workspace/app IDs entered the payload was the right call. My concerns are concentrated in one place: the send is awaited on the critical path. I ran the CLI to check rather than just reading it, and the measurements are in the inline comments. Worth noting `cli-program.ts` is this repo's highest-churn file and its top bug hotspot, so the `runProgram` control-flow change gets extra scrutiny.

packages/cli-core/src/lib/config.ts
```diff
@@ -207,6 +214,40 @@ export async function setRelayEntry(key: string, entry: RelayEntry): Promise<voi
...
+  return true;
+}
+
+/** Persisted telemetry opt-out, set via `clerk telemetry disable`. */
+export async function getTelemetryDisabled(): Promise<boolean> {
+  const config = await readConfig();
+  return config.telemetryDisabled === true;
+}
+
+export async function setTelemetryDisabled(disabled: boolean): Promise<void> {
```
wyattjoh, line 247: `setTelemetryDisabled(true)` clears nothing else, so opting out and later back in resumes the same `machineUuid`. Clearing it on disable would make the opt-out a cleaner privacy boundary — the user gets a fresh identifier if they return, rather than being rejoined to their prior history.

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+  return normalized !== "0" && normalized !== "false";
+};
+
+type OptOutEnvVar = "CLERK_TELEMETRY_DISABLED" | "DO_NOT_TRACK";
+
+function optOutEnvVar(env: EnvLike): OptOutEnvVar | null {
+  if (isOptOutEnv(env.CLERK_TELEMETRY_DISABLED)) return "CLERK_TELEMETRY_DISABLED";
+  if (isOptOutEnv(env.DO_NOT_TRACK)) return "DO_NOT_TRACK";
+  return null;
+}
```
wyattjoh: nit: can be represented simpler as:

```suggestion
const OPT_OUT_ENV_VARS = [
  "CLERK_TELEMETRY_DISABLED",
  "DO_NOT_TRACK",
] as const;

type OptOutEnvVar = (typeof OPT_OUT_ENV_VARS)[number];

function optOutEnvVar(env: EnvLike): OptOutEnvVar | null {
  for (const envVar of OPT_OUT_ENV_VARS) {
    if (isOptOutEnv(env[envVar])) {
      return envVar;
    }
  }

  return null;
}
```

packages/cli-core/src/lib/user-agent.ts
```diff
@@ -4,17 +4,21 @@
...
 
-export function buildUserAgent(): string {
+export function buildUserAgent(env: EnvLike = process.env): string {
   const version = resolveCliVersion() ?? DEV_CLI_VERSION;
   const segments = [`Bun/${Bun.version}`, `${process.platform}-${process.arch}`];
-  if (process.env.CI) segments.push("ci");
-  return `Clerk-CLI/${version} (${segments.join("; ")})`;
+  if (env.CI) segments.push("ci");
+  const agent = detectAiAgent(env);
+  const agentToken = agent ? ` AIAgent/${agent}` : "";
```
wyattjoh: Why not just add this to segments instead? Also, shouldn't this respect the telemetry opt-out?

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+  const current = context;
+  context = null;
+  if (!current) return;
+
+  try {
+    // Re-checked here (not just at start) so `clerk telemetry disable` itself
+    // sees the freshly persisted opt-out and sends nothing.
+    if (!(await getTelemetryStatus()).enabled) return;
+
+    if (await maybeShowTelemetryNotice()) return;
```
wyattjoh: Maybe best to add a comment here specifying that we're skipping sending telemetry _because_ the notice explicitly states that nothing was sent yet.

packages/cli-core/src/cli-program.ts
```diff
@@ -135,6 +142,7 @@ export function createProgram(): Program {
     if (activeEnv !== "production") {
       process.stderr.write(`[${activeEnv.toUpperCase()}]\n`);
     }
+    startCommandTelemetry(actionCommand);
```
wyattjoh: Worthwhile to verify if the `actionCommand` won't ever equal anything that the user passes in as their PII input, such as a mistyped command which might send their token/email/etc here.

packages/cli-core/src/cli-program.ts
djgould: Fixed in 3a4da8d — the send starts at catch-top and overlaps rendering; awaited only right before each exit. Your repro now shows the error at local speed.

packages/cli-core/src/lib/telemetry.ts
djgould: Fixed in d75fa06 — one deadline at finalize entry covers config I/O, resolveProfile, and the POST, and holds even if the work ignores the abort signal. The ~100ms healthy-endpoint tail on success stays by design: outcome/duration require send-at-end.

packages/cli-core/src/lib/telemetry.ts
djgould: Fixed in 720e16a — the telemetry POST is best-effort; permission-shaped failures no longer surface the host-capability hint.

packages/cli-core/src/lib/env-signals.ts
djgould: Fixed in d9e8033 — terminal fallbacks slugged + 32-char cap, CLERK_INSTALL_METHOD allowlisted.

packages/cli-core/src/commands/telemetry/index.ts
djgould: Fixed in 76167b1 — bare enabled/disabled on stdout (whoami pattern), reason on stderr.

packages/cli-core/src/cli-program.ts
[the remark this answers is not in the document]
djgould: Fixed in d309c9d — context starts at the top of the hook; invalid --mode now emits an error event (integration-tested).

packages/cli-core/src/lib/config.ts
```diff
@@ -207,6 +214,40 @@ export async function setRelayEntry(key: string, entry: RelayEntry): Promise<voi
...
+  const config = await readConfig();
+  if (disabled) {
+    config.telemetryDisabled = true;
+  } else {
+    delete config.telemetryDisabled;
+  }
+  await writeConfig(config);
+}
```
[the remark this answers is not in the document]
djgould: Declining in this PR: the unlocked read-modify-write is this file's established pattern for every field (auth, profiles, relay), so locking belongs to the config layer as its own change — follow-up queued for atomic config writes.

packages/cli-core/src/lib/env-signals.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export function detectTerminalProgram(env: EnvLike): string {
+  if (env.LC_TERMINAL) return env.LC_TERMINAL;
+  if (env.WARP_CLIENT_VERSION) return "warp";
+  if (env.WT_SESSION) return "windows_terminal";
+  if (env.KITTY_WINDOW_ID) return "kitty";
+  if (env.ALACRITTY_WINDOW_ID || env.ALACRITTY_LOG) return "alacritty";
+  if (env.WEZTERM_EXECUTABLE || env.WEZTERM_PANE) return "wezterm";
+  if (env.GHOSTTY_RESOURCES_DIR) return "ghostty";
+  return env.TERM_PROGRAM ?? "";
```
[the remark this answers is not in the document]
djgould: Fixed in d9e8033.

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    const machineUuid = await ensureMachineUuid();
+    const resolved = await resolveProfile(process.cwd()).catch(() => undefined);
+    const version = resolveCliVersion() ?? DEV_CLI_VERSION;
```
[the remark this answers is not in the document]
djgould: Fixed in d75fa06.

README.md
```diff
@@ -56,3 +57,16 @@ Commands:
...
 ```
+
+## Telemetry
+
+The Clerk CLI collects usage telemetry: command name, flag names, duration, outcome,
+environment signals (OS, install method, terminal), a random machine identifier — and
+your workspace and app IDs when a project is linked. It never collects command
+arguments, option values, file paths, or personal data. The first run only shows a
+notice and sends nothing, and `clerk --verbose` prints every event before it is sent.
+See https://clerk.com/docs/telemetry for details.
```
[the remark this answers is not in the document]
djgould: Resolved behaviorally in 7614f4b — the grace is universal outside CI; README/changeset carry the CI qualifier.

packages/cli-core/src/lib/user-agent.ts
djgould: Both taken in 3124921: the marker moved into the segment list — `Clerk-CLI/<v> (Bun/…; darwin-arm64; AIAgent/claude_code)` — and it now honors the full opt-out set (`CLERK_TELEMETRY_DISABLED` / `DO_NOT_TRACK` / `clerk telemetry disable`): the UA resolves once per process against the persisted flag and fails toward omission. Kept the `AIAgent/<name>` spelling inside the segment for greppability. (Also rebased your optOutEnvVar array refactor into env-signals.ts, where the helper now lives.)

packages/cli-core/src/lib/telemetry.ts
djgould: Restored in the latest push — the call site now states the coupling: the return exists because the notice copy promises "Nothing has been sent during this run", so the control flow and the printed text have to move together.

packages/cli-core/src/cli-program.ts
djgould: Verified structurally and now regression-tested (latest push): payload.command walks the resolved Command objects' registered .name() literals, never argv — a typo or pasted secret either becomes an argument (never serialized; the existing zsh leak-guard covers that) or exits inside Commander before any hook runs, so no event exists. New test.each cases run `clerk users sk_test_secret123` and top-level `clerk sk_live_pasted` and assert zero events and the secret absent from every captured request body.

packages/cli-core/src/lib/telemetry.ts
djgould: Applied — your 2a450e4 landed it, and it now lives in env-signals.ts after the helper moved there for the User-Agent opt-out work (3124921).

packages/cli-core/src/lib/config.ts
djgould: Fixed in 08f747c — disable sheds the uuid; re-enable starts a fresh, unlinkable identity.

wyattjoh (approved):

packages/cli-core/src/lib/fetch.ts
```diff
@@ -11,10 +11,28 @@
...
+/** Test-only: recompute the User-Agent on the next request. */
+export function _resetUserAgentCache(): void {
+  userAgentPromise = undefined;
+}
+
+function resolveUserAgent(): Promise<string> {
+  // The AIAgent segment exists purely for analytics classification, so it
+  // honors the telemetry opt-outs (env vars and `clerk telemetry disable`).
+  userAgentPromise ??= (async () => {
+    const optedOut =
```
rafa-thayto: The AIAgent segment honors both opt-outs but not the first-run grace. `loggedFetch` backs every API call, so a brand new user's very first request already carries `AIAgent/claude_code` before the disclosure notice has ever been printed, while the changeset promises the first run sends nothing. Gating on the notice state, with CI exempt to match `maybeShowTelemetryNotice`, keeps the header consistent with that promise:

```ts
// config.ts: read-only peek, no mutation
export async function getTelemetryNoticeShown(): Promise<boolean> {
  return (await readConfig()).telemetryNoticeShown === true;
}

// fetch.ts
const graceRun =
  !process.env.CI && !(await getTelemetryNoticeShown().catch(() => false));
const optedOut =
  optOutEnvVar(process.env) !== null ||
  graceRun ||
  (await getTelemetryDisabled().catch(() => true));
```

packages/cli-core/src/commands/telemetry/index.ts
```diff
@@ -0,0 +1,74 @@
...
+  log.success("Telemetry disabled. Nothing will be sent from this machine.");
+}
+
+export async function telemetryEnable(): Promise<void> {
+  await setTelemetryDisabled(false);
+  log.success("Telemetry enabled.");
+  const status = await getTelemetryStatus();
+  if (!status.enabled && status.reason === "env") {
+    log.warn(`\`${status.envVar}\` is still set — telemetry stays disabled until it is unset.`);
+  }
```
rafa-thayto: Only the env reason gets a warning here. On a dev build, `clerk telemetry enable` prints an unqualified "Telemetry enabled." while `clerk telemetry status` immediately reports disabled for the dev-build guard, the same gotcha the env branch protects against. `reason === "config"` is impossible right after `setTelemetryDisabled(false)`, so covering dev-build closes the set.

```suggestion
  if (status.enabled) return;
  if (status.reason === "env") {
    log.warn(`\`${status.envVar}\` is still set, telemetry stays disabled until it is unset.`);
    return;
  }
  if (status.reason === "dev-build") {
    log.warn("This is a dev build (0.0.0-dev), telemetry stays disabled regardless.");
  }
```

rafa-thayto (changes requested):

rafa-thayto (approved):

packages/cli-core/src/lib/fetch.ts
djgould: Right — the segment claimed the events' principle but skipped their disclosure ordering. Fixed in 73d4649 with your shape: read-only getTelemetryNoticeShown() peek, and the UA resolution treats notice-not-yet-shown (outside CI) as a grace condition, failing toward omission. Covered by fetch-layer tests: fresh config → bare UA, post-notice → segment, CI fresh config → segment.

packages/cli-core/src/commands/telemetry/index.ts
djgould: Fixed in 3f1f342 — took your suggestion nearly verbatim (early-return + dev-build warn), with an integration test that `telemetry enable` on a dev build warns instead of over-promising. Agreed the config reason is unreachable right after setTelemetryDisabled(false).

rafa-thayto (approved):

rafa-thayto (approved):

PR #412, opened 2026-08-10
fix(auth): print the sign-in URL always and launch the host browser from WSL

### LINKED ISSUE #411: WSL2: `clerk auth login` needs device flow or printed auth URL fallback
When running `clerk auth login` inside a WSL2 environment, the CLI cannot reliably open a browser, so interactive login may fail or become unusable.

### Problem
`clerk auth login` appears to depend on opening a local browser automatically. In WSL2 (and similar headless/remote terminal environments), that assumption often breaks, which makes login infeasible.

### Expected behavior
Please support one (or both) of these fallback paths when browser launch is unavailable:

1. **Device authorization flow** (recommended)  
   - Show a verification URL + short code in terminal  
   - Let user complete auth in any browser/device  
   - CLI polls and completes login once authorized

2. **Manual browser flow fallback**  
   - Print the exact auth URL to stdout (and optionally “press Enter after auth”)  
   - Keep process waiting and detect completion/token handoff after browser auth

### Why this matters
This improves usability for:
- WSL2 users
- SSH/remote development environments
- CI-like or constrained shells where GUI/browser launch is unavailable

### Repro context
- Environment: WSL2
- Command: `clerk auth login`
- Issue: Browser does not reliably launch from CLI context, blocking login completion

### Suggested UX details
- Detect browser-launch failures and automatically fall back
- Always print auth URL as a backup, even when auto-open succeeds
- Include copy-friendly output and clear next-step prompts
- Exit with actionable error/help text if auth times out

### Notes from AI on resolving it locally
  Root cause: clerk auth login binds its OAuth callback server to 127.0.0.1 only (via Bun.serve({hostname:"127.0.0.1",
  port:0, ...})) and opens the browser via a plain xdg-open <url> call. In WSL2, depending on .wslconfig networking mode
  / whether wslu's wslview is properly wired as the xdg-open/BROWSER handler, xdg-open either fails silently or opens
  nothing, and the CLI just sits on "Waiting for authentication (timeout in 2m)..." with no fallback message — even
  though the local server itself is listening fine and is reachable from Windows via WSL2's default localhostForwarding.

  What actually unblocked it: other commands in the same CLI (e.g. clerk users open, clerk apps open) already have a
  "browser failed to open" fallback that prints the URL (Could not open your browser automatically. Open this URL to
  continue: ...), but clerk auth login doesn't call that same fallback path — it just hangs silently. I shadowed
  xdg-open on PATH with a one-line script that writes $1 to a file instead of trying to open anything, ran clerk auth
  login in the background, grabbed the captured authorize URL, and opened it manually in a Windows browser. The
  already-running local callback server picked up the redirect fine — confirming the only broken part is "opening the
  browser automatically," not the network path back.

  Suggestions for the devs:
  1. Give clerk auth login the same "couldn't open browser" fallback that users open/apps open/impersonate already have
  — check the same "did xdg-open/open() actually succeed" signal and print the authorize URL on failure instead of just
  silently waiting.
  2. Add a --no-browser / --print-url flag (or auto-detect no-DISPLAY/WSL environments) that skips the open attempt
  entirely and just prints the URL, similar to users open --print.
  3. Detect WSL specifically (e.g. /proc/version containing "microsoft") and default to the print-URL path there, since
  xdg-open in WSL is notoriously inconsistent depending on distro/wslu setup — this is probably the single most common
  environment this bites.
  4. Document the workaround now: for anyone else stuck, clerk config commands accept CLERK_PLATFORM_API_KEY, but that
  key type currently has no self-serve Dashboard page to generate (the "API keys" page under an app/instance only shows
  pk_/sk_) — worth flagging as a second, separate gap: either surface where to mint an ak_... key, or document that it
  isn't self-serve yet

rafa-thayto:
## Summary

Fixes #411 — `clerk auth login` was unusable in WSL2 (and SSH/headless shells): the CLI depended on a browser launch that silently fails there, printed no URL, and hung until a generic timeout.

Three layers of fallback, so login always has a path forward:

- **The authorize URL is always printed** before the launch attempt — a launcher can exit 0 without anything visibly opening (WSL's `xdg-open`), so failure detection alone can't cover it. Printing is safe: completing the token exchange requires the PKCE `code_verifier`, which never leaves the CLI process. When no launcher works, a warning points back at the printed URL.
- **WSL now reaches the Windows host browser without extra setup.** WSL is detected via `WSL_DISTRO_NAME`/`WSL_INTEROP` (fallback: `/proc/version` containing "microsoft") and the launcher chain becomes `wslview → powershell.exe → xdg-open…`. `powershell.exe -NoProfile -NonInteractive -Command Start-Process '<url>'` is on PATH in every default WSL install via Windows interop; the URL is single-quoted (embedded `'` doubled) so `&` in OAuth query strings isn't parsed as a PowerShell operator. `explorer.exe` is deliberately excluded: it opens URLs but always exits 1, which the fast-failure grace race would misread as a failed launch.
- **The auth timeout error is actionable**: re-run `clerk auth login` and copy the printed URL into any browser on the machine.

The full round trip works in WSL2 because it forwards `127.0.0.1` between Windows and the VM in both directions, so the host browser still reaches the CLI's OAuth callback server.

Also in this diff: `login.test.ts` had its ~25-line OAuth-success mock block repeated 14 times (my new tests initially added 3 more) — extracted a `mockOAuthSuccess(overrides)` helper, shrinking the file 804 → 486 lines with coverage unchanged.

Out of scope, possible follow-up: the issue's option 1 (RFC 8628 device authorization flow) needs a server-side endpoint on the Clerk OAuth instance; it's the right answer for SSH-to-remote-box sessions where the callback server lives on the far end.

## Test plan

- [x] 4 new `openBrowser` unit tests: wslview preference, PowerShell fallback, single-quote escaping, no-launcher; login tests assert the URL always prints and the PKCE `code_verifier` never appears in output
- [x] Full suite: 2460 pass / 0 fail; `lint`, `typecheck`, `format:check` clean
- [x] Real unmocked runs: macOS CLI with a lying exit-0 launcher (URL still printed), with no launcher (warning + actionable timeout), Docker Linux with `WSL_DISTRO_NAME` set (real spawn of `powershell.exe Start-Process '<url>'` captured), and `curl` playing the browser against the callback server (CLI advanced to token exchange)
- [ ] Confirm on a physical WSL2 machine that `Start-Process` renders the host browser (the one hop not provable from macOS/Docker)

### DIFF
--- a/.changeset/cli-issue-411-context-fix.md
+++ b/.changeset/cli-issue-411-context-fix.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Fix `clerk auth login` in WSL2, SSH, and headless environments: the sign-in URL is now always printed as a copy-friendly manual fallback, WSL launches the Windows host browser via PowerShell interop when `wslview` is not installed, and the authentication timeout error now explains how to recover.
--- a/packages/cli-core/src/commands/auth/README.md
+++ b/packages/cli-core/src/commands/auth/README.md
@@ -11,7 +11,7 @@ Authenticates the user via an OAuth 2.0 PKCE flow. After a successful login (or 
 1. Checks for an existing valid token — if found, prompts to re-authenticate (in agent mode, skips and runs autoclaim immediately)
 2. Generates PKCE parameters (code verifier, challenge, state)
 3. Starts a local HTTP callback server on `127.0.0.1`
-4. Opens the browser to the Clerk OAuth authorization URL
+4. Prints the Clerk OAuth authorization URL and opens the browser to it. The URL is always printed as a manual fallback for environments where browser launch is unreliable (WSL2, SSH, headless shells). In WSL, the Windows host browser is reached via `wslview` or, when that is not installed, the always-present `powershell.exe` interop binary (`Start-Process`).
 5. Waits for the redirect callback with an authorization code
 6. Exchanges the code for an access token
 7. Stores the token and user info in local config
--- a/packages/cli-core/src/commands/auth/login.test.ts
+++ b/packages/cli-core/src/commands/auth/login.test.ts
@@ -1,4 +1,4 @@
-import { test, expect, describe, afterEach, mock, spyOn } from "bun:test";
+import { test, expect, describe, afterEach, beforeEach, mock, spyOn } from "bun:test";
 import { AuthError } from "../../lib/errors.ts";
 import { useCaptureLog, credentialStoreStubs, configStubs } from "../../test/lib/stubs.ts";
 
@@ -101,6 +101,10 @@ describe("login", () => {
   const captured = useCaptureLog();
   const origSpawn = Bun.spawn;
 
+  beforeEach(() => {
+    consoleSpy = spyOn(console, "log").mockImplementation(() => {});
+  });
+
   afterEach(() => {
     mockGetValidToken.mockReset();
     mockStoreToken.mockReset();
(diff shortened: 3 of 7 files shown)


wyattjoh (approved):

stephlocke:
Thanks @rafa-thayto !

PR #416, opened 2026-08-13
fix(init), feat(deploy), feat(link): proxy-aware middleware, provider domains, stale-link recovery

rafa-thayto:
Three fixes to the path a new project walks: `clerk init` scaffolds middleware, `clerk deploy` creates the production instance, `clerk link` points the project at it. Each one broke for a case that is not exotic — a custom matcher, a Vercel domain, an instance created in the Dashboard after linking.

## `fix(init)`: route `/__clerk` proxy paths and compose past existing matchers

- The generated Next.js matcher skips paths that look like static files, so proxied Clerk requests ending in `.js`/`.css`/`.svg` never reached `clerkMiddleware`. `clerk deploy` points `proxy_url` at `https://<domain>/__clerk`, so `/__clerk/(.*)` is now matched explicitly.
- Composition used to bail whenever the existing middleware declared any `export const config` — the common case for anyone who customized their matcher — leaving the file untouched and `clerkMiddleware` never wired in. It now strips the declaration and composes, and the plan says "replacing its matcher" before you confirm.
- Stripping is a balanced, string-aware parse rather than a strip-to-EOF regex. The old pattern deleted everything after `export const config`, so a config export above the handler silently discarded the user's middleware, and a type-annotated `config: MiddlewareConfig` was not matched at all and produced two config exports.
- Removing the config gate exposed default-export shapes it had been masking: a re-exported default now skips instead of emitting two default exports (invalid ESM), and a class default skips instead of being wrapped into something that throws on every request. Identifier defaults still compose — `export default withAuth(handler)` is legitimate and indistinguishable from a non-callable one without types.

## `feat(deploy)`: support hosting-provider domains

`*.vercel.app` and `*.replit.app` are supported Clerk production domains, served through a proxy rather than CNAME records, but the deploy prompt refused them outright alongside a suffix list that did not match what the API rejects.

`POST /v1/platform/applications/{id}/instances` is the one domain-creating endpoint that runs no domain validation and derives no proxy URL, so a provider domain created there keeps its CNAMEs required and can never verify. Deploy now asks the API to derive the proxy via `PATCH /v1/platform/applications/{id}/domain` right after the instance is created, and again on resume when the derivation never happened. Provider domains get proxy setup instructions instead of DNS records they cannot carry, and the previously-ignored `proxy` component is surfaced in deploy status — without it a proxied domain reports as fully verified while its Frontend API is unreachable.

The prompt still refuses the six shared hosting domains the rest of Clerk refuses (`*.netlify.app`, `*.herokuapp.com`, `*.fly.dev`, `*.onrender.com`, `*.web.app`, `*.railway.app`), mirroring `hostingDomainNames` in clerk_go — that endpoint would otherwise create an instance that can never verify and that the CLI has no way to delete. Error mapping now matches the codes the API really returns: `provider_domain_operation_not_allowed` (403), `home_url_taken` (422), `known_hosting_domain`, `proxy_url_required_for_provider_domain`.

## `feat(link)`: recover from links that predate a production instance

A project linked before its production instance existed — one created afterwards in the Dashboard, say — failed every `--instance prod` command with "No production instance configured. Run `clerk link` to set one up". That was a dead end: `clerk link` only offers to re-link to a *different* application.

Commands that resolve an instance through the linked profile now check the application before reporting. If the instance exists upstream they offer to update the link and continue (agent mode fails instead, pointing at `clerk link --refresh`); if the application genuinely has no production instance they point at `clerk deploy`, the only command that can create one. `clerk link --refresh` re-reads the linked application's instances without changing which application the project is linked to, and never prompts.

## Test plan

- [x] `bun run test` (unit)
- [ ] `bun run test:e2e` — CI
- New coverage: `link-refresh.test.ts`, `deploy/proxy.test.ts`, `deploy/prompts.test.ts`, `deploy/errors.test.ts`, plus additions to `nextjs-app.test.ts`, `deploy/{index,copy,status}.test.ts`, and `integration/error-recovery.test.ts`.

Changesets included for all three (`patch`, `minor`, `minor`).


packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -315,22 +334,82 @@ function renameDefaultMiddlewareExport(existing: string): string | null {
...
+
+/**
+ * Remove an `export const config = ...` declaration wherever it sits in the file.
+ * Clerk's matcher replaces it — keeping the original would leave two config exports
+ * in one module, and a narrower matcher would stop Clerk's own routes (including the
+ * `/__clerk` proxy) from running. Only the declaration is removed: everything after
+ * it is user code, and a config export placed above the handler must not take the
+ * handler with it.
+ */
+function stripMiddlewareConfigExport(existing: string): string {
```
wyattjoh, line 390: This strip only understands a bare `{`/`[` literal. Anything else falls through to `existing.slice(valueStart).search(/[;\n]/)` on line 386, which cuts at the first newline and orphans the remainder. Separately, `endOfLiteral` tracks quotes but not comments, so a `}` inside a `//` comment ends the scan early.

I ran the real `composeWithExistingMiddleware` against three realistic inputs. All three produce files that do not parse:

```ts
// in:  export const config = buildConfig({ matcher: ["/foo"] });
async function middleware(req) { return; }

matcher: ["/foo"],          // <- orphaned
});                          // <- orphaned

export default clerkMiddleware(...)
```

```ts
// in:  export const config = { matcher: ["/x"] } satisfies MiddlewareConfig;
async function middleware(req) {}

satisfies MiddlewareConfig;  // <- orphaned

export default clerkMiddleware(...)
```

```ts
// in:  config object containing the comment `// closing } brace in comment`
async function middleware(req) {}

brace in comment             // <- orphaned
  matcher: ["/x"],
};
```

On `main` all three hit `hasMiddlewareConfigExport → return null → skip`, leaving the user's file untouched. This branch turns that safe skip into a silent overwrite: `previewAndConfirm` only prints `MODIFY  middleware.ts — Add clerkMiddleware to existing middleware, replacing its matcher` (`init/preview.ts:13`) with no diff, so nothing surfaces the corruption before the write.

`export const config = { ... } satisfies MiddlewareConfig` is a mainstream Next.js 14/15 TS idiom, so this is not a corner case.

Suggestion: keep the parser-free strip but bail to `null` (the old skip) whenever the value isn't a clean `{`/`[` literal that `endOfLiteral` closes. The cheapest robust guard is to run the stripped result through `new Bun.Transpiler({ loader: "tsx" }).transformSync(...)` in a try/catch and return `null` on a parse error — every unhandled shape degrades back to a safe skip. Teaching `endOfLiteral` to skip `//` and `/* */` is a separate fix that's also needed.

The `class` and `export { default }` bailouts right above this are exactly the right instinct — return `null` rather than emit something that breaks later. That instinct is what's missing here.

###### Sent from Claude

packages/cli-core/src/commands/deploy/status.ts
```diff
@@ -406,11 +411,7 @@ function deployNextAction(
     );
   }
 
-  const pendingComponents = [
-    !componentStatus.dns ? "DNS" : null,
-    !componentStatus.ssl ? "SSL" : null,
-    !componentStatus.mail ? "email DNS" : null,
-  ].filter((value): value is string => value !== null);
+  const pendingComponents = pendingDeployComponents(componentStatus);
```
wyattjoh, line 473: `pendingDeployComponents` now includes `"proxy"`, and `deployNextAction` feeds it straight into the generic tail on line 424:

> `proxy still provisioning for my-app.vercel.app. Re-run \`clerk deploy status\` in a few minutes, DNS propagation can take time.`

That's backwards. A pending proxy check never clears by waiting — the user's app has to start serving `/__clerk`. And there's no way to recover from the payload: `pendingDnsRecords` is empty for provider domains and `DeployStatusReport` carries no `proxyUrl`, so an agent consuming this JSON has nothing to act on.

The human path gets this right (`copy.ts:236` emits `Waiting for ${domain} to serve Clerk's proxy path: …`). The agent path — the one that can't ask a follow-up question — doesn't.

Suggestion: branch in `deployNextAction` when `!componentStatus.proxy`, mirroring `deployStatusPendingFooter`'s wording, and add `proxyUrl` to `DeployStatusReport`. It's already on the snapshot at line 226, so it's just threading.

###### Sent from Claude

packages/cli-core/src/commands/deploy/prompts.ts
```diff
@@ -7,16 +7,30 @@ import { type OAuthPromptField, type OAuthProviderDescriptor } from "./providers
...
+ * instances` — the endpoint `clerk deploy` uses — runs no domain validation at
+ * all, and PLAPI exposes no way to delete a production instance afterwards.
+ * Until that endpoint validates, this prompt is the last exit before an
+ * unusable, undeletable instance.
+ *
+ * Provider domains (*.vercel.app, *.replit.app) are deliberately absent:
+ * Clerk supports them as production domains, served through a proxy rather
+ * than CNAME records. See `configureProviderDomainProxy` in ./proxy.ts.
+ */
+const KNOWN_HOSTING_DOMAIN_SUFFIXES = [
```
wyattjoh, line 23: The doc comment states the guard's purpose precisely: `POST .../instances` runs no domain validation and PLAPI can't delete a production instance, so this prompt is "the last exit before an unusable, undeletable instance." But the list was then set to mirror `hostingDomainNames` — which is the set the API *already* rejects, i.e. the one set that never needed a client-side guard at all.

`.pages.dev` and `.clerk.app` were on the old list and are gone, and `prompts.test.ts` now asserts both are accepted. Neither is in `hostingDomainNames`, and neither is a provider domain (`constants` defaults are `replit.app` and `vercel.app` only). `pages.dev` is a PSL private suffix, so `validators.DomainName`'s `ViaPSL` check passes too.

Net result for `myapp.pages.dev`: the CLI accepts it, the API accepts it, an ordinary CNAME-verified production instance gets created — and it can never verify, because you cannot add `clerk.myapp.pages.dev` under Cloudflare's zone. That is exactly the outcome this guard exists to prevent, and Cloudflare Pages is a common Next.js target.

## Suggestion

Keep the clerk_go mirror as the "API will reject this" tier, and re-add a second tier for domains the API accepts but whose DNS the user provably can't control (`.pages.dev`, `.workers.dev`, `.clerk.app`, `.github.io`). The README's Production Domain Policy table already has the right shape for two tiers, and the "delete this list once CreateInstance validates" note only retires the first one.

###### Sent from Claude

packages/cli-core/src/lib/plapi.ts
```diff
@@ -253,6 +253,25 @@ export async function createProductionInstance(
   return response.json() as Promise<ProductionInstanceResponse>;
 }
 
+export type UpdateApplicationDomainParams = {
+  name?: string;
+  proxy_path?: string;
```
wyattjoh: `resolveProviderDomainProxyUrl` only ever sends `{ name }` and relies on the server-side derivation in `UpdateParams.Normalize()`. Since supplying an explicit `proxy_path` changes validation behavior on the API side — `validateProxyPath` runs, and a blank value actively *clears* the proxy (`params.proxyURL = nil`) — an unused knob here is a trap for the next caller who assumes it's inert. Worth dropping until something needs it.

###### Sent from Claude

packages/cli-core/src/commands/link/index.ts
```diff
@@ -38,6 +52,12 @@ export async function link(options: LinkOptions = {}): Promise<void> {
   const displayPath = repoRoot ?? cwd;
 
   const existing = await resolveProfile(cwd);
+
+  if (options.refresh) {
```
wyattjoh, line 56: The refresh branch returns before `options.app` is ever consulted, so `clerk link --refresh --app app_other` quietly refreshes the *currently* linked app instead of doing anything with `--app`. Given `--refresh` is documented as the non-interactive remedy for agents and CI, a silent no-op on a flag the caller explicitly passed is the wrong failure mode. `throwUsageError` on the combination would be clearer.

###### Sent from Claude

packages/cli-core/src/commands/deploy/copy.ts
```diff
@@ -164,17 +171,44 @@ export function deployComponentLabels(
...
  * email DNS records. Each value comes from the same domain status response.
  */
 export function deployComponentStatus(status: DeployComponentStatus): string {
   const mark = (ok: boolean) => (ok ? green("✓") : yellow("pending"));
-  return `DNS: ${mark(status.dns)}  SSL: ${mark(status.ssl)}  Email DNS: ${mark(status.mail)}`;
+  const line = `DNS: ${mark(status.dns)}  SSL: ${mark(status.ssl)}  Email DNS: ${mark(status.mail)}`;
+  // Only proxied domains have a proxy check; elsewhere the API reports it as
+  // not required and a permanent "Proxy: ✓" would be noise.
+  const proxyPending = !status.proxy;
+  return proxyPending ? `${line}  Proxy: ${mark(status.proxy)}` : line;
```
wyattjoh: Inside `proxyPending ? … : line`, `status.proxy` is always `false`, so `mark(status.proxy)` can only ever return `yellow("pending")` — it could just be that literal.

More worth a second look than the dead call: because the segment is only rendered while pending, the user watching a provider-domain deploy sees `… Proxy: pending` and then the segment simply disappears, rather than turning green. They never get positive confirmation that the one check standing between them and a working Frontend API actually passed. Suppressing a permanent `Proxy: ✓` on CNAME domains makes sense; suppressing the transition on the domains where it's the only meaningful check may not.

###### Sent from Claude

packages/cli-core/src/commands/link/index.ts
```diff
@@ -108,6 +128,43 @@ export async function link(options: LinkOptions = {}): Promise<void> {
...
+
+  await ensureAuth();
+  intro("Refreshing link");
+
+  const app = await withSpinner(`Fetching ${profileLabel(existing.profile)}...`, () =>
+    fetchLinkedApplication(existing.profile),
+  );
+  const { profile, updated } = await refreshProfileInstances(existing.path, existing.profile, app);
+
+  if (updated.length === 0) {
```
wyattjoh: `refreshProfileInstances` deliberately writes to disk when `refreshed.appName !== profile.appName` even with `updated.length === 0` — the comment there explains why. But the caller only inspects `updated`, so a run that renamed the app on disk still prints "Already up to date". Surfacing the name change in `RefreshResult` (alongside `updated`) would let this report what it actually did.

###### Sent from Claude

packages/cli-core/src/commands/deploy/errors.test.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+async function mapped(error: PlapiError): Promise<CliError | PlapiError> {
+  try {
+    await mapDeployError(Promise.reject(error));
+  } catch (caught) {
+    return caught as CliError | PlapiError;
+  }
+  throw new Error("expected mapDeployError to reject");
+}
```
wyattjoh: There's a lot of glue with these tests, is it possible to just test the `mapDeployError` directly here instead of only through this helper?

packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -254,8 +254,12 @@ function nextjsMiddlewareHandler(returnStatement = ""): string {
 function nextjsMiddlewareConfig(): string {
   return `export const config = {
   matcher: [
+    // Skip Next.js internals and all static files, unless found in search params
     "/((?!_next|[^?]*\\\\.(?:html?|css|js(?!on)|jpe?g|webp|png|gif|svg|ttf|woff2?|ico|csv|docx?|xlsx?|zip|webmanifest)).*)",
+    // Always run for API routes
     "/(api|trpc)(.*)",
+    // Always run for Clerk-specific frontend API routes
+    "/__clerk/(.*)",
```
wyattjoh, line 262: Is this an established pattern path or is this new?

packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -290,7 +294,22 @@ ${content}
  * Compose Clerk middleware with existing non-Clerk middleware.
  * Renames the existing default export and wraps it inside clerkMiddleware.
  */
```
wyattjoh, line 296: Should move this to the function that was moved down so it's not detached.

packages/cli-core/src/lib/config.ts
```diff
@@ -397,6 +397,25 @@ export function resolveFetchedApplicationInstance(
...
+async function resolveLinkedInstance(
+  profileKey: string,
+  profile: Profile,
+  instance: string | undefined,
+): Promise<{ id: string; label: string }> {
+  try {
+    return resolveInstanceId(profile, instance);
+  } catch (error) {
+    const { recoverMissingInstance } = await import("./link-refresh.ts");
+    return recoverMissingInstance(error, profileKey, profile, instance);
```
wyattjoh: Why are we dynamically importing here? Why not just include in top level imports?

packages/cli-core/src/lib/config.ts
```diff
@@ -308,7 +308,7 @@ export async function resolveProfile(cwd: string): Promise<
   return undefined;
 }
 
-const INSTANCE_ALIASES: Record<string, "development" | "production"> = {
+export const INSTANCE_ALIASES: Record<string, "development" | "production"> = {
```
wyattjoh: Good idea to maybe swap this over to the following:

```ts
export const INSTANCE_ALIASES = {
  dev: "development",
  development: "development",
  prod: "production",
  production: "production",
} satisfies Record<string, "development" | "production">;
```

So that we can have a new type:

```ts
export type InstanceAliasKeys = keyof typeof INSTANCE_ALIASES; // "dev" | "development" | "prod" | "production"
```

To constrain the new function in the `recoverMissingInstance` `key` parameter.

packages/cli-core/src/commands/init/frameworks/helpers.ts
rafa-thayto: `/__clerk` is the path Clerk's Frontend API proxy uses for provider domains (established on the backend — see the [proxy-fapi docs](https://clerk.com/docs/guides/dashboard/dns-domains/proxy-fapi)). The matcher entry is new in this PR because keyless + Vercel testing flagged its absence: without it, `clerkMiddleware` never runs for proxied FAPI requests. Source: the ["clerk cli keyless + vercel deployment friction" Notion doc](https://app.notion.com/p/clerkdev/clerk-cli-keyless-vercel-deployment-friction-3b92b9ab44fe802bbfddc4ec91a87227).

packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -315,28 +334,120 @@ function renameDefaultMiddlewareExport(existing: string): string | null {
...
+/**
+ * Remove an `export const config = ...` declaration wherever it sits in the file.
+ * Clerk's matcher replaces it — keeping the original would leave two config exports
+ * in one module, and a narrower matcher would stop Clerk's own routes (including the
+ * `/__clerk` proxy) from running. Only the declaration is removed: everything after
+ * it is user code, and a config export placed above the handler must not take the
+ * handler with it.
+ */
+function stripMiddlewareConfigExport(existing: string): string {
+  const declaration = /export\s+const\s+config\b[^=]*=\s*/.exec(existing);
```
dominic-clerk, line 391: **[HIGH] `stripMiddlewareConfigExport` regex can delete the user's existing auth middleware**

`/export\s+const\s+config\b[^=]*=\s*/` matches the first textual `export const config` — including one inside a `//` comment or JSDoc — and `[^=]*` then spans newlines up to the next `=` anywhere in the file, so everything in between (e.g. the whole `export default function middleware(){...}` auth gate) is spliced out. When the deleted span is a balanced construct the result still parses, `composedOrNull` accepts it, and `clerk init --yes`/agent mode writes a bare `clerkMiddleware()` with no confirmation — silently removing authorization. Parse the config declaration with the transpiler/AST (or at least anchor the match to a statement start and reject matches inside comments/strings) instead of the regex.

— Comment generated with Claude with @dominic-clerk's supervision

packages/cli-core/src/commands/init/frameworks/helpers.ts
```diff
@@ -315,28 +334,120 @@ function renameDefaultMiddlewareExport(existing: string): string | null {
...
+    return source;
+  } catch {
+    return null;
+  }
 }
 
 export function composeWithExistingMiddleware(existing: string): string | null {
   const clerkImport = `import { clerkMiddleware } from "@clerk/nextjs/server";\n`;
   const preamble = clerkImport + "\n";
+  const stripped = stripMiddlewareConfigExport(existing);
```
dominic-clerk, line 432: **[MEDIUM] Replacing the user's matcher with `nextjsMiddlewareConfig()` drops authorization on extension-bearing paths**

`composeWithExistingMiddleware` now strips the existing `export const config` and substitutes Clerk's matcher, whose first entry excludes any path that looks like a static file (`.csv`, `.zip`, `.xlsx`, ...). Because the user's middleware is wrapped inside `clerkMiddleware`, its own checks are governed by that matcher too, so a route like `/admin/export.csv` previously protected by `matcher: ["/admin/:path*"]` becomes reachable unauthenticated after `clerk init` (applied without a diff, and with no prompt under `--yes`/agent mode). Merge the user's matcher entries into the new config rather than replacing them, or keep returning `null` when a config export exists.

— Comment generated with Claude with @dominic-clerk's supervision

PR #418, opened 2026-08-17
fix(auth): revoke OAuth sessions on logout

wyattjoh:
## Summary

`clerk auth logout` previously deleted the stored OAuth session locally and never told the authorization server, so the refresh token stayed redeemable until its natural expiry — the local delete was a false reassurance on shared machines, CI runners, and any host the user no longer trusts. The same gap applied to `clerk auth login` when re-authenticating over a live session: a new grant was minted while the old one stayed valid.

Both paths now call the OAuth 2.0 Token Revocation endpoint (RFC 7009) at `/oauth/token/revoke`, which the instance advertises as `revocation_endpoint` in its discovery document and which accepts a public client presenting only its `client_id`. Logout revokes the stored refresh token and then deletes locally in a `finally`, so the local delete is never skippable. Re-authentication captures the outgoing refresh token before the browser flow and revokes it only after the replacement is stored, so an abandoned or failed flow leaves the original session intact rather than stranding the user with no credentials.

Revocation is best-effort throughout: the caller is already discarding the credentials, so a network failure or a server error is logged under `--verbose` and never blocks logout or re-authentication. Per RFC 7009 §2.2 the endpoint also answers `200` for a token it does not recognise, so an already-expired session is indistinguishable from a successful revocation. A refresh that has already failed with `invalid_grant` skips revocation entirely, since that grant is spent server-side and there is nothing left to revoke.

## Test plan

- [x] `bun run test` — 2607 pass, 0 fail
- [x] Verified against the live endpoint that a public client sending only `token` + `client_id` receives `200`, and that omitting `client_id` receives `400 invalid_request`
- [x] Manual, logout: captured a freshly issued refresh token without redeeming it, ran `auth logout --verbose`, confirmed the `POST /oauth/token/revoke` request, then confirmed the captured token was rejected with `invalid_grant`
- [x] Manual, re-authentication: captured the outgoing refresh token without redeeming it, completed `auth login -y`, confirmed a different session was stored, that the old grant was rejected with `invalid_grant`, and that the revoke request was issued only after the replacement was written to the keyring

### DIFF
--- a/.changeset/oauth-revoke-on-logout.md
+++ b/.changeset/oauth-revoke-on-logout.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Revoke the OAuth session with Clerk when signing out, so `clerk auth logout` ends the session everywhere instead of only deleting the local credentials. Re-authenticating over an existing session now revokes the previous one as well.
--- a/packages/cli-core/src/commands/auth/README.md
+++ b/packages/cli-core/src/commands/auth/README.md
@@ -15,7 +15,8 @@ Authenticates the user via an OAuth 2.0 PKCE flow. After a successful login (or 
 5. Waits for the redirect callback with an authorization code
 6. Exchanges the code for an access token
 7. Stores the token and user info in local config
-8. **Autoclaim**: if `.clerk/keyless.json` exists in the current directory, claims the temporary application, links it to the project, and pulls environment variables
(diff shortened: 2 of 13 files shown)


wyattjoh:
**Stack: wyattjoh/oauth-revoke-on-logout**

- **#418 👈 this PR**

rafa-thayto (approved):

PR #420, opened 2026-08-17
fix(cli): exit 130 when Ctrl-C interrupts an operation

wyattjoh:
## Summary

Ctrl-C now reports what the CLI was doing when it arrived. Interrupting anything in progress — an in-flight request, a poll interval or retry backoff, a project generator run by `clerk init`, or `clerk webhooks listen` once it has drained — exits 130, and the interrupted run is reported to telemetry instead of going unrecorded. Only waiting on a human is a clean exit: cancelling a prompt, abandoning browser sign-in, or quitting `$EDITOR` all exit 0.

The 130 is emitted by dying from the signal rather than calling `process.exit(130)`. Both make the shell print 130, but only a real signal death sets `WIFSIGNALED`, which is what a wrapping script inspects; with `exit(130)` a script keeps running after the user pressed Ctrl-C.

A timer is deliberately not a human wait. `clerk deploy status --wait` spends roughly 93 seconds of a 95-second run asleep between polls, so treating a sleep as idle meant Ctrl-C almost always exited 0 — which is exactly what that command's exit code means to a script, "the deploy is complete". `clerk deploy` and its resumable pause report 130 for the same reason: a production instance exists but DNS or OAuth does not, so `clerk deploy && ./cutover.sh` must stop either way. Because `runProgram` hands rendering to the signal handler once an interrupt is latched, both commands now print their resume hint and partial status from their own catch blocks rather than through a thrown error's message.

The bookkeeping tail is not a wait either. Interrupting the update check or the closing animation exits 130, which costs about two seconds per successful command where Ctrl-C halts a wrapping script even though the command itself finished. That is the intended trade: a Ctrl-C is a Ctrl-C, and the exit code says so.

An interrupt also has to be classified the same way everywhere it surfaces. Ctrl-C aborts the in-flight request, so it arrives as an `AbortError` rather than a `UserAbortError`, and callers that recognised only the latter treated a cancellation as a failure: the command wrappers closed their gutter with "Failed", and `clerk init --keyless` blamed the network for a fifteen-second timeout and carried on with the rest of setup. Both now share the predicate `withSpinner` already used. The telemetry bypass is narrowed for the same reason — only the shutdown flush reports the very interrupt that triggered it, so only it may outlive one; applying that bypass to every send let a Ctrl-C mid-POST record the run's success event, with the context already cleared so the handler could not replace it with the abort.

Finally, an interrupted `clerk deploy status --wait` now reports what the last completed poll established rather than the snapshot taken before the wait began, so a component that verified moments before the Ctrl-C is no longer listed as pending.

## Test plan

- [x] `bun run test` — 2641 pass, 0 fail
- [x] Unit coverage for the human-wait counter, the latch, `sleep`'s classification, and every `exitInterrupted` guard
- [x] Subprocess test asserting the shell-visible contract: `WIFSIGNALED` / `WTERMSIG=2` for work, `WEXITSTATUS=0` for a human wait
- [x] Verified a poll-loop-shaped interrupt dies from SIGINT and stops a `&&` chain, and that a human wait still exits 0 and continues it
- [ ] E2E (`bun run test:e2e`) — left to CI


packages/cli-core/src/lib/signals.ts
```diff
@@ -1,10 +1,169 @@
...
+ * can `process.removeListener("SIGINT", CLI_SIGINT_HANDLER)` and install its own
+ * graceful-drain handling without disturbing anything else.
+ *
+ * Async on purpose: awaiting the telemetry flush keeps the event loop alive
+ * long enough to report the interrupted run, which `process.exit` from a
+ * synchronous handler never could. A second Ctrl-C during that window quits
+ * immediately.
+ */
+export const CLI_SIGINT_HANDLER = async (): Promise<void> => {
+  if (interrupted !== null) exitInterrupted(interrupted);
```
rafa-thayto: This guard calls `exitInterrupted` without `return`, which is the exact pattern the new interrupts.md rule warns about ("Call it as `return exitInterrupted(...)` at call sites too"). Under a no-op `process.exit` stub, the same stub `listen.test.ts` installs, a second Ctrl-C falls through into `beginInterrupt()`, the raw-mode and cursor writes, and a second `reportAndExitInterrupted`, so telemetry flushes twice. The existing test only passes because its stub throws instead of no-op'ing.

```suggestion
  if (interrupted !== null) return exitInterrupted(interrupted);
```

packages/cli-core/src/commands/webhooks/listen.ts
```diff
@@ -134,15 +139,20 @@ export async function webhooksListen(options: WebhooksListenOptions = {}): Promi
...
   process.on("SIGINT", () => {
     if (shuttingDown) {
-      // Double Ctrl+C: force-quit immediately.
-      process.exit(EXIT_CODE.SIGINT);
+      // Double Ctrl+C: force-quit immediately, skipping the telemetry flush the
+      // drain would otherwise do. `return` because a stubbed `process.exit`
+      // under test would otherwise fall through and start a second drain.
+      return exitInterrupted(EXIT_CODE.SIGINT);
     }
     void (async () => {
```
rafa-thayto, line 154: The drain is a fire-and-forget `void (async () => ...)()` with no `.catch`. If anything in it rejects (the dynamic telemetry import inside `reportAndExitInterrupted`, or `client?.stop()` throwing), the rejection is unobserved and the process never reaches the signal-death exit this PR exists to guarantee. A `.catch` that falls back to `exitInterrupted` keeps the contract even on an unexpected error.

```ts
    (async () => {
      beginInterrupt();
      shuttingDown = true;
      resolveSetupGate();
      client?.stop();
      const pending = [...inFlight, ...(tokenRotationTask ? [tokenRotationTask] : [])];
      await Promise.race([
        Promise.allSettled(pending),
        new Promise<void>((resolve) => setTimeout(resolve, 2_000)),
      ]);
      await reportAndExitInterrupted(EXIT_CODE.SIGINT);
    })().catch(() => exitInterrupted(EXIT_CODE.SIGINT));
```

packages/cli-core/src/lib/fetch.ts
```diff
@@ -52,16 +62,26 @@ export interface ApiResponse {
   rawBody: string;
 }
 
+/** Ctrl-C aborts the request, without discarding a caller's own timeout signal. */
+function interruptSignalFor(
+  own: RequestInit["signal"],
+  ignoreInterrupt: boolean | undefined,
+): RequestInit["signal"] {
+  if (ignoreInterrupt) return own;
+  return own ? AbortSignal.any([own, interruptSignal()]) : interruptSignal();
```
rafa-thayto, line 71: `AbortSignal.any([own, interruptSignal()])` registers an abort listener on the shared process-lifetime signal for every request that brings its own signal, and the listener is only released when the derived signal is garbage collected. `webhooks/forward.ts` passes `AbortSignal.timeout(30_000)` per delivery and deliveries run concurrently, so a burst can pile enough listeners on the shared signal to trip the default max-listeners warning. Raising the ceiling on the shared signal documents that it is intentionally observed by unbounded concurrent requests.

```ts
// in signals.ts
import { setMaxListeners } from "node:events";

let controller = new AbortController();
setMaxListeners(0, controller.signal);

export function _resetInterruptState(): void {
  controller = new AbortController();
  setMaxListeners(0, controller.signal);
  waits = 0;
  interrupted = null;
}
```

packages/cli-core/src/commands/deploy/status-command.ts
```diff
@@ -35,10 +36,22 @@ export async function deployStatus(options: DeployStatusOptions = {}): Promise<v
...
+  if (state.kind === "active" && shouldWait) {
+    try {
+      outcome = await runWait(state, { triggerCheck: !preflightTriggered });
+    } catch (error) {
+      if (interruptedExitCode() === null) throw error;
+      // Ctrl-C mid-poll. `runProgram` hands the exit to the signal handler, so
+      // nothing below this frame runs and the command would otherwise print
+      // nothing at all. Emit what the last completed poll established; the exit
+      // code stays 130, so no script reads this as a finished deploy.
+      emitReport(buildDeployStatusReport(state, null));
```
rafa-thayto: The comment says this emits "what the last completed poll established", but `state` here is the snapshot resolved before `runWait` started, and `waitForDeployStatus` never writes its polled status back into it. If DNS turned valid on the last poll before Ctrl-C, the report still shows the pre-wait pending status. Either thread the latest polled status out of the wait loop, or reword the comment to say the report reflects the pre-wait state.

```ts
let lastKnownStatus: DeployComponentStatus | undefined;
try {
  outcome = await runWait(state, {
    triggerCheck: !preflightTriggered,
    onProgress: (status) => { lastKnownStatus = status; },
  });
} catch (error) {
  if (interruptedExitCode() === null) throw error;
  const partialState = lastKnownStatus
    ? { ...state, snapshot: { ...state.snapshot, componentStatus: lastKnownStatus } }
    : state;
  emitReport(buildDeployStatusReport(partialState, null));
  throw error;
}
```

packages/cli-core/src/commands/api/index.ts
```diff
@@ -157,7 +156,7 @@ export async function api(
       throw error;
     }
   } catch (error) {
-    closeStatus = error instanceof UserAbortError || isPromptExitError(error) ? "paused" : "failed";
+    closeStatus = error instanceof UserAbortError ? "paused" : "failed";
```
rafa-thayto: This exact ternary now appears verbatim in five files this PR edits (here, apps/create.ts, apps/list.ts, config/push.ts, users/list.ts). Since every call site was already open in the diff, a shared helper next to `UserAbortError` would keep the classification in one place.

```ts
// errors.ts
export function closeStatusForError(error: unknown): "paused" | "failed" {
  return error instanceof UserAbortError ? "paused" : "failed";
}

// each call site
closeStatus = closeStatusForError(error);
```

packages/cli-core/src/commands/deploy/index.ts
```diff
@@ -87,10 +82,19 @@ export async function deploy(_options: DeployOptions = {}) {
     const ctx = await resolveDeployContext();
     await runDeploy(ctx);
   } catch (error) {
+    // Ctrl-C during a request or a DNS poll rejects with an `AbortError`, not a
+    // `UserAbortError`, and `runProgram` hands rendering to the signal handler
+    // the moment an interrupt is latched — so nothing below would ever print.
+    // A half-finished deploy is exactly when the resume hint matters, so emit it
+    // here as a side effect rather than relying on the thrown error's message.
+    if (interruptedExitCode() !== null && isInsideGutter()) {
```
rafa-thayto, line 90: This catch calls `isInsideGutter()` three times and `pausedOutro(pausedOperationNotice())` twice across three sequential ifs, which reads as three independent conditions when there are really two outcomes. Hoisting the gutter check and collapsing the duplicate paused call makes the branching visible.

```ts
} catch (error) {
  const insideGutter = isInsideGutter();
  if (interruptedExitCode() !== null) {
    if (insideGutter) pausedOutro(pausedOperationNotice());
    throw error;
  }
  if (error instanceof DeployPausedError) {
    if (insideGutter) outro("Paused");
  } else if (error instanceof UserAbortError) {
    if (insideGutter) pausedOutro(pausedOperationNotice());
    throw new UserAbortError();
  }
  throw error;
}
```

rafa-thayto (changes requested):

packages/cli-core/src/commands/deploy/index.ts
wyattjoh: `isInsideGutter()` isn't a pure read. `outro` and `pausedOutro` both call `popPrefix()` (`lib/spinner.ts:54` and `:76`), and `runDeploy` closes its own gutter on several paths — `outro("Cancelled")` at `:157` and `:364`, `outro("Success")` at `:642` — so by the time this catch runs the prefix may already be popped. The `finally` at `:105` re-reads it for exactly that reason: it's what stops a second `outro("Failed")` printing after an inner `outro("Cancelled")` already closed the block. Hoisting to a single `const insideGutter` is safe *inside* the catch today, since `DeployPausedError` and `UserAbortError` are mutually exclusive and only one branch runs — but it puts a stale value one edit away from the `finally`, where it would double-print on every cancelled deploy.

The restructure also isn't behavior-preserving. Line `:90` currently requires both `interruptedExitCode() !== null` **and** `isInsideGutter()`. When an interrupt is latched but the gutter is already closed, it falls through to `:97` and rethrows a fresh `UserAbortError()` rather than the original `error`. Moving the gutter check into an inner `if` makes the interrupt branch throw unconditionally, changing what escapes on that path.

I've left the duplication for now. If you think the fall-through at `:90` is itself the bug — that a latched interrupt should always throw `error` regardless of gutter state — I'd agree that's worth fixing, but as a deliberate behavioral change rather than a side effect of a readability pass. Is that what you were after?

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -232,6 +236,12 @@ async function buildAndSend(
     body: JSON.stringify({ events: [event] }),
     signal,
     bestEffort: true,
+    // Only the shutdown flush reports the interrupt, so only it may outlive
+    // one. A normal end-of-command flush must stay interruptible: bypassing
+    // the signal there would let a Ctrl-C mid-POST record the run's success
+    // event, and `context` is already cleared so the handler cannot replace
+    // it with the abort event.
+    ignoreInterrupt: outlivesInterrupt,
```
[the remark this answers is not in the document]
wyattjoh: Aborting that request is the intended half of this change. Previously `ignoreInterrupt: true` was passed on *every* send, so a Ctrl-C mid-POST let the normal `outcome: "success"` event complete and be recorded for a run the user had just interrupted. `.claude/rules/interrupts.md` scopes that bypass to the shutdown flush only ("exists solely for the shutdown telemetry flush — do not add other callers"), which is what this now implements.

The second half — the shutdown flush finding no `context` — is real, but it predates this PR rather than being introduced by it. `finalizeAndSendTelemetry` clears `context` unconditionally at entry, and those lines are untouched here:

```ts
const current = context;
context = null;
if (!current) return;   // the shutdown flush no-ops here
```

So the abort event was never being sent in this window. What changed is that the incorrect success event no longer masks the gap: it went from recording the wrong outcome to recording nothing. That is a strict improvement for data integrity, but it is not yet the behavior the rule describes.

Fixing it properly means holding `context` until a send actually lands, so an aborted normal flush leaves it for the shutdown flush to re-send as `"abort"`. That needs a de-dupe latch, because both flushes can be in flight simultaneously (1500ms normal deadline vs. the 250ms interrupt budget) and the handler's dynamic `import("./telemetry.ts")` makes the ordering racy. Getting that wrong double-sends events, which is worse than the current gap. Tracking it as a follow-up rather than expanding this PR's scope.

packages/cli-core/src/commands/api/index.ts
wyattjoh: Extracted, with two deviations from the suggestion.

It could not live in `errors.ts`: the predicate needs `interruptedExitCode()`, and `signals.ts` already imports `EXIT_CODE` from `errors.ts`, so that direction is an import cycle. It also could not live in `spinner.ts` next to the existing private `isCancelled` — 17 test files stub that module with `mock.module`, and adding an export there failed 49 tests with `SyntaxError: Export named 'closeStatusForError' not found in module`. It ended up in `lib/signals.ts`, which owns the interrupt state and is mocked by no test file.

The second deviation is the predicate itself. `UserAbortError` alone is not the right test here: Ctrl-C aborts the in-flight request, so it surfaces as an `AbortError` and these wrappers were closing their gutter with "Failed" on a plain interrupt. The helper reuses the same condition `withSpinner` already applied (`spinner.ts:90`), so the wrappers and the spinner now agree on what counts as a cancellation:

```ts
export function closeStatusForError(error: unknown): "paused" | "failed" {
  return isCancelled(error) ? "paused" : "failed";
}
```

All five call sites now use it.

packages/cli-core/src/commands/deploy/status-command.ts
wyattjoh: Took the first option — threading the polled status out — since the comment was describing the behavior that was intended rather than the one that shipped, and the report genuinely could list a component as pending after it had verified.

`DeployProgressHandlers` gained an optional `onStatus`, fired in `waitForDeployStatus` after each poll resolves a fresh status (both the initial read and each retry iteration). `deployStatus` captures the latest into `lastPolledStatus` and passes it on the interrupt path:

```ts
const partial = lastPolledStatus ? { verified: false, status: lastPolledStatus } : null;
emitReport(buildDeployStatusReport(state, partial));
```

`buildDeployStatusReport` already prefers `outcome?.status ?? snapshot.componentStatus`, so the fresh status flows into `pendingDnsRecords` via `cnameTargetPending` and a record that verified on the last poll is no longer reported as pending. `verified: false` is always correct on this path — had a poll returned complete, `waitForDeployStatus` would have returned normally and never reached the catch. When no poll completed, it passes `null` and falls back to the pre-wait snapshot exactly as before.

rafa-thayto (approved):

PR #422, opened 2026-08-17
fix(auth): report OAuth revocation failures

wyattjoh:
## Summary

Follow-up to #418, which could fail to revoke a session while still reporting success. `revokeToken` now returns its outcome instead of void, checks the response status so a permanent 4xx stops looking like success, and resolves the OAuth config inside its try block so a malformed `CLERK_OAUTH_BASE_URL` no longer aborts sign-out before `clearAuth` runs. Logout and re-auth warn on failure instead of printing an unconditional success, and logout flags `CLERK_PLATFORM_API_KEY` when set.

The outgoing session is now read inside the OAuth flow, just before the token exchange overwrites it, rather than before the browser flow and gated on a network probe that swallows its errors. That gate could orphan a live grant; the earlier placement also let a concurrent refresh rotate the token away, which makes revocation a silent no-op. The new placement closes the first and shrinks the second to one token request.

Docs corrected: revocation ends the current environment's grant and its refresh token, but the access token is a JWT the server refuses to revoke, valid until it expires.

## Test plan

- [x] `format:check`, `lint`, `typecheck`, `test` (2618 pass, 0 fail)
- [x] New tests fail when either fix is reverted; both mutations passed the old suite
- [ ] Manual: `clerk auth logout --verbose`, confirm the refresh token no longer redeems
- [ ] Manual: `CLERK_OAUTH_BASE_URL=not-a-url clerk auth logout`, confirm it warns and still clears

### DIFF
--- a/.changeset/oauth-revoke-hardening.md
+++ b/.changeset/oauth-revoke-hardening.md
@@ -1,0 +1,7 @@
+---
+"clerk": patch
+---
+
+Report when signing out cannot revoke the session with Clerk, instead of always printing a clean sign-out. Local credentials are still removed either way, and `clerk auth logout` now also warns when `CLERK_PLATFORM_API_KEY` is set, since that key keeps authenticating requests.
+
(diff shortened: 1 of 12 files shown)


wyattjoh:
**Stack: wyattjoh/oauth-revoke-hardening**

- **#422 👈 this PR**

rafa-thayto (approved):

PR #425, opened 2026-08-18
fix(cli): report interrupted runs in deploy status and telemetry

wyattjoh:
## Summary

Two reporting gaps left open by #420, both on the Ctrl-C path.

`clerk deploy status` prints nothing at all when an interrupt lands before its wait loop starts. Only the wait was covered by the interrupt catch that #420 added, so an interrupt during application resolution, the preflight DNS check, or the state read escaped uncovered, and `runProgram` returns early without rendering anything. The catch now spans the whole command and emits whatever has been established: a full report once the live state resolves, and a `state: "interrupted"` report before that. The latter asserts nothing about the deploy, so agent-mode stdout stays parseable without claiming there is no production instance — the question an interrupted run never got to answer.

An interrupted run also sent no telemetry event at all. `finalizeAndSendTelemetry` cleared the module context at entry, so a Ctrl-C mid-POST left the shutdown flush with nothing to report and the `outcome: "abort"` event was never sent. The context is now held until a send actually lands, which is enough on its own to keep a run to one event: a normal flush still running when the shutdown flush starts can no longer land, because its POST is composed with the interrupt signal that has already been aborted, and one that landed before the interrupt has already recorded itself. The shutdown flush therefore never waits on the normal one, whose remaining config and Git reads observe no signal and could otherwise burn its entire 250ms budget.

One case is left alone deliberately: a Ctrl-C in the bookkeeping tail after the POST has landed still records the success. That event cannot be retracted, and sending a second would double-count the run.

## Test plan

- [x] `bun run format:check`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run test` — 2650 pass, 0 fail

### DIFF
--- a/.changeset/interrupt-report-gaps.md
+++ b/.changeset/interrupt-report-gaps.md
@@ -1,0 +1,8 @@
+---
+"clerk": patch
+---
+
+Report what was established when Ctrl-C interrupts `clerk deploy status`, and record interrupted runs in usage telemetry.
+
+- `clerk deploy status` no longer exits silently when the interrupt arrives before the deploy state is read. Agent mode emits a report with `state: "interrupted"`, which asserts nothing about the deploy.
+- An interrupted command now sends its `abort` outcome instead of no event at all.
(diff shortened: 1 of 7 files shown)


wyattjoh:
**Stack: wyattjoh/interrupt-report-gaps**

- **#425 👈 this PR**

rafa-thayto (approved):

PR #434, opened 2026-08-21
fix(version): honor CLI_VERSION define under Bun 1.4.0

rafa-thayto:
## Summary

CI started failing on every PR today: the 3 `compiled version` tests in `version.test.ts` fail because setup-bun now resolves `>=1.3.13` to the freshly released **Bun 1.4.0**, which no longer substitutes `--define` globals during **macro execution**. The `typeof CLI_VERSION` guard inside `version.macro.ts` therefore always saw `undefined`, and compiled binaries ignored the injected version and reported the checkout-derived dev version. This would also have broken the next **release build** (binaries reporting `3.1.0-dev.*`, flipping `IS_DEV_BUILD` behaviors: keychain namespace, update-check suppression).

Fix: the macro no longer reads `CLI_VERSION` at all — it returns only the checkout-derived fallback string. The define check and the one-line dev classification now live in `version.ts` module code, where `--define` still folds reliably on both 1.3.x and 1.4.0. `.claude/rules/versioning.md` updated to match.

Repro (before this fix): `bunx bun@1.4.0 test packages/cli-core/src/lib/version.test.ts` fails the 3 compiled-version tests on plain main; Bun 1.3.14 passes.

## Test plan
- `bun test version.test.ts`: 9/9 pass on Bun 1.3.14 AND Bun 1.4.0
- `bun run lint`, `typecheck`, `test` (2653 pass) green

Note: PR #433 is red on CI solely because of this — rerun it after this merges.

### DIFF
--- a/.changeset/version-define-bun-1-4.md
+++ b/.changeset/version-define-bun-1-4.md
@@ -1,0 +1,5 @@
+---
+"clerk": patch
+---
+
+Fix release binaries reporting a dev version when compiled with Bun 1.4.0: the `CLI_VERSION` define check moved out of the version macro (Bun 1.4.0 no longer substitutes `--define` globals during macro execution) into module code, so injected release versions are honored again.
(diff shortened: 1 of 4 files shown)


wyattjoh (approved):

rafa-thayto:
Superseded by #435 — the same Bun 1.4 fix (CLI_VERSION define check moved out of the macro into version.ts module scope, macro reduced to the checkout fallback, versioning rule doc updated, patch changeset) landed on main in 432ea9f. Closing to avoid a duplicate changelog entry and rename churn.

PR #435, opened 2026-08-21
fix(cli): await the next-steps outro so post-command output prints below it

rafa-thayto:
## Summary

Fixes the garbled "Update available" output reported in [Slack](https://clerkinc.slack.com/archives/C0944EQCUKY/p1787174889556909): after `clerk switch-env`, the update notice printed **on top of** the "Next steps" block (`Update available: 3.0.1 → 3.1.0henticate for this environment`), with missing spaces, doubled backticks, and a stray duplicate "Next steps" header.

**Root cause:** `outro(steps)` in `lib/spinner.ts` is async — it awaits `animateHeader`, which parks the cursor on the header line for ~450ms of shine animation before restoring it. Several commands called `outro(...)` without `await` (a floating promise), so the command's promise resolved at the animation's first frame. Commander's `postAction` hook then ran `maybeNotifyUpdate`, whose lines printed at the parked cursor position — overwriting the step lines mid-block, while the animation's remaining frames repainted the header further down.

**Fix:** `await` every `outro(...)` call site (47 sites across 18 command files; `deploy` and `webhooks token` already awaited correctly). Output printed after a command now always lands below the finished block — the ordering `.claude/rules/interrupts.md` already documents as the post-command tail.

## Test plan

- New regression test `switch-env/update-notice-race.test.ts` runs the **real** spinner/gradient on a forced-interactive TTY, prints the update notice the instant `switchEnv()` resolves (exactly what `postAction` does), and asserts it lands after the animation's cursor-restore escape. Red before the fix (notice at stream index 606, cursor restored at 4377), green after. The assertion is enforced by promise sequencing, not wall-clock timing.
- `bun run format:check` / `lint` / `typecheck` clean; full unit suite 2654 pass / 0 fail.
- Changeset: `clerk` patch.


thiskevinwang (approved):
LGTM! Looks like repeated applying of missed `await` to `outro()`

Consider a eslint or oxlint rule to catch this at compiler level

packages/cli-core/src/commands/switch-env/update-notice-race.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  test("output printed after the command resolves lands below the finished animation", async () => {
+    await switchEnv("staging");
+    // The postAction hook prints the update notice as soon as the action's
+    // promise resolves — reproduce that exact ordering.
+    log.warn(NOTICE);
+    // Drain any animation frames still in flight so the buffer is complete
+    // (and so a buggy run doesn't leak writes into the next test).
+    await Bun.sleep(700);
```
wyattjoh: nit: With the fix applied, nothing is in flight when `switchEnv()` resolves, so this is 700ms of unconditional dead time on every green run — it makes this the slowest file in the suite (~1.2s).

I checked what happens without it: the test still passes on the fixed code (~490ms, three consecutive runs), and it still goes red on the un-awaited code — it just fails on `expect(cursorRestoredAt).toBeGreaterThanOrEqual(0)` with `-1` instead of on the ordering assertion. And since this is the only test in the file, the "leak into the next test" half of the comment doesn't apply.

Could drop it, or swap it for a short bounded poll until `CURSOR_SHOW` shows up in the buffer so the red case keeps its more precise failure message without the fixed cost.

###### Sent from Claude

wyattjoh (commented):
Checked that the regression test genuinely goes red on the un-awaited code and green on the fix, and that the Bun 1.4 define change still resolves correctly on Bun 1.3.14 across all three build shapes (release define, no define, dev-flavored define). Two nits below; neither blocks.

###### Sent from Claude

wyattjoh (approved):

PR #441, opened 2026-08-22
feat(telemetry): give every failure an error code and report where a run stopped

djgould:
`clerk init` fails on ~21% of runs and telemetry can't say why or where. This makes both visible:

- Every `CliError` now carries an error code (19 sites were uncoded and collapsed into a generic `cli_error`).
- New `stage` field reports the furthest step a run reached — on success, error, and abort. `clerk init` is instrumented (`flags → detect → … → done`).
- `clerk update` now distinguishes a bad `--channel` from an unreachable registry.

First-run disclosure notice updated for the new field.

### DIFF
--- a/packages/cli-core/src/commands/update/index.ts
+++ b/packages/cli-core/src/commands/update/index.ts
@@ -2,6 +2,7 @@ import type { Program } from "../../cli-program.ts";
 import { isAgent, isHuman } from "../../mode.ts";
 import { green, cyan, yellow, dim } from "../../lib/color.ts";
 import { CliError, ERROR_CODE } from "../../lib/errors.ts";
+import { isCancelled } from "../../lib/signals.ts";
 import {
   asdfPluginFromPath,
   asdfReshim,
@@ -272,9 +273,10 @@ export async function update(options: UpdateOptions): Promise<void> {
     withSpinner("Checking for updates...", async () => fetchLatestVersion(channel)).catch(
       (error: unknown) => {
         // A registry that answered — badly, or without the requested channel —
-        // already carries its own code. Only transport and timeout failures are
-        // genuinely "unreachable", and retrying is only right for those.
-        if (error instanceof CliError) throw error;
+        // already carries its own code, and a Ctrl-C is the user's decision,
+        // not the network's. Only transport and timeout failures are genuinely
+        // "unreachable", and retrying is only right for those.
+        if (error instanceof CliError || isCancelled(error)) throw error;
         throw new CliError("Could not reach npm registry. Check your network connection.", {
           code: ERROR_CODE.REGISTRY_UNREACHABLE,
         });
--- a/packages/cli-core/src/commands/update/index.ts
+++ b/packages/cli-core/src/commands/update/index.ts
@@ -1,7 +1,7 @@
 import type { Program } from "../../cli-program.ts";
 import { isAgent, isHuman } from "../../mode.ts";
 import { green, cyan, yellow, dim } from "../../lib/color.ts";
-import { CliError } from "../../lib/errors.ts";
+import { CliError, ERROR_CODE } from "../../lib/errors.ts";
 import {
   asdfPluginFromPath,
   asdfReshim,
@@ -134,12 +134,18 @@ async function runGlobalInstall(
     const stderr = result.stderr.toString();
     const hint = globalInstallCommand(installer, packageSpec);
     if (stderr.includes("EACCES") || stderr.includes("permission denied")) {
-      throw new CliError(`Permission denied. Try: sudo ${hint}`);
+      throw new CliError(`Permission denied. Try: sudo ${hint}`, {
+        code: ERROR_CODE.UPDATE_PERMISSION_DENIED,
+      });
     }
     if (result.exitCode === 127 || stderr.includes("not found")) {
-      throw new CliError(`${installer} not found on PATH.`);
+      throw new CliError(`${installer} not found on PATH.`, {
+        code: ERROR_CODE.INSTALLER_NOT_FOUND,
+      });
     }
-    throw new CliError(`Update failed: ${stderr.trim() || "unknown error"}`);
+    throw new CliError(`Update failed: ${stderr.trim() || "unknown error"}`, {
+      code: ERROR_CODE.UPDATE_FAILED,
+    });
   }
 
   // Homebrew installs whatever version its tap currently publishes, ignoring
@@ -154,6 +160,7 @@ async function runGlobalInstall(
         `Homebrew tap is stale: installed ${installed}, expected ${targetVersion}. ` +
           `Update via another installer (e.g. \`npm install -g ${packageSpec}\`) ` +
           `or wait for the tap to catch up.`,
+        { code: ERROR_CODE.UPDATE_FAILED },
       );
     }
   }
@@ -262,9 +269,17 @@ export async function update(options: UpdateOptions): Promise<void> {
   if (isHuman()) intro("Checking for updates");
 
   const [latest, installDirs] = await Promise.all([
-    withSpinner("Checking for updates...", async () => fetchLatestVersion(channel)).catch(() => {
-      throw new CliError("Could not reach npm registry. Check your network connection.");
-    }),
+    withSpinner("Checking for updates...", async () => fetchLatestVersion(channel)).catch(
+      (error: unknown) => {
+        // A registry that answered — badly, or without the requested channel —
+        // already carries its own code. Only transport and timeout failures are
+        // genuinely "unreachable", and retrying is only right for those.
+        if (error instanceof CliError) throw error;
+        throw new CliError("Could not reach npm registry. Check your network connection.", {
+          code: ERROR_CODE.REGISTRY_UNREACHABLE,
+        });
+      },
+    ),
     getInstallerPackageDirs(),
   ]);
 
--- a/packages/cli-core/src/lib/telemetry.ts
+++ b/packages/cli-core/src/lib/telemetry.ts
@@ -47,18 +47,21 @@ export type TelemetryResult = {
  * bare string so a typo or a rename that misses a call site fails to compile
  * instead of silently splitting the funnel into two buckets in the warehouse,
  * and so no interpolated value (a path, a project name) can reach the payload.
+ *
+ * Declared in execution order: this is the funnel, so a new stage goes where it
+ * runs, not at the end. `already_set_up` is a terminal branch off `scaffold`.
  */
 export type TelemetryStage =
   | "flags"
   | "detect"
+  | "bootstrap"
   | "strategy"
   | "link"
-  | "bootstrap"
   | "install"
   | "scaffold"
+  | "already_set_up"
   | "keys"
   | "skills"
-  | "already_set_up"
   | "done";
 
 /** Structural slice of Commander's Command — avoids its generic types. */
@@ -73,7 +76,7 @@ type TelemetryContext = {
   command: string;
   flags: string;
   startedAt: number;
-  /** Furthest stage reached — see setTelemetryStage. */
+  /** Last stage set — see setTelemetryStage. */
   stage: TelemetryStage | null;
 };
 
--- a/packages/cli-core/src/lib/telemetry.ts
+++ b/packages/cli-core/src/lib/telemetry.ts
@@ -42,6 +42,25 @@ export type TelemetryResult = {
   errorCode?: string;
 };
 
+/**
+ * Closed set of drop-off points a command can report. A union rather than a
+ * bare string so a typo or a rename that misses a call site fails to compile
+ * instead of silently splitting the funnel into two buckets in the warehouse,
+ * and so no interpolated value (a path, a project name) can reach the payload.
+ */
+export type TelemetryStage =
+  | "flags"
+  | "detect"
+  | "strategy"
+  | "link"
+  | "bootstrap"
+  | "install"
+  | "scaffold"
+  | "keys"
+  | "skills"
+  | "already_set_up"
+  | "done";
+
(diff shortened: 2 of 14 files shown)


packages/cli-core/src/lib/update-check.ts
```diff
@@ -106,11 +107,23 @@ export async function fetchLatestVersion(distTag: string, timeoutMs = 1500): Pro
...
+    if (!isNpmDistTagsResponse(data)) {
+      throw new CliError("Unexpected response shape from the npm registry.", {
+        code: ERROR_CODE.UPDATE_FAILED,
+      });
+    }
     const version = data["dist-tags"][distTag];
-    if (!version) throw new Error(`dist-tag "${distTag}" not found`);
+    if (!version) {
+      throw new CliError(`Release channel "${distTag}" does not exist.`, {
+        code: ERROR_CODE.USAGE_ERROR,
```
wyattjoh: This gets the code but not the exit code, so the "bad `--channel` reports a usage error" goal only half-lands.

`new CliError(..., { code: ERROR_CODE.USAGE_ERROR })` omits `exitCode`, which defaults to [`EXIT_CODE.GENERAL`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/errors.ts#L9) (1) rather than [`EXIT_CODE.USAGE`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/errors.ts#L11) (2). Confirmed by running it:

```
code: usage_error | exitCode: 1 | EXIT_CODE.USAGE: 2
```

[`errors.ts:164`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/errors.ts#L164) documents "For usage/validation errors, **prefer `throwUsageError`**", and every other `USAGE_ERROR` site routes through [`throwUsageError`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/errors.ts#L485) — [`input-json.ts:71`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/input-json.ts#L71), [`input-json.ts:115`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/input-json.ts#L115), [`api/index.ts:37`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/api/index.ts#L37). This is the only one that constructs `CliError` directly.

The consequence is exactly the distinction this change set out to create: a script or agent branching on exit code 2 to tell bad input from a transport failure sees 1 for both, and can't decide whether retrying is right.

The test at [`update-check.test.ts:233`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/update-check.test.ts#L233) asserts only `code`, which is why this passes CI.

(No impact on the background notifier — [`maybeNotifyUpdate`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/update-check.ts#L154) catches everything into `log.debug`.)

<details>
<summary>Suggested fix</summary>

Minimal version, applicable inline:

```suggestion
        code: ERROR_CODE.USAGE_ERROR,
        exitCode: EXIT_CODE.USAGE,
```

That needs `EXIT_CODE` added to the import at [`update-check.ts:11`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/update-check.ts#L11):

```ts
import { CliError, ERROR_CODE, EXIT_CODE } from "./errors.ts";
```

The more idiomatic version, matching the guidance in `errors.ts` and the other three call sites — replaces lines 122–126 and drops the need for `CliError` here entirely:

```ts
    if (!version) {
      throwUsageError(`Release channel "${distTag}" does not exist.`);
    }
```

Either way, tighten the test so the gap can't reopen:

```ts
    await expect(fetchLatestVersion("typo")).rejects.toMatchObject({
      code: ERROR_CODE.USAGE_ERROR,
      exitCode: EXIT_CODE.USAGE,
    });
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/commands/update/index.ts
```diff
@@ -262,9 +269,17 @@ export async function update(options: UpdateOptions): Promise<void> {
...
   const [latest, installDirs] = await Promise.all([
-    withSpinner("Checking for updates...", async () => fetchLatestVersion(channel)).catch(() => {
-      throw new CliError("Could not reach npm registry. Check your network connection.");
-    }),
+    withSpinner("Checking for updates...", async () => fetchLatestVersion(channel)).catch(
+      (error: unknown) => {
+        // A registry that answered — badly, or without the requested channel —
+        // already carries its own code. Only transport and timeout failures are
+        // genuinely "unreachable", and retrying is only right for those.
+        if (error instanceof CliError) throw error;
```
wyattjoh: The comment above this reasons that only transport and timeout failures are genuinely "unreachable". An interrupt is neither, but it lands in this same branch.

`loggedFetch` composes the caller's signal with [`interruptSignal()`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/signals.ts#L60), so Ctrl-C aborts the in-flight fetch and rejects with `AbortError` — not a `CliError` — and gets converted to `REGISTRY_UNREACHABLE`.

There's already a shared predicate for this, [`isCancelled()`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/signals.ts#L103), whose own docstring says it exists so "the command wrappers agree with `withSpinner` on what counts as a cancellation" and warns that "an interrupt aborts the in-flight request, which rejects with `AbortError`, not `UserAbortError`."

The impact is bounded — the SIGINT handler races to flush an `abort` event and re-raise, so it often wins. But when the command path wins, a user-initiated cancel reports `registry_unreachable` at exit 1 and can print "Check your network connection." This is pre-existing (the old blanket `.catch(() => ...)` swallowed everything), so the change is an improvement; it just stops one case short of the reasoning in its own comment.

<details>
<summary>Suggested fix</summary>

```suggestion
        if (error instanceof CliError || isCancelled(error)) throw error;
```

Plus the import alongside the existing one at [`update/index.ts:4`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/update/index.ts#L4):

```ts
import { isCancelled } from "../../lib/signals.ts";
```

The comment just above is worth extending too, since it currently enumerates the cases and now misses one:

```ts
        // A registry that answered — badly, or without the requested channel —
        // already carries its own code, and a Ctrl-C is the user's decision, not
        // the network's. Only transport and timeout failures are genuinely
        // "unreachable", and retrying is only right for those.
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -42,6 +42,25 @@ export type TelemetryResult = {
...
+ * bare string so a typo or a rename that misses a call site fails to compile
+ * instead of silently splitting the funnel into two buckets in the warehouse,
+ * and so no interpolated value (a path, a project name) can reach the payload.
+ */
+export type TelemetryStage =
+  | "flags"
+  | "detect"
+  | "strategy"
+  | "link"
+  | "bootstrap"
```
wyattjoh: The union is documented as a drop-off funnel, and its declaration order is the only ordering signal that ships to whoever builds that funnel. Two entries are out of order relative to how `init` actually runs.

`bootstrap` ([`init/index.ts:293`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/init/index.ts#L293) and [`:310`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/init/index.ts#L310)) fires inside the `resolved = ...` call at [`:107`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/init/index.ts#L107), so it runs before `strategy` at [`:131`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/init/index.ts#L131). And `already_set_up` ([`:170`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/init/index.ts#L170)) fires before `keys` ([`:179`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/commands/init/index.ts#L179)).

Actual execution order:

```
flags → detect → bootstrap → strategy → link → install → scaffold → [already_set_up] → keys → skills → done
```

Declared order:

```
flags → detect → strategy → link → bootstrap → install → scaffold → keys → skills → already_set_up → done
```

A funnel chart ordered by the enum shows two backwards transitions, which is the legibility problem this change exists to fix.

<details>
<summary>Suggested fix</summary>

Reorder to match execution and say that the order carries meaning — replaces lines 45–62:

```ts
/**
 * Closed set of drop-off points a command can report. A union rather than a
 * bare string so a typo or a rename that misses a call site fails to compile
 * instead of silently splitting the funnel into two buckets in the warehouse,
 * and so no interpolated value (a path, a project name) can reach the payload.
 *
 * Declared in execution order: this is the funnel, so a new stage goes where it
 * runs, not at the end. `already_set_up` is a terminal branch off `scaffold`.
 */
export type TelemetryStage =
  | "flags"
  | "detect"
  | "bootstrap"
  | "strategy"
  | "link"
  | "install"
  | "scaffold"
  | "already_set_up"
  | "keys"
  | "skills"
  | "done";
```

If the order genuinely isn't meant to be significant, the one-line version works just as well — say so explicitly so nobody downstream infers a funnel from it:

```ts
 * Declaration order is not significant; the warehouse defines the funnel order.
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/lib/telemetry.ts
```diff
@@ -54,6 +73,8 @@ type TelemetryContext = {
   command: string;
   flags: string;
   startedAt: number;
+  /** Furthest stage reached — see setTelemetryStage. */
```
wyattjoh: [`setTelemetryStage`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/telemetry.ts#L190) is `context.stage = stage` — last write wins. That's equivalent to "furthest" only while every call site is monotonic, which nothing enforces.

`bootstrap` already has two call sites, and a stage set inside a shared helper or a retry loop would silently turn the value into "last" rather than "furthest" with no compile error to catch it.

<details>
<summary>Suggested fix</summary>

Cheapest option — match the wording the function's own docstring already gets right at [`telemetry.ts:184`](https://github.com/clerk/cli/blob/e87753f753304a0667e500b4a40db035a2c119b0/packages/cli-core/src/lib/telemetry.ts#L184) ("The last stage set is the one sent"):

```suggestion
  /** Last stage set — see setTelemetryStage. */
```

Or keep the stronger wording and actually enforce it, which also makes the union's order load-bearing in a way the compiler checks:

```ts
const STAGE_ORDER: readonly TelemetryStage[] = [
  "flags", "detect", "bootstrap", "strategy", "link",
  "install", "scaffold", "already_set_up", "keys", "skills", "done",
];

export function setTelemetryStage(stage: TelemetryStage): void {
  if (!context) return;
  const seen = context.stage;
  if (seen && STAGE_ORDER.indexOf(stage) <= STAGE_ORDER.indexOf(seen)) return;
  context.stage = stage;
}
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/lib/update-check.ts
djgould: Fixed in 61a02d9a — routed through `throwUsageError` so it exits 2, and the test asserts `exitCode` now.

packages/cli-core/src/commands/update/index.ts
djgould: Fixed in 61a02d9a — `isCancelled(error)` rethrows alongside `CliError`, comment extended as suggested.

packages/cli-core/src/lib/telemetry.ts
djgould: Reordered to execution order in 61a02d9a, with the doc note that order carries meaning. (#442 stacks login stages as a second per-command group.)

packages/cli-core/src/lib/telemetry.ts
djgould: Took the docstring fix in 61a02d9a — "Last stage set". Skipped STAGE_ORDER enforcement: #442 adds a second command's stages to the union, and cross-command ordering shouldn't be load-bearing.

packages/cli-core/src/lib/update-check.ts
```diff
@@ -106,11 +107,23 @@ export async function fetchLatestVersion(distTag: string, timeoutMs = 1500): Pro
...
+      });
+    }
```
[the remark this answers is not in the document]
djgould: Fixed in 61a02d9a — missing tag → usage error, present-but-non-string/empty value → `update_failed`, regression test added.

wyattjoh (approved):

PR #442, opened 2026-08-22
fix(auth): give each login failure its own error code and stage

djgould:
Stacked on #441 — retarget to `main` once it merges.

Every failure in the `auth login` browser flow reported as one generic `unexpected_error`. Each now has its own code — `auth_timeout`, `oauth_provider_error`, `oauth_state_mismatch`, `oauth_no_code`, `callback_bind_failed` — and login records which step it reached (`session_check → awaiting_callback → token_exchange → store → done`), so a walked-away browser wait is distinguishable from real breakage.

### DIFF
--- a/.changeset/auth-login-error-codes.md
+++ b/.changeset/auth-login-error-codes.md
@@ -1,5 +1,5 @@
 ---
-"clerk": minor
+"clerk": patch
 ---
 
 Report why `clerk auth login` failed instead of collapsing every failure into one error.
--- a/.changeset/auth-login-error-codes.md
+++ b/.changeset/auth-login-error-codes.md
@@ -1,0 +1,6 @@
+---
+"clerk": minor
+---
+
+Report why `clerk auth login` failed instead of collapsing every failure into one error.
+The browser sign-in wait, OAuth provider errors, state mismatches, missing authorization codes, and callback bind failures now carry distinct error codes, and login records which step of the flow it reached.
--- a/packages/cli-core/src/commands/auth/login.test.ts
+++ b/packages/cli-core/src/commands/auth/login.test.ts
@@ -97,6 +97,7 @@ mock.module("../../lib/autoclaim.ts", () => ({
 }));
 
 const { setLogLevel } = await import("../../lib/log.ts");
+const telemetryMod = await import("../../lib/telemetry.ts");
 const { login } = await import("./login.ts");
 
 describe("login", () => {
@@ -608,5 +609,56 @@ describe("login", () => {
     await runLogin({ showNextSteps: false });
 
     expect(captured.err).not.toContain("Next steps:");
+  });
+
+  describe("telemetry stages", () => {
+    function trackStages() {
+      const stage = spyOn(telemetryMod, "setTelemetryStage");
+      return {
+        calls: () => stage.mock.calls.map((call) => call[0]),
+        restore: () => stage.mockRestore(),
+      };
+    }
+
+    test("a completed login reports the terminal stage", async () => {
+      mockGetValidToken.mockResolvedValue(null);
+      mockOAuthSuccess();
+      const stages = trackStages();
+
+      await runLogin();
+
+      expect(stages.calls().at(-1)).toBe("done");
+      stages.restore();
+    });
+
+    // The browser wait is where the flow is known to lose people; a run that
+    // stops there must be attributable to that step and no other.
+    test("a login abandoned during the browser wait stops at awaiting_callback", async () => {
+      mockGetValidToken.mockResolvedValue(null);
+      mockBunSpawn();
+      mockStartAuthServer.mockReturnValue({
+        port: 54321,
+        waitForCallback: mock().mockRejectedValue(new Error("Authentication timed out.")),
+        stop: mock(),
+      });
+      const stages = trackStages();
+
+      await expect(runLogin()).rejects.toThrow("Authentication timed out");
+
+      expect(stages.calls().at(-1)).toBe("awaiting_callback");
+      stages.restore();
+    });
+
+    test("a failure exchanging the code stops at token_exchange", async () => {
+      mockGetValidToken.mockResolvedValue(null);
+      mockOAuthSuccess();
+      mockExchangeCodeForToken.mockRejectedValue(new Error("exchange failed"));
+      const stages = trackStages();
+
+      await expect(runLogin()).rejects.toThrow("exchange failed");
+
+      expect(stages.calls().at(-1)).toBe("token_exchange");
+      stages.restore();
+    });
   });
 });
--- a/packages/cli-core/src/commands/auth/login.test.ts
+++ b/packages/cli-core/src/commands/auth/login.test.ts
@@ -612,12 +612,20 @@ describe("login", () => {
   });
 
   describe("telemetry stages", () => {
+    let stageSpy: ReturnType<typeof spyOn> | undefined;
+    let callerStageSpy: ReturnType<typeof spyOn> | undefined;
+
+    afterEach(() => {
+      stageSpy?.mockRestore();
+      callerStageSpy?.mockRestore();
+      stageSpy = undefined;
+      callerStageSpy = undefined;
+    });
+
     function trackStages() {
-      const stage = spyOn(telemetryMod, "setTelemetryStage");
-      return {
-        calls: () => stage.mock.calls.map((call) => call[0]),
-        restore: () => stage.mockRestore(),
-      };
+      const spy = spyOn(telemetryMod, "setTelemetryStage");
+      stageSpy = spy;
+      return { calls: () => spy.mock.calls.map((call) => call[0]) };
     }
 
     test("a completed login reports the terminal stage", async () => {
@@ -628,7 +636,6 @@ describe("login", () => {
       await runLogin();
 
       expect(stages.calls().at(-1)).toBe("done");
-      stages.restore();
     });
 
     // The browser wait is where the flow is known to lose people; a run that
@@ -646,7 +653,30 @@ describe("login", () => {
       await expect(runLogin()).rejects.toThrow("Authentication timed out");
 
       expect(stages.calls().at(-1)).toBe("awaiting_callback");
-      stages.restore();
+    });
+
+    // `init` and `link` invoke login mid-flow; a clean return must hand the
+    // caller's stage back instead of leaving the outer command at "done".
+    test("a nested login hands the caller's stage back on success", async () => {
+      mockGetValidToken.mockResolvedValue(null);
+      mockOAuthSuccess();
+      callerStageSpy = spyOn(telemetryMod, "currentTelemetryStage").mockReturnValue("link");
+      const stages = trackStages();
+
+      await runLogin({ showNextSteps: false });
+
+      expect(stages.calls().at(-1)).toBe("link");
+    });
+
+    test("a failure persisting the token stops at store", async () => {
+      mockGetValidToken.mockResolvedValue(null);
+      mockOAuthSuccess();
+      mockStoreToken.mockRejectedValue(new Error("keychain unavailable"));
+      const stages = trackStages();
+
+      await expect(runLogin()).rejects.toThrow("keychain unavailable");
+
+      expect(stages.calls().at(-1)).toBe("store");
     });
 
     test("a failure exchanging the code stops at token_exchange", async () => {
@@ -658,7 +688,6 @@ describe("login", () => {
       await expect(runLogin()).rejects.toThrow("exchange failed");
 
       expect(stages.calls().at(-1)).toBe("token_exchange");
-      stages.restore();
     });
   });
 });
--- a/packages/cli-core/src/commands/auth/login.ts
+++ b/packages/cli-core/src/commands/auth/login.ts
@@ -25,7 +25,7 @@ import { attemptAutoclaim, type AutoclaimResult } from "../../lib/autoclaim.ts";
 import { openBrowser } from "../../lib/open.ts";
 import { cyan, dim } from "../../lib/color.ts";
 import { log } from "../../lib/log.ts";
-import { setTelemetryStage } from "../../lib/telemetry.js";
+import { currentTelemetryStage, setTelemetryStage } from "../../lib/telemetry.ts";
 import { ensureFirstApplication } from "../../lib/first-application.ts";
 
 interface LoginOptions {
@@ -134,6 +134,18 @@ async function performOAuthFlow(): Promise<OAuthFlowResult> {
 }
 
 export async function login(options: LoginOptions = {}): Promise<UserInfo> {
+  // `init` and `link` call this mid-flow and share the one process-global
+  // telemetry stage. Login's own markers are worth having while it runs, but
+  // on a clean return the caller's stage comes back so the rest of *their*
+  // work isn't reported as `done`. On a throw the login stage stands: that is
+  // genuinely where the run stopped.
+  const callerStage = currentTelemetryStage();
+  const userInfo = await runLogin(options);
+  if (callerStage) setTelemetryStage(callerStage);
+  return userInfo;
+}
+
+async function runLogin(options: LoginOptions = {}): Promise<UserInfo> {
   const { showNextSteps = true, yes } = options;
   intro("Signing in");
   setTelemetryStage("session_check");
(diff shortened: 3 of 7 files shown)


packages/cli-core/src/commands/auth/login.ts
```diff
@@ -183,6 +189,7 @@ export async function login(options: LoginOptions = {}): Promise<UserInfo> {
   await withSpinner("Setting up your default application...", async () => ensureFirstApplication());
 
   bar();
+  setTelemetryStage("done");
```
wyattjoh, line 205: `login()` isn't only the `clerk auth login` command body — `init` calls it from `resolveAuthLabel` ([`init/index.ts:460`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/init/index.ts#L460)), `link` from `ensureAuth` ([`link/index.ts:119`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/link/index.ts#L119)), and `doctor` from its context fixer. The telemetry context is process-global ([`telemetry.ts:87`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/telemetry.ts#L87)), created once per run by the `preAction` hook against the top-level command, so a stage set in here lands on *that* command's event.

The concrete failure: `clerk link` in a logged-out directory calls `login()`, which reaches this line and sets `stage: "done"`. Control returns to `ensureAuth`, and `link` then does the work it actually exists to do — `resolveProfile`, app selection, writing the profile. If any of that throws, the event is `command: "link", outcome: "error", stage: "done"`. `link` sets no stages of its own, so this is the only stage it will ever report, and it reports it exclusively on runs that failed *after* completing auth.

`init` has the same shape: [`setTelemetryStage("link")`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/init/index.ts#L154), then `authenticateAndLink` → `login()` → `"done"`, then the rest of `authenticateAndLink` plus `detectAndInstall` all run before init's next marker ([`"keys"` at index.ts:179](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/init/index.ts#L179)). Every failure in that window is attributed to `"done"`.

That inverts what the stage field is for. The same hazard applies to the whole flat [`TelemetryStage`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/telemetry.ts#L51-L69) union — a future command adding a `"link"` or `"keys"` stage will silently collide with init's.

<details>
<summary>Suggested fix — restore the caller's stage on a clean return</summary>

Restoring only on a clean return keeps the useful half: a run that dies inside the browser wait still reports `awaiting_callback`, correctly attributed to whichever command was running. It's the successful return that has to hand the stage back.

In `lib/telemetry.ts`:

```ts
/** Read the stage a caller had set, so a nested flow can hand it back. */
export function currentTelemetryStage(): TelemetryStage | null {
  return context?.stage ?? null;
}
```

In `commands/auth/login.ts` — rename the existing body to `runLogin` and wrap it:

```ts
export async function login(options: LoginOptions = {}): Promise<UserInfo> {
  // `init`, `link`, and `doctor` call this mid-flow and share the one
  // process-global stage. Its own markers are worth having while it runs, but
  // on a clean return the caller's stage comes back so the rest of *their*
  // work isn't reported as `done`. On a throw the login stage stands: that is
  // genuinely where the run stopped.
  const callerStage = currentTelemetryStage();
  const userInfo = await runLogin(options);
  if (callerStage) setTelemetryStage(callerStage);
  return userInfo;
}
```

Gating the stage calls on login being the invoked command works too, and is more explicit — but it needs a signal login doesn't currently have (`showNextSteps: false` correlates with nested calls today, which is a coincidence worth not depending on).

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -121,6 +124,7 @@ async function performOAuthFlow(): Promise<OAuthFlowResult> {
     }),
   );
 
+  setTelemetryStage("store");
```
wyattjoh, line 127: `"store"` is the last stage set before `performOAuthFlow` returns, so it covers `storeToken`, `fetchUserInfo`, `setAuth`, and then back in `login()`: `revokeToken` for the superseded session and [`ensureFirstApplication`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/auth/login.ts#L189).

`ensureFirstApplication` is the one that matters — it's a PLAPI round trip that creates an application, and it's the slowest and most failure-prone step in the tail of this flow. A user whose login dies there produces `stage: "store"`, which reads as a credential-store problem and points debugging at the keychain instead of the API.

<details>
<summary>Suggested fix — give the app-creation step its own marker</summary>

In [`lib/telemetry.ts`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/telemetry.ts#L63-L67):

```diff
   // `clerk auth login`
   | "session_check"
   | "awaiting_callback"
   | "token_exchange"
   | "store"
+  | "first_application"
```

In [`commands/auth/login.ts`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/auth/login.ts#L187-L189):

```diff
   // Best-effort: ensure the user has at least one application so downstream
   // commands (clerk link, clerk init) have something to operate on.
+  setTelemetryStage("first_application");
   await withSpinner("Setting up your default application...", async () => ensureFirstApplication());
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/lib/auth-server.test.ts
```diff
@@ -31,6 +33,50 @@ describe("auth-server", () => {
...
+    const errorPromise = server.waitForCallback().catch((e: unknown) => e);
+
+    expect(fire).toBeDefined();
+    fire?.();
+
+    const error = await errorPromise;
+    expect(error).toMatchObject({ code: ERROR_CODE.AUTH_TIMEOUT });
+    expect((error as Error).message).toContain("timed out");
+
+    timeoutSpy.mockRestore();
```
wyattjoh: `timeoutSpy.mockRestore()` and `server.stop()` run after the assertions, so any failing `expect` above them skips both. The consequences are asymmetric with the usual leaked-mock nit: a live `spyOn(globalThis, "setTimeout")` stays installed for the rest of the file, and every later test that calls `startAuthServer` gets its deadline swallowed and its `fire` callback captured into a stale closure. One real assertion failure here cascades into a set of confusing unrelated ones.

The file already has the right pattern for this — [`serveSpy` and `clearTimeoutSpy`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/auth-server.test.ts#L8-L17) are `let`-declared at describe scope and restored in `afterEach`.

<details>
<summary>Suggested fix — move both into the existing afterEach</summary>

```diff
 describe("auth-server", () => {
   let serveSpy: ReturnType<typeof spyOn> | undefined;
   let clearTimeoutSpy: ReturnType<typeof spyOn> | undefined;
+  let timeoutSpy: ReturnType<typeof spyOn> | undefined;
+  let openServer: { stop: () => void } | undefined;
   useCaptureLog();

   afterEach(() => {
     serveSpy?.mockRestore();
     clearTimeoutSpy?.mockRestore();
+    timeoutSpy?.mockRestore();
+    openServer?.stop();
     serveSpy = undefined;
     clearTimeoutSpy = undefined;
+    timeoutSpy = undefined;
+    openServer = undefined;
   });
```

Then in the timeout test, assign instead of declaring, and drop the trailing cleanup:

```diff
-    const timeoutSpy = spyOn(globalThis, "setTimeout").mockImplementation(((
+    timeoutSpy = spyOn(globalThis, "setTimeout").mockImplementation(((
       cb: () => void,
       ms?: number,
       ...rest: unknown[]
     ) => {
@@
     const server = startAuthServer("test-state");
+    openServer = server;
     const errorPromise = server.waitForCallback().catch((e: unknown) => e);
@@
     expect((error as Error).message).toContain("timed out");
-
-    timeoutSpy.mockRestore();
-    server.stop();
   });
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/commands/auth/login.test.ts
```diff
@@ -609,4 +610,55 @@ describe("login", () => {
...
+
+    test("a completed login reports the terminal stage", async () => {
+      mockGetValidToken.mockResolvedValue(null);
+      mockOAuthSuccess();
+      const stages = trackStages();
+
+      await runLogin();
+
+      expect(stages.calls().at(-1)).toBe("done");
+      stages.restore();
```
wyattjoh: Same shape as the `auth-server.test.ts` teardown, lower blast radius: `stages.restore()` is unreachable if `expect(stages.calls().at(-1))` fails, leaving the module spy installed for the two tests that follow — each of which then calls `spyOn` on an already-spied property.

While you're here: the three tests cover `done`, `awaiting_callback`, and `token_exchange`, but nothing pins `session_check` or `store`, and nothing covers `login()` invoked as a subroutine — which is where the stage bookkeeping actually goes wrong.

<details>
<summary>Suggested fix — register the restore in a teardown hook</summary>

```diff
   describe("telemetry stages", () => {
+    let stageSpy: ReturnType<typeof spyOn> | undefined;
+
+    afterEach(() => {
+      stageSpy?.mockRestore();
+      stageSpy = undefined;
+    });
+
     function trackStages() {
-      const stage = spyOn(telemetryMod, "setTelemetryStage");
-      return {
-        calls: () => stage.mock.calls.map((call) => call[0]),
-        restore: () => stage.mockRestore(),
-      };
+      const spy = spyOn(telemetryMod, "setTelemetryStage");
+      stageSpy = spy;
+      return { calls: () => spy.mock.calls.map((call) => call[0]) };
     }
```

Each test then drops its trailing `stages.restore()` line.

A fourth case worth adding, which is what LOG-001 is about:

```ts
test("a nested login leaves the caller's stage in place", async () => {
  mockGetValidToken.mockResolvedValue(null);
  mockOAuthSuccess();
  telemetryMod.setTelemetryStage("link");
  const stages = trackStages();

  await runLogin({ showNextSteps: false });

  expect(stages.calls().at(-1)).toBe("link");
});
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/lib/auth-server.ts
```diff
@@ -266,7 +280,14 @@ export function startAuthServer(expectedState: string): AuthServerResult {
       target: "127.0.0.1:0",
       label: CALLBACK_PATH,
     });
-    throw error;
+    // A sandbox or firewall that forbids binding loopback fails every login on
+    // the machine; it is a distinct condition from anything the user did.
+    throw new CliError(
```
wyattjoh, line 288: `errorMessage(error)` preserves the message text, but the thrown `CliError` carries no reference to the original — [`CliErrorOptions`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/errors.ts#L143-L155) has no `cause`, and [`observeHostCapabilityFailure`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/host-execution.ts#L205-L216) only logs when `isAgent()` and the failure matches its sandbox heuristic. In a normal human-mode run the errno and stack are gone before anyone can look at them, and `EACCES` (sandbox/firewall) versus `EADDRINUSE` versus something else is exactly the distinction the new `callback_bind_failed` code exists to help diagnose.

<details>
<summary>Suggested fix — keep the original recoverable under --verbose</summary>

Every other branch in this file already debug-logs before rejecting; this one is the outlier.

```suggestion
    log.debug(`auth-server: bind failed — ${error instanceof Error ? (error.stack ?? error.message) : String(error)}`);
    throw new CliError(
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

.changeset/auth-login-error-codes.md
```diff
@@ -0,0 +1,6 @@
+---
+"clerk": minor
```
wyattjoh: The commit is `fix(auth): ...` and the change is a diagnostics fix rather than new CLI surface. The repo's [bump table](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/.claude/skills/changesets/references/policy.md#L73-L96) maps `fix:` to `patch` and reserves `minor` for new user-facing features, commands, or flags, with an explicit default-to-`patch` rule when the split is unclear. The parent PR's `minor` is correct for its own `feat(telemetry):` commit, but that doesn't carry over.

<details>
<summary>Suggested fix</summary>

```suggestion
"clerk": patch
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -25,6 +25,7 @@ import { attemptAutoclaim, type AutoclaimResult } from "../../lib/autoclaim.ts";
 import { openBrowser } from "../../lib/open.ts";
 import { cyan, dim } from "../../lib/color.ts";
 import { log } from "../../lib/log.ts";
+import { setTelemetryStage } from "../../lib/telemetry.js";
```
wyattjoh: Every other import in this file — including [`../../lib/errors.ts`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/auth/login.ts#L20-L27) and `../../lib/log.ts` a few lines up — uses a `.ts` specifier, as do ~1100 imports across `cli-core`. There's a small pocket of `.js` specifiers ([`lib/skills.ts:10`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/lib/skills.ts#L10), [`init/scan.ts:2`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/init/scan.ts#L2)) and the parent commit added one more at [`init/index.ts:24`](https://github.com/clerk/cli/blob/d3ff1393382fcad8a318a44b885a9a02380a85b4/packages/cli-core/src/commands/init/index.ts#L24), so this is drifting rather than isolated. Bun resolves both, but it's worth keeping the new ones consistent.

<details>
<summary>Suggested fix</summary>

```suggestion
import { setTelemetryStage } from "../../lib/telemetry.ts";
```

</details>

###### Sent from Claude

- [x] Reviewed by @wyattjoh

packages/cli-core/src/commands/auth/login.ts
djgould: Fixed in bf122db3 — took the wrapper: `currentTelemetryStage()` reader, and `login()` hands the caller's stage back on a clean return (login's own stage stands on a throw). Mutation-checked: the nested test fails with the restore removed.

packages/cli-core/src/commands/auth/login.ts
djgould: Fixed in bf122db3 — `first_application` stage added before `ensureFirstApplication`.

packages/cli-core/src/lib/auth-server.test.ts
djgould: Fixed in bf122db3 — `timeoutSpy` and the open server moved into the existing afterEach.

packages/cli-core/src/commands/auth/login.test.ts
djgould: Fixed in bf122db3 — spy restore moved to afterEach, and added the nested-login handback test plus a `store` pin. Skipped a `session_check` pin: `getExistingSession` swallows every failure, so no run can terminate at that stage.

packages/cli-core/src/lib/auth-server.ts
djgould: Fixed in bf122db3 — original error (stack included) debug-logged before the CliError wraps it.

.changeset/auth-login-error-codes.md
djgould: Fixed in bf122db3 — `patch`.

packages/cli-core/src/commands/auth/login.ts
djgould: Fixed in bf122db3 here, and the parent's `init/index.ts` specifier in c4658a78 on #441.

wyattjoh (approved):

PR #439, opened 2026-08-21
feat(init): reconcile native iOS configuration

seanperez29:
## Summary

- wire iOS inspection and local setup into clerk init with preview, dry-run, consent, and apply boundaries
- plan and reconcile native application registration against the selected Clerk development instance
- add explicit opt-ins for Sign in with Apple and the prebuilt ClerkKitUI authentication experience
- extend the supporting Platform API, link, environment, deploy, fixtures, E2E coverage, and release changeset

## Stack

- PR 1: #431 — iOS project inspection foundations
- PR 2: #438 — transactional local iOS setup
- PR 3 of 4: this PR
- PR 4: #432 — iOS-aware doctor diagnostics

## Safety

- dry-run stays unauthenticated and read-only
- local and remote mutations have separate preview and consent boundaries
- sensitive keys and provider configuration stay out of plans, logs, JSON output, and telemetry
- ambiguous, generated, unsafe, dirty, or stale project states remain blocked for manual review

## Testing

- bun run format:check
- bun run lint
- bun run typecheck
- bun run test — 3,129 passing
- bun run test:e2e:op was attempted; local 1Password authorization timed out before the E2E suite started

PR #444, opened 2026-08-24
ci(release): reject fork PRs before checking out snapshot code

dominic-clerk:
## Summary

- The `!snapshot` job warned on fork PRs but still checked out the fork's head and ran `bun install` + `bun run version-packages:snapshot` in the default-branch `issue_comment` context with a `pull-requests: write` token — `is_fork` only gated the downstream jobs. Fork-controlled `package.json` scripts / `bunfig.toml` ran before the warning could be acted on.
- Fail the job **before** checkout when `head.repo != base.repo` (also covers deleted forks).
- Add the same freshness guard as `clerk/javascript`'s `snapshot-release`: fail if the PR was updated after the `!snapshot` comment was posted, so the reviewer approved the code that actually runs.
- `persist-credentials: false` on the snapshot checkout.
- Drop the `is_fork` output and the five downstream `if:` gates — a failed `snapshot` job already stops the chain and `snapshot-notify-failure` posts the failure comment.

## Context

The fork check was added in #45 to keep fork code away from the OIDC `publish` job (`is_fork` gates on downstream jobs). Review on that PR also flagged that the `snapshot` job itself still checks out and runs the fork head before the gate applies (`snapshot.yml:69` thread); that part was never addressed and carried through the #135 consolidation. This PR closes that remaining gap.

## Test plan

- [x] `actionlint .github/workflows/release.yml` passes
- [ ] `!snapshot` on a same-repo PR still publishes
- [ ] `!snapshot` on a fork PR fails at "Validate PR source and freshness" with no checkout, and the failure comment is posted

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -381,7 +381,6 @@ jobs:
     outputs:
       version: ${{ steps.version.outputs.version }}
       sha: ${{ steps.pr.outputs.sha }}
-      is_fork: ${{ steps.pr.outputs.is_fork }}
     steps:
       - name: React to comment
         env:
@@ -391,27 +390,36 @@ jobs:
           gh api "repos/${GH_REPO}/issues/comments/${{ github.event.comment.id }}/reactions" \
             -f content=eyes --silent
 
-      - name: Get PR head SHA
+      # Runs before checkout: this job executes the PR's package.json scripts
+      # in the default branch's privileged context, so fork code must never be
+      # fetched. Also reject PRs that changed after the comment was posted so
+      # the reviewer approved the code that actually runs.
+      - name: Validate PR source and freshness
         id: pr
         env:
           GH_TOKEN: ${{ github.token }}
           GH_REPO: ${{ github.repository }}
+          COMMENT_CREATED_AT: ${{ github.event.comment.created_at }}
         run: |
           pr_json=$(gh api "repos/${GH_REPO}/pulls/${{ github.event.issue.number }}")
           sha=$(echo "$pr_json" | jq -r '.head.sha')
           head_repo=$(echo "$pr_json" | jq -r '.head.repo.full_name')
           base_repo=$(echo "$pr_json" | jq -r '.base.repo.full_name')
-          is_fork="false"
+          pr_updated_at=$(echo "$pr_json" | jq -r '.updated_at')
           if [ "$head_repo" != "$base_repo" ]; then
-            is_fork="true"
-            echo "::warning::Snapshot requested on fork PR (head: ${head_repo}) — publish will be skipped."
+            echo "::error::Snapshots are restricted to branches within ${base_repo} (PR head: ${head_repo})."
+            exit 1
+          fi
+          if [[ "$pr_updated_at" > "$COMMENT_CREATED_AT" ]]; then
+            echo "::error::The PR has been updated since !snapshot was posted. Review the changes and comment !snapshot again."
+            exit 1
           fi
           echo "sha=${sha}" >> "$GITHUB_OUTPUT"
-          echo "is_fork=${is_fork}" >> "$GITHUB_OUTPUT"
 
       - uses: actions/checkout@v7
         with:
           ref: ${{ steps.pr.outputs.sha }}
+          persist-credentials: false
       - uses: oven-sh/setup-bun@v2
       - run: bun install --frozen-lockfile
 
@@ -429,7 +437,6 @@ jobs:
 
   snapshot-ci:
     needs: snapshot
-    if: needs.snapshot.outputs.is_fork != 'true'
     uses: ./.github/workflows/ci.yml
     with:
       ref: ${{ needs.snapshot.outputs.sha }}
@@ -437,7 +444,6 @@ jobs:
 
   snapshot-build:
     needs: [snapshot, snapshot-ci]
-    if: needs.snapshot.outputs.is_fork != 'true'
     uses: ./.github/workflows/build-binaries.yml
     with:
       version: ${{ needs.snapshot.outputs.version }}
@@ -447,7 +453,6 @@ jobs:
 
   snapshot-sign-macos:
     needs: [snapshot, snapshot-build]
-    if: needs.snapshot.outputs.is_fork != 'true'
     uses: ./.github/workflows/sign-macos.yml
     with:
       artifact-prefix: clerk-snapshot
@@ -460,7 +465,6 @@ jobs:
 
   snapshot-smoke-test:
     needs: [snapshot, snapshot-build, snapshot-sign-macos]
-    if: needs.snapshot.outputs.is_fork != 'true'
     uses: ./.github/workflows/smoke-test.yml
     with:
       version: ${{ needs.snapshot.outputs.version }}
@@ -469,7 +473,6 @@ jobs:
 
   snapshot-publish:
     needs: [snapshot, snapshot-build, snapshot-sign-macos, snapshot-smoke-test]
-    if: needs.snapshot.outputs.is_fork != 'true'
     # Must run on GitHub-hosted runner for npm OIDC trusted publishing
     runs-on: ubuntu-latest
     timeout-minutes: 15
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -393,9 +393,11 @@ jobs:
 
       # Runs before checkout: this job executes the PR's package.json scripts
       # in the default branch's privileged context, so fork code must never be
-      # fetched. Also reject PRs whose head commit is newer than the comment so
-      # the reviewer approved the code that actually runs. Keyed on the commit
-      # date rather than the PR's updated_at, which moves on comments/labels.
+      # fetched. The fork check plus the MEMBER/OWNER gate above are the trust
+      # boundary. The freshness check is only a race guard for the honest case
+      # (a colleague pushes between the comment and this step); it keys on the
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -391,27 +390,36 @@ jobs:
...
           base_repo=$(echo "$pr_json" | jq -r '.base.repo.full_name')
-          is_fork="false"
+          pr_updated_at=$(echo "$pr_json" | jq -r '.updated_at')
           if [ "$head_repo" != "$base_repo" ]; then
-            is_fork="true"
-            echo "::warning::Snapshot requested on fork PR (head: ${head_repo}) — publish will be skipped."
+            echo "::error::Snapshots are restricted to branches within ${base_repo} (PR head: ${head_repo})."
+            exit 1
+          fi
+          if [[ "$pr_updated_at" > "$COMMENT_CREATED_AT" ]]; then
```
wyattjoh: `.updated_at` on a PR is the issue record's timestamp, not a code timestamp. It moves for comments, comment edits, labels, renames, review requests, review submissions, and cross-references — not just pushes. So this gate rejects on activity that has nothing to do with the code being published.

I checked this against this repo's own PRs. On #436 the head commit landed at `19:06:16` but `updated_at` is `20:19:13`, moved by a `review_requested` event and then the merge. On #439 the last comment was `21:00:59` and `updated_at` is `21:01:30`. On #442 `updated_at` is exactly the last comment's `created_at`, which confirms the mechanism.

The happy path does survive: posting `!snapshot` sets `updated_at` to exactly `comment.created_at`, and the comparison is strict `>`, so a quiet PR passes. But the window between the comment and this API call is wide open. `coderabbitai` and `changeset-bot` both comment and *edit* their comments on PRs here — on #443 the coderabbit comment was created at `14:06:10` and edited at `14:07:10`. Any of that landing in the seconds before this step runs fails the job with a message telling the maintainer their code changed when it did not.

It fails closed, which is the right direction, but it will fail closed for the wrong reason. Key on something code-specific instead:

```bash
head_committed_at=$(gh api "repos/${GH_REPO}/commits/${sha}" --jq '.commit.committer.date')
if [[ "$head_committed_at" > "$COMMENT_CREATED_AT" ]]; then
```

Committer dates are client-settable, but the trust boundary here is already MEMBER/OWNER on a same-repo branch, so that is not the threat this gate defends against. If you want something not forgeable, walk the issue timeline for `head_ref_force_pushed` and `committed` events newer than the comment.

###### Sent from Claude

.github/workflows/release.yml
```diff
@@ -469,7 +473,6 @@ jobs:
 
   snapshot-publish:
```
wyattjoh, line 482: `persist-credentials: false` is a good addition on the `snapshot` job's checkout, but `snapshot-publish` a few lines below checks out the same PR SHA without it:

```yaml
      - uses: actions/checkout@v7
        with:
          ref: ${{ needs.snapshot.outputs.sha }}
```

That job is the more privileged of the two — it holds `id-token: write` for npm OIDC trusted publishing, runs `bun install --frozen-lockfile` against the same PR-authored `package.json`, and then publishes. If leaving the `GITHUB_TOKEN` in `.git/config` is worth avoiding in the versioning job, it is worth avoiding here. Nothing in that job pushes or fetches after checkout, so adding the flag should be free.

###### Sent from Claude

.github/workflows/release.yml
```diff
@@ -391,27 +390,36 @@ jobs:
...
-          is_fork="false"
+          pr_updated_at=$(echo "$pr_json" | jq -r '.updated_at')
           if [ "$head_repo" != "$base_repo" ]; then
-            is_fork="true"
-            echo "::warning::Snapshot requested on fork PR (head: ${head_repo}) — publish will be skipped."
+            echo "::error::Snapshots are restricted to branches within ${base_repo} (PR head: ${head_repo})."
+            exit 1
+          fi
+          if [[ "$pr_updated_at" > "$COMMENT_CREATED_AT" ]]; then
+            echo "::error::The PR has been updated since !snapshot was posted. Review the changes and comment !snapshot again."
```
wyattjoh: Both new rejections are user-caused conditions a maintainer needs to act on, but `::error::` annotations only show up in the run log. What actually lands on the PR is `snapshot-notify-failure`'s 😕 reaction plus:

> ## Snapshot failed
> The snapshot publish workflow failed. [View the workflow run](...) for details.

So "you commented `!snapshot` on a fork PR" and "the branch moved, comment again" are indistinguishable from "the macOS notarization step broke" until someone opens the logs. This is a regression in feedback quality from the old fork path, which at least completed green.

Since the job has `pull-requests: write` already, consider posting the reason directly from this step before exiting — or have the validation step write its reason to an output that `snapshot-notify-failure` folds into the comment body.

###### Sent from Claude

wyattjoh (commented):
The reordering is the right fix — validating before `actions/checkout` closes the window where fork code was fetched and `bun install`ed in the privileged context, and failing the job outright is more durable than five duplicated `if:` guards. My concerns are all with the freshness gate that came along for the ride.

###### Sent from Claude

.github/workflows/release.yml
```diff
@@ -391,27 +390,36 @@ jobs:
           gh api "repos/${GH_REPO}/issues/comments/${{ github.event.comment.id }}/reactions" \
             -f content=eyes --silent
 
-      - name: Get PR head SHA
+      # Runs before checkout: this job executes the PR's package.json scripts
+      # in the default branch's privileged context, so fork code must never be
+      # fetched. Also reject PRs that changed after the comment was posted so
```
wyattjoh: The Snapshot section in `docs/releasing.md` documents the MEMBER/OWNER trigger restriction but predates both gates this step adds. Worth a sentence each: snapshots only run on branches in `clerk/cli`, and the PR must not have been touched since the `!snapshot` comment. Both are things a maintainer will otherwise discover by getting a red X.

Worth noting for the commit message too — the subject covers the fork rejection, but the freshness gate and `persist-credentials: false` are separate hardening changes riding along in the same commit.

###### Sent from Claude

.github/workflows/release.yml
dominic-clerk: Agreed — `updated_at` is the issue timestamp and the bot comment/edit race is real here. Switched to `repos/{repo}/commits/{sha}` → `.commit.committer.date` in e548e59, and set `LC_ALL=C` on the step so the string compare doesn't depend on runner collation. Stopped short of walking the timeline: this gate is a courtesy against accidental races between same-repo collaborators, not a security boundary, so a client-settable date is fine.

.github/workflows/release.yml
dominic-clerk: Done in e548e59. Also gave `snapshot-publish` its own fork re-check before checkout — that job mints the OIDC token, so it shouldn't rely on an upstream step for the boundary. Confirmed `persist-credentials: false` is safe there: `scripts/releaser.ts` only tags/pushes when `--tag` is unset (stable), and the PR comment uses `GH_TOKEN`, not git credentials.

.github/workflows/release.yml
dominic-clerk: Done in e548e59 via the output route: the validate step writes `reason=…` before `exit 1`, the job exposes it, and `snapshot-notify-failure` puts it at the top of the comment (run link stays below). Went with this over posting from the step itself to avoid a double comment. Added a test-plan item to confirm the output propagates through `needs` on a failed job.

.github/workflows/release.yml
dominic-clerk: Docs updated in e548e59 (one sentence covering both gates). For the commit message: kept the original commit as-is rather than force-pushing over a reviewed commit; the PR body already lists the freshness gate and `persist-credentials` separately, so the squash message will carry them.

.github/workflows/release.yml
```diff
@@ -391,27 +391,41 @@ jobs:
...
+            exit 1
+          }
           if [ "$head_repo" != "$base_repo" ]; then
-            is_fork="true"
-            echo "::warning::Snapshot requested on fork PR (head: ${head_repo}) — publish will be skipped."
+            reject "Snapshots are restricted to branches within ${base_repo} (PR head: ${head_repo})."
+          fi
+          head_committed_at=$(gh api "repos/${GH_REPO}/commits/${sha}" --jq '.commit.committer.date')
+          if [[ "$head_committed_at" > "$COMMENT_CREATED_AT" ]]; then
+            reject "The PR head moved after \`!snapshot\` was posted (${sha:0:7} committed at ${head_committed_at}). Review the changes and comment \`!snapshot\` again."
```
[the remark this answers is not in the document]
dominic-clerk: Not adopting this. The freshness gate isn't a security boundary — the fork check and the MEMBER/OWNER comment gate are. Exploiting a backdated commit requires write access to `clerk/cli` (to push to a same-repo branch) plus timing the push inside the seconds between the comment and this step; anyone with that access is already inside the trust perimeter, and the push is visible on the PR immediately. The gate exists for the honest race (reviewer approves X, a colleague pushes Y before the runner picks it up), and the commit date handles that.

The suggested fix also doesn't provide what it claims: there is no `synchronize` event in the issue timeline API (that's a webhook action). Timeline `committed` events carry the same git author/committer dates as the commit object, so walking the timeline is only unforgeable for `head_ref_force_pushed`. The server-side signal would be `PushEvent.created_at` from the Events API, which is capped and delayed — not something to put in a release gate.

Tightened the comment above the step in 19c8251 so nobody later treats this check as a trust boundary.

wyattjoh (approved):

PR #469, opened 2026-09-02
docs: make README agent-first for init

s0uthpaw:
## Summary
- Add an agent-first "Add Clerk to an app" section to the CLI README
- Lead with `npx -y clerk@latest init` and scope accountless wording to supported frameworks
- Add `clerk doctor` as the setup verification step

## Tests
- `git diff --check -- README.md`
- Not run: Bun README/format checks because `bun`/`bunx` are unavailable on PATH in this shell

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,41 @@
 
 The Clerk command-line interface.
 
+## Add Clerk to an app
+
+```sh
+npx -y clerk@latest init
+```
+
+`clerk init` is the setup path for developers and coding agents. It detects the
+framework, installs the matching Clerk SDK for npm-based projects when it is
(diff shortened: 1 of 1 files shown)


royanger (approved):
🚀

manovotny:
Pushed some changes directly in c22a9182.

- Scoped the accountless setup claims to a first `clerk init` run, since later runs keep an existing unclaimed app instead of replacing it unless you pass `--fresh`.
- Dropped "selected" from the env-file bullet — the accountless path re-detects the env file on its own rather than reusing the one init picks in context, so the two can diverge.

manovotny (approved):

PR #471, opened 2026-09-03
fix(auth): explain managed-workspace claim rejections during login

Zertsov:
## Problem

The Platform API now returns 403 `accountless_application_managed_workspace` from `POST /v1/platform/accountless_applications/claim` when the token's workspace is provisioned by an integration such as Vercel Marketplace or Stripe. The CLI mapped every 403 to "your account does not have an active organization", which is wrong for that case and gives no way forward.

## Change

- `classifyClaimError` checks the error code before the status table. A 403 with the managed-workspace code becomes a new `managed_workspace` result; any other 403 keeps `no_organization`.
- The warning prints the API's `long_message` verbatim after `Unable to claim - `, for example: `Unable to claim - The target application cannot be claimed into the current workspace. Select a different workspace and try again.` If the body has no message, the fallback is `Unable to claim - this workspace is managed by an integration provider.`
- The claim breadcrumb is kept for `managed_workspace`, so the next `clerk auth login` from a workspace the user owns retries the claim.
- A next-steps entry for the same case. Patch changeset for `clerk`.

## Linear

- AIE-1668 — implemented here. This PR owns the CLI classification, message, and retry guidance for the managed-workspace rejection.

## Tests

Unit tests cover classification with and without the error code, and the warning text with and without `long_message`. `cli-core` typecheck and lint clean.


packages/cli-core/src/lib/next-steps.ts
```diff
@@ -33,6 +33,10 @@ export const NEXT_STEPS = {
     "Run `clerk auth login` again to retry auto-claim",
     "Run `clerk link` to connect your application manually",
   ],
+  AUTOCLAIM_SWITCH_WORKSPACE: [
+    "Switch to a workspace you own in the Clerk Dashboard",
+    "Run `clerk auth login` again to claim the application",
+  ],
```
thiskevinwang, line 39: 💭 
```suggestion
  AUTOCLAIM_SWITCH_WORKSPACE: [
    "Select a different workspace in the Clerk Dashboard",
    "Run `clerk auth login` again to proceed,
  ],
```

Zertsov: Applied via the suggestion commit; the missing closing quote is fixed in 02c67eec.

packages/cli-core/src/lib/autoclaim.ts
```diff
@@ -28,7 +39,7 @@ export async function attemptAutoclaim(cwd: string): Promise<AutoclaimResult> {
   const appName = await deriveProjectName(cwd);
   const result = await tryClaim(breadcrumb.claimToken, appName);
 
-  if (result.status === "failed") return result;
+  if (result.status === "failed" || result.status === "managed_workspace") return result;
```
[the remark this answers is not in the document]
Zertsov: Not changing this. The early return for `"failed"` is pre-existing behavior on `main` (`autoclaim.ts:31` before this PR): a transient claim failure keeps the breadcrumb on purpose, and login already says "It will be retried on your next `clerk auth login`." This PR only adds `"managed_workspace"` to the same condition.

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -224,6 +224,37 @@ const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
     "Auto-claim failed due to a temporary error. It will be retried on your next `clerk auth login`.",
 };
 
+const MANAGED_WORKSPACE_FALLBACK_REASON = "this workspace is managed by an integration provider.";
```
thiskevinwang: Following BE:`accountless_application_managed_workspace` -> CLI:`managed_workspace` -> BE:longMessage...

...`MANAGED_WORKSPACE_FALLBACK_REASON` looks like it'll never used. I'd say drop this.

packages/cli-core/src/commands/auth/login.ts
```diff
@@ -224,6 +224,37 @@ const CLAIM_WARNINGS: Partial<Record<AutoclaimResult["status"], string>> = {
...
+}
+
+function lowercaseFirst(text: string): string {
+  return text.charAt(0).toLowerCase() + text.slice(1);
+}
+
+function claimWarning(result: AutoclaimResult): string | undefined {
+  if (result.status === "managed_workspace") return managedWorkspaceWarning(result.longMessage);
+  return CLAIM_WARNINGS[result.status];
+}
```
thiskevinwang: This feels like an unnecessary amount of message massaging.

Any reason we can't use the longmessage as-is?

packages/cli-core/src/commands/auth/login.ts
Zertsov: Dropped in `6ed181b0bd`. The CLI now prints the API `long_message` verbatim after `Unable to claim - `, with a one-line fallback only when the body has no message. Resolving.

<span title="This comment was written by AI, reviewed &amp; directed by Mitch">~ 🤖</span>

packages/cli-core/src/commands/auth/login.ts
Zertsov: Agreed. `6ed181b0bd` and `5111a80513` print the `long_message` as-is, no sentence surgery. Resolving.

<span title="This comment was written by AI, reviewed &amp; directed by Mitch">~ 🤖</span>

wyattjoh (approved):

PR #475, opened 2026-09-08
ci(release): sign Windows binaries with Azure Artifact Signing

### LINKED ISSUE #474: when i type clerk init
Clerk 3.3.0 on Windows x64: clerk.exe is unsigned and blocked by application control, producing Code Integrity event 3077 and spawnSync UNKNOWN.

wyattjoh:
## Summary

Windows binaries have never been Authenticode-signed. On machines running WDAC, Smart App Control, or AppLocker in enforcing mode, the OS refuses to execute the unsigned `clerk.exe` entirely — the block surfaces to users as `Failed to start Clerk CLI: spawnSync ... UNKNOWN` from the npm wrapper shim, with Code Integrity event 3077 in the Windows event log. This affects every Windows user on a managed machine, not just the `init` command.

This adds a `sign-windows` reusable workflow that mirrors `sign-macos`, and wires it into the stable, canary, and snapshot pipelines so publishing is gated on signing succeeding. Signing runs through Azure Artifact Signing (formerly Trusted Signing), authenticating via OIDC federated credentials rather than a stored certificate. The job runs on a Windows runner because the signing engine is Windows-only, and it has no local script because the action owns credential handling and the `signtool` invocation.

Azure issues short-lived leaf certificates of roughly 72 hours, so the RFC-3161 countersignature is what keeps a published binary trusted after the leaf expires. The verify step therefore treats a missing timestamp as a hard failure alongside an invalid signature status, and uses `Get-AuthenticodeSignature` rather than locating `signtool.exe` in the Windows SDK.

All six Azure secrets are declared `required: true`, so a release fails loudly rather than silently shipping an unsigned binary. Each of the three call sites sets `permissions: id-token: write`, since the repository default of `contents: read` would otherwise deny the OIDC token.

## Blocked on provisioning

This is a draft because the release will fail until the Azure side exists: an Artifact Signing account and certificate profile, completed Microsoft identity validation for the Clerk org, an app registration holding the `Trusted Signing Certificate Profile Signer` role, and the six repository secrets.

The federated credential subject must be `repo:clerk/cli:ref:refs/heads/main`. OIDC inside a reusable workflow keys off the caller's ref, and all three channels are driven from `main` — including snapshots, which trigger on `issue_comment` — so one branch-scoped credential covers every path.

## Test plan

- [ ] `bun run format:check`, `bun run lint`, `bun run typecheck`, `bun run test` pass locally
- [ ] Workflow YAML parses and every `needs` reference resolves to a defined job
- [ ] After secrets are provisioned, a `!snapshot` run signs both win32 targets and the verify step reports `Valid` with a timestamp
- [ ] Confirm the signed `clerk.exe` executes on a WDAC-enforcing Windows machine

Fixes #474

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -82,8 +82,25 @@ jobs:
       APPLE_API_KEY_ID: ${{ secrets.APPLE_API_KEY_ID }}
       APPLE_API_ISSUER_ID: ${{ secrets.APPLE_API_ISSUER_ID }}
 
+  sign-windows:
+    needs: [versioning, build]
+    if: needs.versioning.outputs.release_created == 'true'
+    permissions:
+      contents: read
+      id-token: write
+    uses: ./.github/workflows/sign-windows.yml
+    with:
+      artifact-prefix: clerk
+    secrets:
+      AZURE_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
+      AZURE_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
+      AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+      AZURE_SIGNING_ENDPOINT: ${{ secrets.AZURE_SIGNING_ENDPOINT }}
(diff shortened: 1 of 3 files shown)


Dev-v1:
thank you

